Issue #45, opened 2026-05-04
.NET QueueWatcherBackend NoFallback tests time out across all platforms


russellromney:
## what

`Honker.Tests.BindingTests.QueueWatcherBackend{ManyWritersOneWorker, OneWriterOneWorker, OneWriterManyWorkers}NoFallback` time out (~15 s each) on all three .NET platforms (Linux / macOS / Windows) for every backend value (`null`, `"kernel"`, `"shm"`).

## why it matters

these are pre-existing failures on `main` — first seen on commit `ab6849c` ("Prepare binding parity release") in run [25313027526](https://github.com/russellromney/honker/actions/runs/25313027526). they propagate into every PR opened against main and look like that PR's fault.

PR #42 is the most recent example — Phase Mantle landed cleanly across every binding, but the .NET matrix shows red because of these inherited failures. dotnet built fine, every PhaseMantleTests case passes; only these QueueWatcherBackend NoFallback tests time out.

similar shape to #44 (`OutboxLockTests.OutboxRetriesOnException` / `BindingTests.WaitResultObservesSaveRacingTheCall` flakes) — race tests with tight timeouts that don't tolerate CI scheduling jitter. likely the worker isn't getting woken inside the 15 s window when fallback polling is disabled.

## what to look at

- `BindingTests.QueueWatcherBackendManyWritersOneWorkerNoFallback`
- `BindingTests.QueueWatcherBackendOneWriterOneWorkerNoFallback`
- `BindingTests.QueueWatcherBackendOneWriterManyWorkersNoFallback`

each fires for backend ∈ { `null`, `"kernel"`, `"shm"` } so 9 failures total per platform. all timing out at exactly 15 s, which suggests the wake never arrives rather than arriving slow. `NoFallback` in the name means polling fallback is disabled — so this exercises the watcher backend's wake path directly.

most likely cause: the .NET update watcher poll cadence or the test's `WaitUntilAsync` window is too tight for CI runners. similar to the bumps we did on the kernel-watcher latency test.

## not in scope here

- not blocking PR #42 on this — we'll merge despite the failure, since it's pre-existing.
- not touching the .NET binding's underlying watcher logic. these tests are exercising real wake mechanics; the failure means the test's tolerance for slow runners is too tight, not that the binding is broken.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

russellromney:
Fixed by 285d9bb on main ("Stabilize watcher proof CI"). Same root cause as #44.

[...]

PR #46, opened 2026-05-04
roadmap: Phase DiMaggio + Berra; fix .NET timeouts (closes #44, #45)

### LINKED ISSUE #44: .NET test suite has timing-flaky race tests
## what

`Honker.Tests.BindingTests.WaitResultObservesSaveRacingTheCall` flakes intermittently on CI. Hit it on PR #42's macOS dotnet job (run [25307232927](https://github.com/russellromney/honker/actions/runs/25307232927)). Earlier flake of a different .NET test (`OutboxLockTests.OutboxRetriesOnException`) on PR #30's Linux job has the same shape.

## why it matters

it's a real failure mode, not a setup/network issue. test fails because the racy thing it's trying to test loses the race under CI load. two consecutive PR runs hitting different .NET race tests means the .NET test suite has at least two flaky tests sitting on tight timing.

doesn't block any PR — both have been retry-able — but each flake costs a CI cycle and adds noise.

## what to look at

- `WaitResultObservesSaveRacingTheCall` (this PR) — a wait-for-result test that races result-save against the wait call. probably needs a longer timeout / WaitUntilAsync window for slow runners.
- `OutboxLockTests.OutboxRetriesOnException` (PR #30) — outbox retry test with a similar timeout shape.

both probably want the same fix: bump the `WaitUntilAsync` timeout or use a more generous polling cadence under CI.

## not in scope here

- not touching the .NET binding's underlying logic. these tests are exercising real race conditions; the failure means the test's tolerance for slow runners is too tight, not that the binding is broken.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

russellromney:
two unrelated changes bundled because they're both small:

## roadmap (HN feedback)

- **Phase DiMaggio** — adaptive polling backoff. After IDLE_THRESHOLD_S of no commits, scale poll interval up geometrically (1ms → cap at 1s). Reset to 1ms on any wake. Addresses both the polling-overhead critique and the mobile/battery concern from the HN thread. Default backend, no opt-in.

- **Phase Berra** — HN-feedback docs sweep. Update README prior-art for Oban-supports-SQLite (@arlobish), add Graphile Worker (@odie5533), publish bench numbers (@andrewstuart), add a "When NOT to use" + multi-process WAL story (@ncruces / defangs the "SQLite isn't concurrent" critique from @maxdo), surface JVM binding in README + nav.

## .NET test timeout fixes — closes #44, #45

`WaitReady` 15s → 60s. The helper is spawned via `dotnet test --filter ...` which has 5-15s of test-discovery boot overhead. The original window had no margin — every `QueueWatcherBackend*NoFallback` test was timing out at exactly 15s waiting for the helper to write its ready file. 60s comfortably covers boot; locally still returns immediately.

`WaitResultObservesSaveRacingTheCall` < 1s → < 1.5s. The assertion is "wake came via watcher, not via the 5s fallback poll." 1s was too tight (Thread.Sleep(50) on Windows can be 60-90ms, plus inter-thread coordination); 1.5s still rules out the fallback by 3x.

`OutboxRetriesOnException` WaitUntilAsync 3s → 10s, outer cts 5s → 15s. 3 retries + backoff + worker scheduling needed more headroom.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -550,6 +550,95 @@ intentionally last in the sequence after the easier wins.
   rate-limit-ish patterns; revisit if it's not enough.
 - No quota / capacity limits per queue. SQLite is the limit.
 
+## Phase DiMaggio — Adaptive Polling Backoff
+
+> After: Phase Echo · Before: 1.0 release prep
+
+DiMaggio's hitting streak was 56 games of consistent contact — pace
+matching context. The default polling backend currently runs at a
+fixed 1 ms cadence regardless of activity: idle databases burn the
+same CPU as busy ones. Recurring HN ask: "wind back the polling to
+once a second when nothing is happening."
+
+This is a change to the **default** backend (not the experimental
+ones). Affects everyone, no opt-in. Addresses both the polling-
+overhead critique and the mobile / battery concern in one move.
+
+### Scope
+
+- [ ] Track `time_since_last_wake` in `run_poll_loop`.
+- [ ] If > `IDLE_THRESHOLD_S` (default 5 s), scale poll interval up
+      geometrically (1 ms → 2 → 4 → ... cap at `MAX_POLL_INTERVAL_MS`,
+      default 1000 ms).
+- [ ] Reset to 1 ms on any wake (commit observed via PRAGMA, or any
+      conservative-wake path).
+- [ ] Configurable via `WatcherConfig::idle_threshold_s` and
+      `max_poll_interval_ms` (defaults sensible; users don't need
+      to think about it).
+- [ ] Wire through Python + Node bindings (other bindings track
+      under Phase Echo).
+
+### Acceptance
+
+- [ ] Idle database: after `IDLE_THRESHOLD_S` of no commits, poll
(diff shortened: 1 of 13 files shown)


russellromney:
Superseded.

Main shipped 285d9bb "Stabilize watcher proof CI" while this PR was iterating, which independently solves the same .NET subprocess-helper problem (and does it more cleanly — factored `RunWorker`/`RunWriter`, `RequiredEnv` helper).

The roadmap additions (Phase DiMaggio + Phase Berra) carved out into #47.

Other branch work that didn't make the cut:
- The `SHARED_WATCHERS` cache + `signal_close` split in honker-extension. Real perf win but coupled to a UpdatePoller redesign main doesn't need; safe port requires more surgery than a cherry-pick. Leaving as a follow-up if/when poller spawn cost actually shows up in a profile.
- `UpdatePoller` dispatcher rewrite + `Database` eager-poller — only made sense paired with the cache.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #57, opened 2026-05-23
Bundle the SQLite extension in the Ruby gem (and compile it for other platforms)

cllns:
First of all, thanks for creating Honker! I'm a core team member of [Hanakai](https://hanakai.org/hanami), which includes the web framework Hanami. We don't pull in any Rails dependencies so this give us some great parity with the Solid* gems from Rails. Happy to help see this project succeed!

This mirrors PR #53 (Python wheel packaging) for the Ruby binding. The `honker` gem, now 0.2.0, ships the SQLite loadable extension itself, so `Honker::Database.new` resolves it automatically and `extension_path:` is optional.

Heavily used Claude for this. Tested it to the extent possible without being able to run GitHub actions and do a real release.

## Summary

- Precompiled platform gems for x86_64 Linux, aarch64 Linux, and Apple Silicon macOS bundle the prebuilt extension in `lib/honker/` (where sqlite and nokogiri put it, the convention in ruby gems). These are the three most popular platforms, so they're the highest priority.
- The generic gem (every other platform, and `github:` installs) compiles the extension from bundled Rust source on install, the way sqlite and nokogiri do. A Rust toolchain is required for that path.
- New `release-ruby.yml` workflow builds, smoke-tests, and publishes every gem.
- New `Honker::ExtensionResolver`; `extension_path:` is now an optional keyword. `Honker::Error` is now defined (it was referenced by `CoreWatcher#wait` but never declared, a latent `NameError`).
- Declares the `fiddle` runtime dependency (no longer a default gem in Ruby 3.5+) and sets `rubygems_mfa_required`.

## Design notes and trade-offs

- **Platform gems, not one fat gem.** One precompiled gem per OS/arch; RubyGems picks the match automatically. Costs more release artifacts, but each install is instant with nothing to compile.
- **Three precompiled targets.** Windows and Intel macOS are intentionally not precompiled (Windows cannot be verified locally; both are lower-traffic); those users get the generic gem instead.
- **The generic gem compiles on install** instead of shipping nothing. Trade-off: a source install needs a Rust toolchain. `extconf.rb` does a pre-flight check and fails with a clear, actionable message when `cargo` is missing. The payoff: `gem install` works on every platform, and `github:` works.
- **`extension_path:` is now optional**, resolved in order: an explicit argument, then `HONKER_EXTENSION_PATH`, then the bundled extension. Backward compatible; an explicit path still works.
- **Release-workflow actions are SHA-pinned.** The workflow holds the RubyGems publish secret, and the repo's zizmor lint requires pinning for any workflow except `ci.yml`.

## Scripts
| When | Script | Purpose |
| --- | --- | --- |
| **CI / stage** | `scripts/copy-ruby-extension.sh` | Stages the extension (built ahead of time by `cargo build --release -p honker-extension` on the matching CI runner) into `lib/honker/` for a platform gem. |
| **CI / stage** | `scripts/copy-ruby-sources.sh` | Vendors the `honker-extension` and `honker-core` crate source into `ext/honker/` for the generic gem. |
| **CI / verify** | `scripts/proof/check-ruby-gem.rb` | Release proof: a platform gem bundles the extension; `--generic` checks the source gem ships the crate source and a compile step. |
| **CI / verify** | `scripts/proof/ruby-gem-smoke.rb` | Opens a DB and runs an enqueue/claim/ack round-trip against an installed gem. |
| **Install** | `packages/honker-ruby/ext/honker/extconf.rb` | Compiles the extension from the vendored crate source whenever the generic gem is installed: in `release-ruby.yml`'s smoke step, in `verify-ruby-gem-local.sh`, and on end-user machines. |
| **Local** | `scripts/verify-ruby-gem-local.sh` | Full pre-push check: builds, installs, and smoke-tests both gems (the current platform pre-built version and the compiling from source one), then runs the specs (`--docker` adds Linux checks). |

Happy to respond to feedback, but also feel free to take things from here and tweak at your will. No offense taken :)

Workflow triggers:

- `release-ruby.yml` runs on `rb-v*` tags or manual dispatch.
- `ci.yml` runs the Ruby specs on every PR.
- `package-proof.yml` and `scary-nightly.yml` run `package-install-smoke.sh` on a weekly/nightly schedule.

## Verification

- macOS: platform gem and generic (compile-on-install) gem built, installed, and smoke-tested; all specs pass.
- Docker `linux/amd64` and `linux/arm64`: the Linux platform gems built and smoke-tested.
- Docker `linux/amd64`: the generic gem compiled on install and passed the smoke test; installing without a Rust toolchain failed cleanly with the actionable error.
- `github:`-style install verified via a local `git:` source with `glob: "packages/honker-ruby/honker.gemspec"`.
- `scripts/verify-ruby-gem-local.sh --docker` is green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -409,6 +409,7 @@ jobs:
           bundle exec ruby -Ilib spec/honker_spec.rb
           bundle exec ruby -Ilib spec/smoke_spec.rb
           bundle exec ruby -Ilib spec/parity_spec.rb
+          bundle exec ruby -Ilib spec/extension_resolution_spec.rb
       - name: Elixir
         working-directory: packages/honker-ex
         env:
--- a/.github/workflows/release-ruby.yml
+++ b/.github/workflows/release-ruby.yml
@@ -1,0 +1,101 @@
+name: Release · RubyGems
+
+on:
+  push:
+    tags:
+      - "rb-v*"
+  workflow_dispatch:
+    inputs:
+      publish:
+        description: "Upload built gems to RubyGems"
+        required: true
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+jobs:
+  platform-gems:
+    name: ruby gem · ${{ matrix.gem_platform }}
+    timeout-minutes: 20
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      # Native builds on a matching runner for each precompiled gem.
+      # Windows, Intel macOS, and musl are intentionally not built here;
+      # those users install the generic gem (see the generic-gem job).
+      matrix:
+        include:
+          - os: ubuntu-latest
+            gem_platform: x86_64-linux
+          - os: ubuntu-24.04-arm
+            gem_platform: aarch64-linux
+          - os: macos-latest
+            gem_platform: arm64-darwin
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8  # stable
+      - uses: ruby/setup-ruby@c4e5b1316158f92e3d49443a9d58b31d25ac0f8f  # v1.306.0
+        with:
+          ruby-version: "3.4"
+
+      - name: Build platform gem
+        shell: bash
+        env:
+          HONKER_GEM_PLATFORM: ${{ matrix.gem_platform }}
+        run: |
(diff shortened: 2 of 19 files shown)


nshki:
Just want to chip in and say I'm looking forward to this work moving forward! I'm a developer interested in using Honker in a Hanami project and found this PR via Sean on the fediverse (Mastodon).

Tested this locally and this branch worked on a Silicon Mac. The precompiled extension wasn't provided since the release workflows aren't yet available, but after getting `cargo` on my system, I was able to run `bundle install` and everything got installed properly. Also confirmed a successful write to a SQLite file.

russellromney:
Thanks for the PR! looks good

PR #64, opened 2026-07-03
docs: fix scheduler README snippet and document queue-in-transaction deadlock

jjshanks:
## Summary

Two documentation bugs in the Python bindings, both found while building an app
against `honker` 0.2.6 and verified against this repo's source and the published
wheel. **Docs/docstrings only — no behavior change.**

## 1. Quick-start scheduler snippet doesn't work

`packages/honker/README.md` showed:

```python
sched = db.scheduler()
sched.add("fast", queue="emails", schedule="@every 1s", payload={"kind": "tick"})
```

Every line of that fails against the real API:

- `Database` has no `scheduler()` method (`hasattr(db, "scheduler")` is `False`).
- `Scheduler.add()` is keyword-only, so the positional `"fast"` is a `TypeError`.
- `schedule` must be a `CronSchedule` from `crontab()` / `every_s()`; a raw
  string raises `AttributeError: 'str' object has no attribute 'expr'`.

Replaced it with the real API — the same form `examples/scheduler.py` already
uses correctly:

```python
from honker import Scheduler, crontab, every_s

sched = Scheduler(db)
sched.add(name="fast", queue="emails", schedule=every_s(1), payload={"kind": "tick"})
sched.add(name="nightly", queue="emails", schedule=crontab("0 3 * * *"))

await sched.run()
```

## 2. Constructing a `Queue` inside an open transaction deadlocks (undocumented)

Calling `db.queue(name)` for a **not-yet-seen** name while a `db.transaction()`
is already open hangs forever:

```python
with db.transaction() as tx:
    q = db.queue("new_name")   # hangs — never returns
    q.enqueue({"x": 1}, tx=tx)
```

`Queue._init_schema` (run on first construction) opens its **own** nested
transaction on the single write connection; the inner `BEGIN` blocks on the lock
the outer transaction holds. Because the wait is in the native extension (GIL
held), it can't be interrupted — Ctrl-C and signal-based timeouts don't fire, so
the process must be `kill`ed. This is easy to hit since the transactional-outbox
pattern is literally "open a transaction, then enqueue."

`Database.queue` had **no docstring at all**. This PR:

- documents the constraint on `Database.queue`, with the correct
  construct-first-then-`tx=` pattern;
- documents `Database.stream` and notes it is **safe** inside a transaction
  (no schema-init), since the asymmetry is surprising;
- adds a maintainer note at `Queue._init_schema` explaining the nested-`BEGIN`
  hazard;
- adds a one-line pointer in the README Notes.

(A behavioral fix — reuse the open transaction, or fail fast instead of hanging —
would be a good follow-up; this PR is docs-only.)

## Verification

- `python -m py_compile packages/honker/python/honker/_honker.py` — passes.
- Ran the corrected README scheduler snippet against 0.2.6 — registers `fast`
  and `nightly` successfully.
- Reproduced the deadlock (`timeout` exit 124) and confirmed the documented
  workaround (construct outside the tx) returns cleanly.

## Scope

Docs and docstrings only — no runtime behavior changes.

### DIFF
--- a/packages/honker/README.md
+++ b/packages/honker/README.md
@@ -65,23 +65,34 @@ import time
 q.enqueue({"to": "later@example.com"}, run_at=int(time.time()) + 10)
 ```
 
-Recurring schedules use `schedule` expressions:
+Recurring schedules are registered on a `Scheduler`. Build the schedule with
+`crontab(expr)` or `every_s(n)`, then run the scheduler loop (it enqueues onto
+the target queue on each boundary; a normal worker claims the jobs):
 
 ```python
-sched = db.scheduler()
-sched.add("fast", queue="emails", schedule="@every 1s", payload={"kind": "tick"})
+from honker import Scheduler, crontab, every_s
+
+sched = Scheduler(db)
+sched.add(name="fast", queue="emails", schedule=every_s(1), payload={"kind": "tick"})
+sched.add(name="nightly", queue="emails", schedule=crontab("0 3 * * *"))
+
+await sched.run()  # acquires the leader lock and fires due tasks
 ```
 
-Supported schedule forms:
+`add()` is keyword-only and `schedule` must be a `CronSchedule` (from `crontab`
+or `every_s`), not a raw string. Supported schedule expressions:
 
-- 5-field cron: `0 3 * * *`
-- 6-field cron: `*/2 * * * * *`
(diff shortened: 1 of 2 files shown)


russellromney:
Thanks! doing a follow up pr as well to address the dedalock

Issue #67, opened 2026-07-20
AbortSignal listener leak in abortPromise() → unbounded O(n²) CPU growth in waitForUpdateOrTimeout loops


Rinse12:
## Summary

`abortPromise()` in `packages/honker-node/api.js` leaks an `AbortSignal` `'abort'` listener on **every** call when the signal doesn't abort. Because it's used inside `Promise.race(...)` in `waitForUpdateOrTimeout()`, which the long-running `ClaimWaker.next()` and `_leaderLoop()` loops call against a signal that lives for the whole process, listeners accumulate without bound. `EventTarget.addEventListener` is O(n) in the current listener count, so cost grows O(n²) and eventually saturates the event loop — in our case pegging a Node process at ~80% CPU and starving unrelated I/O (libp2p fetch streams) after ~a day of uptime.

Seen with `@russellthehippo/honker-node@0.3.3`; the code is unchanged on `main` (0.4.0).

## The bug

`packages/honker-node/api.js:38`:

```js
function abortPromise(signal) {
  if (!signal) return new Promise(() => {});
  if (signal.aborted) return Promise.resolve();
  return new Promise((resolve) => {
    signal.addEventListener('abort', resolve, { once: true }); // :42
  });
}
```

`{ once: true }` removes the listener **only if the `abort` event fires**. But the callers use it in a race:

```js
async function waitForUpdateOrTimeout(updateEvents, signal, timeoutMs) {
  if (aborted(signal)) return;
  if (timeoutMs == null) {
    await Promise.race([updateEvents.next().catch(() => undefined), abortPromise(signal)]);
    return;
  }
  const ms = Math.max(0, timeoutMs);
  await Promise.race([updateEvents.next().catch(() => undefined), delay(ms), abortPromise(signal)]);
}
```

In `ClaimWaker.next()` (`:182`) and `_leaderLoop()` (`:632`) this runs in a `while` loop with a long-lived `signal`. The race is won by `updateEvents.next()` or `delay(ms)` on essentially every iteration, so the `abort` listener created by `abortPromise` is **never** removed. One listener leaks per loop iteration for the life of the signal.

(The losing `delay(ms)` timer also stays pending until it fires, but that's bounded by `timeoutMs`; the unbounded leak is the abort listener.)

## Reproduction

```js
import { getEventListeners } from "node:events";
const delay = (ms) => new Promise((r) => setTimeout(r, ms));
const aborted = (s) => Boolean(s?.aborted);
[...]

PR #69, opened 2026-07-20
fix(node): reuse pending update waits across timeouts

### LINKED ISSUE #68: Node scheduler leaks blocked Tokio threads when update wait loses timeout race
## Summary

`waitForUpdateOrTimeout()` starts a new native `UpdateEvents.next()` call on every polling iteration. When `delay(ms)` wins the `Promise.race`, the losing `next()` promise remains pending and cannot be cancelled. The next iteration starts another one.

The native implementation runs each `next()` with `tokio::task::spawn_blocking()` around `Receiver::recv()`, so repeated timeouts accumulate blocked Tokio tasks and OS threads. In a long-running Docker deployment this reached 526 PIDs (approximately Tokio's 512 blocking-thread ceiling) and about 900 MiB of steady resident memory.

Observed with `@russellthehippo/honker-node@0.3.3`; the same code is present on `main` / 0.4.0.

This is related to, but distinct from, #67: that issue covers the losing `abortPromise()` and its listener. This issue covers the losing native update wait.

## Cause

`packages/honker-node/api.js` currently does:

```js
await Promise.race([
  updateEvents.next().catch(() => undefined),
  delay(ms),
  abortPromise(signal),
]);
```

When there is no database update before `ms`, the timer resolves but `updateEvents.next()` continues waiting. Scheduler and claim-waker loops then call the helper again.

The native implementation in `packages/honker-node/src/lib.rs` uses a blocking receive:

```rust
tokio::task::spawn_blocking(move || {
    let r = rx.lock();
    r.recv().map_err(napi_err)
})
```

Because the receiver is mutex-protected, the first call blocks in `recv()` and subsequent calls block waiting for the mutex, consuming Tokio's blocking pool until an update or close releases them.

## Deterministic reproduction

At the JavaScript API seam, a fake update source makes the accumulation visible without relying on `/proc`:

```js
const buildApi = require('./packages/honker-node/api');
const { ClaimWaker } = buildApi({});

let calls = 0;
const queue = {
  _db: {
    updateEvents: () => ({
      next: () => {
        calls += 1;
        return new Promise(() => {});
      },
      close() {},
    }),
  },
  claimOne: () => null,
  _nextClaimAt: () => null,
};

const controller = new AbortController();
setTimeout(() => controller.abort(), 50);
await new ClaimWaker(queue, { idlePollS: 0.001 }).next('worker', {
  signal: controller.signal,
});
console.log(calls); // dozens of outstanding next() calls; expected: 1
```

Against the native binding on Linux, a 1 ms scheduler loop reproduced 107 additional OS tasks in roughly 0.5 seconds. Reusing the single outstanding update wait reduced growth to two tasks.

## Expected behavior

Each `UpdateEvents` instance should have at most one outstanding native `next()` call from the timeout helper. If a timeout or abort wins, a later iteration should reuse that pending wait. Once it settles, the following iteration can create a new wait.

I have a regression test and a focused patch ready and will open a linked PR.

lklynet:
## Summary

- reuse one pending `UpdateEvents.next()` promise per update source when a timeout wins the race
- remove the cached promise after the native wait settles so the next database update is still observed normally
- add a deterministic public-API regression test and include it in the Node CI job

## Why

`Promise.race()` does not cancel its losing promises. Each poll timeout previously left a native `UpdateEvents.next()` call blocked in Tokio and started another on the next iteration. Long-running schedulers and claim wakers could therefore fill Tokio's blocking pool and retain hundreds of OS threads.

This is separate from the losing `abortPromise()` listener reported in #67; this PR only addresses the native update wait.

Closes #68.

## Validation

- Regression before the fix: 43 outstanding `next()` calls after 50 ms
- Regression after the fix: one outstanding call, stable across repeated runs
- `npm run build`
- Node/native suite: 60 passed, 0 failed, 16 feature-gated skips

The aggregate `npm test` command was also attempted locally; its Python interop cases require `.venv/bin/python`, which was not present in the clean checkout. The Node/native portions above completed successfully, and the repository CI provisions the required Python environment for the cross-language jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -297,7 +297,7 @@ jobs:
       - name: Test (basic — node only)
         working-directory: packages/honker-node
         shell: bash
-        run: node --test test/basic.js
+        run: node --test test/api.test.js test/basic.js
       - name: Test (parity — node API surface)
         working-directory: packages/honker-node
         shell: bash
--- a/packages/honker-node/api.js
+++ b/packages/honker-node/api.js
@@ -43,18 +43,34 @@ function abortPromise(signal) {
   });
 }
 
+const pendingUpdateWaits = new WeakMap();
+
+function nextUpdate(updateEvents) {
+  const existing = pendingUpdateWaits.get(updateEvents);
+  if (existing) return existing;
+
+  const pending = updateEvents.next().catch(() => undefined);
+  pendingUpdateWaits.set(updateEvents, pending);
+  void pending.finally(() => {
+    if (pendingUpdateWaits.get(updateEvents) === pending) {
+      pendingUpdateWaits.delete(updateEvents);
+    }
+  });
+  return pending;
+}
+
 async function waitForUpdateOrTimeout(updateEvents, signal, timeoutMs) {
   if (aborted(signal)) return;
   if (timeoutMs == null) {
     await Promise.race([
-      updateEvents.next().catch(() => undefined),
+      nextUpdate(updateEvents),
       abortPromise(signal),
     ]);
     return;
   }
   const ms = Math.max(0, timeoutMs);
   await Promise.race([
-    updateEvents.next().catch(() => undefined),
+    nextUpdate(updateEvents),
     delay(ms),
     abortPromise(signal),
   ]);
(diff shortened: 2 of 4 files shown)


russellromney:
Nice

reviewing

PR #14, opened 2026-04-30
fix: reopen Windows package coverage

russellromney:
## Summary
- add explicit close paths for Python and Node database/watcher handles
- fix Node cross-language tests to find the venv Python on Windows
- reopen Windows CI coverage for `rust-extension`, `python`, and `node`

## Verification
- `maturin develop --release` in `packages/honker`
- `python -m pytest tests/test_litenotify.py::test_listener_drop_allows_clean_reuse tests/test_resource_bounds.py::test_many_simultaneous_listeners_bounded_thread_count tests/test_extension_interop.py::test_extension_and_python_share_schema -q --tb=short`
- `node --expose-gc --test test/basic.js test/cross_lang.js`
- `node --expose-gc --test test/parity.test.js`

PR #78, opened 2026-07-27
ci: add lint gate, dependabot, aggregate status check, lockfile pinning

jjshanks:
Four gaps from a CI review, plus the fallout from turning clippy on.

Dependabot (.github/dependabot.yml, new)
  release-crates.yml's comment promised "Dependabot will open PRs" for
  the SHA-pinned actions, but no config existed. Pinning without an
  upgrade prompt just means sitting on known-bad versions quietly.
  Covers github-actions and cargo weekly, the binding ecosystems
  (npm, bun, nuget, bundler, mix, gomod, maven) monthly, grouped so a
  quiet week is one PR rather than nine. honker-jvm/honker-kotlin have
  no CI job, so this is currently the only thing watching them.

Lint job (ci.yml, Cargo.toml, Makefile)
  Nothing ran cargo fmt or clippy. New `lint` job runs both, with a
  second clippy pass over the cfg-gated kernel-watcher/shm-fast-path
  code the default-feature pass never type-checks. Lint policy lives in
  the root [workspace.lints.clippy] table so `make lint` and CI agree.

  Turning it on surfaced 11 findings. Nine were mechanical and are
  fixed here: redundant u32 casts in cron.rs, io::Error::new(Other, _)
  -> io::Error::other, a single-arm match, an explicit copy closure, a
  collapsible if, and a manual Result::unwrap_or in the extension's
  watcher_wait. Two are design opinions (too_many_arguments on the
  published enqueue signature, type_complexity on two SQL row tuples)
  and are allowed at workspace level with rationale — refactoring a
  published API belongs in a deliberate 0.3, not a lint sweep.

Aggregate status check (ci.yml)
  The python and node matrices are built with fromJson and run a
  smaller set on pull_request than on push, so job names differ by
  event and no hand-maintained required-checks list is correct for
  both. New `ci-ok` job has one stable name, depends on every real job,
  and asserts each result is success — treating skipped as failure
  except for perf-regression, which is gated off on PRs by design.
  `if: always()` is load-bearing: a skipped required check reads as a
  pass to branch protection.

Lockfile pinning
  package-lock.json and bun.lock were gitignored, so `npm install`
  floated @napi-rs/cli ^3.0.0 and `bun install --frozen-lockfile` was
  pinning against a file that never existed. Both lockfiles are now
  committed and the node job uses `npm ci`. Every cargo invocation
  across CI and the release workflows gained --locked, including
  `cargo publish --locked` — Cargo.lock was committed but nothing
  enforced it. Only the three crates with a committed lock are
  affected; packages/honker and packages/honker-node still resolve
  freely.

Also: `apt-get update` before the C++ dep install, and an adhoc-packages suppression for release-ruby.yml, whose two flagged `gem install` calls install the locally built gem under test — there is no registry to pin against.

Verified: make lint, cargo test --locked (default and experimental features), cargo build --locked, npm ci, bun install --frozen-lockfile, and zizmor --min-severity low across all workflows. Cargo.lock is unchanged, confirming --locked is safe to add.

Claude-Session: https://claude.ai/code/session_0148fYm2uL3rwzAPU1Npnh6z


jjshanks:
@russellromney I don't have a mac so can't verify a fix for the flaky test. Filed https://github.com/russellromney/honker/issues/80 to address it.

jjshanks:
### Code review

Reviewed at `bbc4415` across five passes (diff scan, git history/blame, prior-PR guidance, CI correctness, workflow security). Found 3 high-severity issues. All three are fixed in `ed161cf`, pushed to this branch.

**1. `ci-ok`'s branch-protection instruction silently un-gates zizmor** (flagged independently by three passes)

"Point branch protection at this job and nothing else" drops `zizmor` from required checks. zizmor is a separate workflow and `needs:` cannot reach across workflows, so `ci-ok` cannot subsume it — while zizmor.yml's own header states "Failing this job blocks the PR". The gate lost is the one covering the release workflows that hold `CARGO_REGISTRY_TOKEN`.

https://github.com/russellromney/honker/blob/bbc441502d143fce7cc3a92ab67537cd1aa87b8f/.github/workflows/ci.yml#L458-L463

It also could not simply be added to protection as-is: its `pull_request` trigger was path-filtered to `.github/**`, and a required check that never reports pins the PR in "Expected — waiting for status" instead of passing.

https://github.com/russellromney/honker/blob/bbc441502d143fce7cc3a92ab67537cd1aa87b8f/.github/workflows/zizmor.yml#L12-L17

Fixed: the comment now names both required checks, and zizmor's PR path filter is removed so it always reports (the `push` filter is kept).

**2. The new `xfail` is broader than the flake it documents**

A non-strict `xfail` on the whole `"kernel"` param swallows every failure mode on Windows, not just missed wakes — including the runaway-watcher upper bound, which the `min_wakes = 1` fudge it replaces *did* still enforce, and any exception out of `open()` / `update_events()`. The replacement proves less than the fudge it removes.

https://github.com/russellromney/honker/blob/bbc441502d143fce7cc3a92ab67537cd1aa87b8f/tests/test_watcher_backends.py#L85-L92

Fixed: upper bound is checked first and on every platform; `pytest.xfail()` is raised imperatively only when the kernel backend under-delivers on win32. A Windows run that does deliver now reports a plain pass rather than xpass noise.

**3. `ignore_cleanup_errors=True` retires a regression detector**

It is unconditional, so it also swallows WinError 32 on *passing* tests — the signal that honker-core's explicit `close()` still releases SQLite handles in time. Treating this in the harness was already rejected once: #14 was closed as superseded by #18, with "Test-harness retry loops (this PR's approach) treated the symptom; the underlying handles still leaked." The fixture's own comment only justifies the flag for *failing* tests.

https://github.com/russellromney/honker/blob/bbc441502d143fce7cc3a92ab67537cd1aa87b8f/tests/conftest.py#L18-L27

Fixed: cleanup is strict again; errors are tolerated only when the test actually failed.

Also fixed, all confirmed defects in lines this PR adds: `make help` ran command substitution on the backticked `lint` and printed `/bin/sh: 1: lint: not found` ([Makefile#L21-L23](https://github.com/russellromney/honker/blob/bbc441502d143fce7cc3a92ab67537cd1aa87b8f/Makefile#L21-L23)); `Cargo.toml` says "three allows" with two defined ([Cargo.toml#L22-L24](https://github.com/russellromney/honker/blob/bbc441502d143fce7cc3a92ab67537cd1aa87b8f/Cargo.toml#L22-L24)); and the release runbook's `git commit -am` leaves `Cargo.lock` stale, which `--locked` now converts into a release failure *after* the tag is pushed.

Verified clean: `npm ci` (exit 0, including win32/darwin/arm64 dry-runs), `bun install --frozen-lockfile`, all three `--locked` manifests have tracked locks and pass `cargo metadata --locked`, `cargo package --locked` handles the new workspace lints, `dependabot.yml` validates against the SchemaStore 2.0 schema with 0 errors and all 11 directories hold a real manifest, `ci-ok`'s `needs` covers all seven jobs (and its skip logic is correct against simulated `needs` payloads), and all six clippy rewrites are behavior-preserving.

Not fixed, left for your call:

- `ci-ok`'s `needs` is hand-maintained: a newly added job that isn't listed goes red while `ci-ok` stays green. That list has churned five times in about three months, and this PR adds to it.
- `packages/honker-node/test/watcher_backends.js` keeps the identical `minWakes = 1` Windows fudge, so Node and Python now assert different contracts for the same backend on the same platform.
- `npm ci` exits 0 when only `optionalDependencies` drift, silently installing none of the native packages — so every version bump needs a lockfile regen or the published binary stops being exercised.
- `scripts/proof/package-install-smoke.sh` still runs `npm install` against the now-tracked lockfile, in the real checkout.
- `cargo run --locked --release | tee` masks cargo's exit status (no `pipefail`).
- No `uv`/`pip` dependabot entry for `packages/honker/pyproject.toml` + `uv.lock`.
- `make lint-fix` omits the feature-gated clippy pass that `make lint` runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #123, opened 2026-08-27
Add transactional queue lifecycle events


russellromney:
## Problem

Honker doesn't currently expose a cross-process feed of queue/job lifecycle activity. This makes it difficult to build live dashboards, metrics adapters, or debugging tools that observe enqueue, claim, completion, retry, and failure activity across workers and language bindings.

## Proposal

Add an opt-in `QueueEvents` API backed by a bounded Honker Stream.

Queue events should be appended by `honker-core` in the same SQLite transaction as the queue state transition. A rolled-back or unsuccessful transition must not produce an event.

Initial event types:

- `enqueued`
- `claimed`
- `completed`
- `retry_scheduled`
- `dead_lettered`
- `cancelled`

Each event should include a version, stream offset, event type, queue, job ID, timestamp, attempt count, and relevant worker/error/run-at metadata. Job payloads should be excluded by default.

The feed should support:

- Global or per-queue consumption
- Resuming from an offset within the retention window
- Bounded retention
- Cross-process and cross-language visibility
- Persistent opt-in configuration so all producers and workers use consistent behavior

## Semantics

- Events describe successful committed state transitions, not method calls.
- Stale/failed `ack`, `retry`, etc. do not emit events.
- Consumers may deduplicate by offset.
- Events are intended for observability and may have gaps after retention trimming.
- Queue correctness must never depend on an event consumer.

## Non-goals

- A permanent audit log
- Exactly-once event delivery
- Business-critical workflow triggers
- Heartbeat, progress, or arbitrary custom events
- A new event subsystem separate from Honker Streams

Core schema, event semantics, transactional emission, and retention should land centrally. Thin binding adapters and documentation can follow as binding-parity issues.

russellromney:
Updated status: #124 and #125 are separate again. #124 contains typed jobs. The `feat/node-queue-events` branch still contains the core and Node queue event work from #125.

GitHub still marks #125 as merged because of the earlier stacked merge, but the event commits are not in `main`. After #124 lands, the event branch needs a new PR to `main` or another clear merge path.

Keep this issue open until the event code reaches `main`. Cross-binding event APIs are tracked in #137.

PR #124, opened 2026-08-27
feat(node): add typed job payloads and details

### LINKED ISSUE #119: honker-node missing types and ergonomic improvements
Hello and thank you for the work on #99 , that cleared up everything for me and I'm trying out the honker-node package now 🎉 

I have a few pieces of feedback and some questions with what I've used so far:

### missing types  

The [`Queue` wrapper type](https://github.com/russellromney/honker/blob/main/packages/honker-node/wrapper.d.ts#L162) is missing types for [`getJob`](https://github.com/russellromney/honker/blob/main/packages/honker-node/api.js#L449) and [`cancel`](https://github.com/russellromney/honker/blob/main/packages/honker-node/api.js#L443).

### additional properties?

Using `queue.getJob` exposes more properties that I think would be helpful to add to `Job` with regard to state discovery. Having `run_at`, `priority`, `created_at`, and `state` be *officially* available in the Job class would be great. A user inspecting a job would definitely want to know how long a job has been running/queued, when its due to run (if delayed), and if its currently running or not.

### convenience generic type for `payload`

It would be nice if `Queue` and/or `Job` could get a generic type representing the type of `payload` so that we can get strongly typed job payloads out of the box. Right now, payload is always `JsonValue` and to type it for some domain a user, minimally, has to do

```ts
import type {Job} from '@russellthehippo/honker-node';

type HonkerJob<T> = Omit<Job, 'payload'> & {payload: T};

// ... enqueue job etc...
const waker = q.claimWaker();
const job = await waker.next('worker-1') as unknown as HonkerJob<MyDomainObj>;
```

To type this I think it could be done solely by modifying `wrapper.d.ts`:

```ts
export class Job<T = JsonValue> {
 // ...
  readonly payload: T
}

export class Queue<T = JsonValue> {
 // ...
  enqueue(payload: T, opts?: EnqueueOptions): number
  enqueueTx(tx: Transaction | any, payload: T, opts?: EnqueueOptions): number
  claimBatch(workerId: string, n: number): Job<T>[]
  claimOne(workerId: string): Job<T> | null
  claim(workerId: string, opts?: { idlePollS?: number | null, signal?: AbortSignal }): AsyncIterableIterator<Job<T>>
}

export class Database {
// ...
queue<T = JsonValue>(name: string, opts?: QueueOptions): Queue<T>
}
```

### events on queue/job activity

This might be out of scope but I'm wondering if there is a way to go about getting events on queue activity. For example, I'd like to be able to monitor, globally, when a new job is added with `q.enqueue`, or when a `job.ack()` or `job.retry()` is fired. I show realtime queue activity to my users, now, based on listening to some event emitters and I'm hoping there is an analogue for honker.

Thanks for your time and effort on this library!

russellromney:
## Summary

- Add a TypeScript payload type to Node queues, jobs, claim wakers, and outboxes. It defaults to `JsonValue` so existing code keeps working.
- Add typed camelCase job snapshots from `getJob()`.
- Add the missing `getJob()` and `cancel()` type declarations.
- Return full job details from the core claim call.
- Make `getJob()` return null for a job from another queue.
- Add docs and TypeScript compile tests.
- Fix stale native-loader version messages and check those messages in CI.

## Behavior changes

`getJob()` now maps the core snake_case row to the documented camelCase `JobSnapshot` shape. `CHANGELOG.md` calls this out because the runtime method existed before its TypeScript declaration.

`getJob()` now returns null when the job belongs to another queue. Job IDs are unique across the database, so the old ID-only lookup could return an SMS payload from `emails.getJob(smsId)`. That would break the new `JobSnapshot<TPayload>` type. The check stays in the Node wrapper. `honker_get_job(?)` does not change.

`cancel()` is still global and the docs now say so. Reading the job first and cancelling in a second query would have a race. #134 tracks the core and binding changes needed to fix that safely.

## Scope

Queue events are separate work in #125. They are not part of this PR. #123 tracks the core and Node event work. #137 tracks event APIs for other bindings.

## Verification

- `cargo test -p honker-core --release`
- `npm test` in `packages/honker-node` - 116 tests, 114 passed, 2 Windows-only skips
- Full Python test suite
- `make lint`
- Separate-process Node producer, `getJob()` reader, and consumer/ack test
- Strict TypeScript public API compile test
- Cross-queue `getJob()` regression test
- Node version check now covers both loader guards and their error messages

## Follow-up issues

- #133 - Return SQLite errors instead of hiding them. `fail()` can otherwise lose a damaged job row.
- #134 - Make global database methods and queue-scoped methods consistent.
- #135 - Add a stable `claimed_at` value.
- #136 - Add full job details and typed payloads to every binding.

Closes #119

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -428,6 +428,9 @@ jobs:
         run: |
           npm ci
           npx napi build --platform --release --features kernel-watcher,shm-fast-path
+      - name: Test Node TypeScript declarations
+        working-directory: packages/honker-node
+        run: npm run test:types
       - name: Test Node ↔ Python stream checkpoint journeys
         shell: bash
         run: |
@@ -466,7 +469,7 @@ jobs:
       - name: Test (basic — node only)
         working-directory: packages/honker-node
         shell: bash
-        run: node --test test/api.test.js test/basic.js
+        run: node --test test/api.test.js test/basic.js test/typed_jobs_e2e.js
       - name: Test (parity — node API surface)
         working-directory: packages/honker-node
         shell: bash
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,19 @@
 - Regression test claims a job at a deadline with one claim forced to
   return empty: it fails on the unpatched `api.js` and passes on the fix.
 
+## Unreleased — typed Node jobs
+
+- Node `Queue`, `Job`, `JobSnapshot`, `ClaimWaker`, and `Outbox` APIs now carry
+  a payload generic that defaults to `JsonValue`, preserving existing callers
+  while allowing application-specific payload contracts.
+- `Queue.getJob()` and `Queue.cancel()` are included in the public TypeScript
+  declarations. Claimed jobs and job snapshots expose state, priority,
+  scheduling, attempt, worker, creation, claim-expiry, and job-expiry details.
+- The Rust claim ABI now returns the complete live-job snapshot, so claimed
+  jobs provide the same operational details as `getJob()`.
+- **Behavior change:** Node `getJob()` now returns the documented camelCase
+  `JobSnapshot` shape instead of exposing the SQL ABI's raw snake_case row.
+
 ## Unreleased — Node checkpoint interoperability (Phase Robinson)
 
 - Node stream checkpoint calls now use the shared SQL ABI's canonical
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,12 +30,25 @@
   scheduling, attempt, worker, creation, claim-expiry, and job-expiry details.
 - The Rust claim ABI now returns the complete live-job snapshot, so claimed
   jobs provide the same operational details as `getJob()`.
+- Every job field carries TSDoc. All four timestamps (`runAt`, `createdAt`,
+  `claimExpiresAt`, `expiresAt`) are Unix epoch seconds, not milliseconds, and
+  now say so on hover. `Job` declares `implements JobSnapshot<TPayload>`, so a
+  field added to one shape and not the other fails the type test.
+- `Job.claimExpiresAt` is `number`, not `number | null`. A claimed job always
+  holds a lease, so that null branch could never be taken. `JobSnapshot` keeps
+  `number | null` for pending jobs.
 - **Behavior change:** Node `getJob()` now returns the documented camelCase
   `JobSnapshot` shape instead of exposing the SQL ABI's raw snake_case row.
-  It also returns `null` for a job that belongs to a different queue, instead
-  of returning that queue's row. Job ids are globally unique, so the previous
-  unscoped lookup could hand back a payload that did not match the queue's
-  declared `TPayload`. `cancel()` is unchanged and remains not queue-scoped.
+  `row.run_at` becomes `job.runAt` and reads back as `undefined` under the old
+  name, and `payload` arrives already JSON-decoded, so an existing
+  `JSON.parse(row.payload)` must be dropped.
+- **Behavior change:** Node `getJob()` returns `null` for a job that belongs to
+  a different queue, instead of returning that queue's row. Job ids are globally
+  unique, so the previous unscoped lookup could hand back a payload that did not
+  match the queue's declared `TPayload`. There is no unscoped replacement yet
+  (#134); callers that relied on the global lookup can run
+  `SELECT honker_get_job(?)` on their own connection, which still returns the
+  raw snake_case row. `cancel()` is unchanged and remains not queue-scoped.
 
 ## Unreleased — Node checkpoint interoperability (Phase Robinson)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@
   jobs provide the same operational details as `getJob()`.
 - **Behavior change:** Node `getJob()` now returns the documented camelCase
   `JobSnapshot` shape instead of exposing the SQL ABI's raw snake_case row.
+  It also returns `null` for a job that belongs to a different queue, instead
+  of returning that queue's row. Job ids are globally unique, so the previous
+  unscoped lookup could hand back a payload that did not match the queue's
+  declared `TPayload`. `cancel()` is unchanged and remains not queue-scoped.
 
 ## Unreleased — Node checkpoint interoperability (Phase Robinson)
 
--- a/honker-core/src/honker_ops.rs
+++ b/honker-core/src/honker_ops.rs
@@ -648,7 +648,7 @@ fn dead_letter_exhausted_claimable(conn: &Connection, queue: &str) -> rusqlite::
     Ok(count)
 }
 
-/// Returns JSON text: `[{"id":1,"queue":"...","payload":"...","worker_id":"...","attempts":N,"claim_expires_at":T}, ...]`
+/// Returns JSON text containing the complete claimed-job snapshot.
 pub fn claim_batch(
     conn: &Connection,
     queue: &str,
@@ -679,8 +679,10 @@ pub fn claim_batch(
            ORDER BY priority DESC, run_at ASC, id ASC
            LIMIT ?3
          )
-         RETURNING id, queue, payload, worker_id, attempts, claim_expires_at",
+         RETURNING id, queue, payload, state, priority, run_at, worker_id,
+                   claim_expires_at, attempts, max_attempts, created_at, expires_at",
     )?;
+    #[allow(clippy::type_complexity)]
     let rows = stmt.query_map(rusqlite::params![worker_id, queue, n, timeout_s], |row| {
         Ok((
             row.get::<_, i64>(0)?,
@@ -689,20 +691,45 @@ pub fn claim_batch(
             row.get::<_, String>(3)?,
             row.get::<_, i64>(4)?,
             row.get::<_, i64>(5)?,
+            row.get::<_, String>(6)?,
+            row.get::<_, i64>(7)?,
+            row.get::<_, i64>(8)?,
+            row.get::<_, i64>(9)?,
+            row.get::<_, i64>(10)?,
+            row.get::<_, Option<i64>>(11)?,
         ))
     })?;
     let mut out = Vec::new();
     for row in rows {
-        let (id, q, payload, w, attempts, claim_expires_at) = row?;
+        let (
+            id,
+            q,
+            payload,
+            state,
+            priority,
+            run_at,
+            w,
+            claim_expires_at,
+            attempts,
+            max_attempts,
+            created_at,
+            expires_at,
+        ) = row?;
         // payload stays a JSON string (double-encoded on the wire) so
         // every binding's existing parse path keeps working.
         out.push(json!({
             "id": id,
             "queue": q,
             "payload": payload,
+            "state": state,
+            "priority": priority,
+            "run_at": run_at,
             "worker_id": w,
             "attempts": attempts,
             "claim_expires_at": claim_expires_at,
+            "max_attempts": max_attempts,
+            "created_at": created_at,
+            "expires_at": expires_at,
         }));
     }
     Ok(Value::Array(out).to_string())
@@ -1738,6 +1765,7 @@ fn now_unix(conn: &Connection) -> rusqlite::Result<i64> {
 mod real_arg_tests {
     use crate::{attach_honker_functions, bootstrap_honker_schema};
     use rusqlite::Connection;
+    use serde_json::Value;
 
     fn db() -> Connection {
         let conn = Connection::open_in_memory().unwrap();
@@ -1761,6 +1789,42 @@ mod real_arg_tests {
             )
             .unwrap();
         assert!(id > 0);
+    }
+
+    #[test]
+    fn claim_batch_returns_the_complete_job_snapshot() {
+        let conn = db();
+        let id: i64 = conn
+            .query_row(
+                "SELECT honker_enqueue('emails', '{\"to\":\"alice@example.com\"}',
+                                       NULL, NULL, 7, 5, 600)",
+                [],
+                |r| r.get(0),
+            )
+            .unwrap();
+
+        let claimed: String = conn
+            .query_row(
+                "SELECT honker_claim_batch('emails', 'worker-1', 1, 300)",
+                [],
+                |r| r.get(0),
+            )
+            .unwrap();
+        let rows: Value = serde_json::from_str(&claimed).unwrap();
+        let job = &rows[0];
+
+        assert_eq!(job["id"], id);
+        assert_eq!(job["queue"], "emails");
+        assert_eq!(job["payload"], r#"{"to":"alice@example.com"}"#);
+        assert_eq!(job["state"], "processing");
+        assert_eq!(job["priority"], 7);
+        assert_eq!(job["worker_id"], "worker-1");
+        assert_eq!(job["attempts"], 1);
+        assert_eq!(job["max_attempts"], 5);
+        assert!(job["run_at"].as_i64().unwrap() > 0);
+        assert!(job["claim_expires_at"].as_i64().unwrap() > 0);
+        assert!(job["created_at"].as_i64().unwrap() > 0);
+        assert!(job["expires_at"].as_i64().unwrap() > 0);
     }
 
     #[test]
--- a/packages/honker-node/README.md
+++ b/packages/honker-node/README.md
@@ -62,6 +62,28 @@ Delayed jobs use `runAt`:
 q.enqueue({ to: "later@example.com" }, { runAt: Math.floor(Date.now() / 1000) + 10 });
 ```
 
+TypeScript callers can give a queue a payload contract. Claimed jobs and
+`getJob()` snapshots preserve it:
+
+```ts
+interface EmailPayload {
+  to: string;
+  template: "welcome" | "receipt";
+}
+
+const emails = db.queue<EmailPayload>("emails");
+emails.enqueue({ to: "alice@example.com", template: "welcome" });
+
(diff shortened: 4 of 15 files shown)


russellromney:
Accidentally merged #125 into this. Splitting back up

FoxxMD (approved):
LGTM!

Everything is in here I'd expect there to be in terms of job details and accessibility.

I also overwrote my own `wrapper.d.ts` with the PR's and the typing for queue/job works as expected.

Send it 🚀

russellromney:
## Adversarial review — #124

Every finding below is labelled **RAN** (I executed it) or **INFERRED** (read from source, not executed).

### 1. The test suite cannot tell `run_at` from `created_at` on a claimed job — **RAN**

Mutating the `Job` constructor in `api.js` to

```js
this.runAt = row.created_at;
this.createdAt = row.run_at;
```

leaves **53/53 tests passing**. The same swap in the `getJob` snapshot mapping also passes 53/53.

Cause: every test enqueues with no delay, so `run_at == created_at` to the second, and `typed_jobs_e2e` only asserts `claimed.runAt === snapshot.runAt` — two values that are equal under the swap. The Rust test `claim_batch_returns_the_complete_job_snapshot` asserts only `> 0` for both (INFERRED — read, not run).

Control: deleting `if (row.queue !== this.name) return null;` **does** fail a test, so the queue-scoping test is load-bearing. The gap is specific to these two fields.

This matters because #136's own checklist requires "a delayed job reports the correct `run_at`", and #124 has no delayed-job test. If a mapping swap ships, every consumer's "job due at" silently reads back as the creation time and CI stays green.

**Fix — one line.** A delayed job is not claimable, so the claimed path needs a *backdated* `run_at`:

```js
const id = q.enqueue(payload, { runAt: nowSeconds - 100 });
// still claimable, and run_at !== created_at
const job = q.claimOne('w');
assert.notStrictEqual(job.runAt, job.createdAt);
assert.strictEqual(job.runAt, snapshot.runAt);
```

Verified equivalent in Python, Ruby and Elixir (RAN): `run_at = now - 100` yields a 100s gap and the job is still claimable.

### 2. `cancel` is global while `getJob` is scoped — you can destroy a job you cannot read — **RAN**

```js
emails.cancel(smsJobId)   // => true, the SMS row is deleted
emails.getJob(smsJobId)   // => null, both before and after
```

The JSDoc notes the asymmetry, but the consequence is that the natural defensive idiom `if (q.getJob(id)) q.cancel(id)` now silently never fires for a foreign id, while a bare `q.cancel(id)` still deletes it. Scoping is also applied in JS *after* the core returns the whole row, so it is an ergonomic filter, not a boundary.

### 3. Scoping removes the only by-id lookup, with no replacement — **RAN**

There is no `Database.getJob` in `api.js`. Before this PR `db.queue(anything).getJob(id)` was a working global lookup. Admin/inspection code that relied on it now gets `null` — no error, no deprecation. #134 tracks the DB-level API but it does not exist yet.

### 4. The declared type is wrong in three ways — **RAN**

- Non-JSON payload (raw-SQL enqueue): `typeof snapshot.payload === 'string'`, and the same for `claimOne()`. (Tracked.)
- **Not tracked:** a JSON `null` payload yields `payload === null` while `JobSnapshot<T>.payload: T` is non-nullable. Reachable with no raw SQL at all — any producer, including another binding, can write it.
- The comment "scoping the lookup to this queue keeps the payload type honest" does not hold for the same queue name: `db.queue<A>('shared')` and `db.queue<B>('shared')` still hand each other's payloads across.

Also note `JobSnapshot.payload` is **decoded** in Node while Python's and Ruby's snapshot payload is **raw JSON text** (RAN for Python/Ruby). #124 introduces this divergence: same field name, different type per binding. It is documented in the Ruby/Elixir PRs but not here.

### 5. Silent break on a published package — low

`getJob` went from a raw snake_case row with a string payload to camelCase with a decoded payload. `row.run_at` becomes `undefined` with no throw. The CHANGELOG documents it; `package.json` stays at 0.5.1.

### What holds up

`tsc` passes and the `@ts-expect-error` line makes the type test non-vacuous by construction. Dead-lettered / acked / never-existed all return `null`. `workerId` and `claimExpiresAt` are correctly `null` rather than `0`/`""`. `attempts` vs `maxAttempts` and `expiresAt` vs `claimExpiresAt` are both genuinely discriminated.

**Recommendation:** hold for finding 1 (it is the base of a nine-PR stack, and the same hole is now mutation-proven in #143, #148 and #149) and finding 3 (a capability removed with no replacement).

russellromney:
## Finding 1 fixed — the `run_at` / `created_at` gap is now load-bearing

Commit: `518d3c0` — `test(node): prove a claimed job reports run_at, not created_at`

Added `a claimed job reports its own run_at, not its created_at` to `test/typed_jobs_e2e.js`. It enqueues with a **back-dated absolute `runAt`** (`now - 100`) rather than a delay, so the job stays claimable while `run_at != created_at`, and it pins `runAt` to the exact value it was enqueued with instead of comparing it to another field.

### Mutation proof

Both mapping sites were mutated, run, and restored. In each case **only the new test fails** — the other two tests in the file stay green, which is exactly the blind spot you reported.

**Mutation A — claimed `Job` constructor (`api.js:265/270`)**

```js
this.runAt = row.created_at;
this.createdAt = row.run_at;
```

```
✖ a claimed job reports its own run_at, not its created_at
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  + actual - expected

  + 1788318131
  - 1788318031
           ^
      at TestContext.<anonymous> (test/typed_jobs_e2e.js:141:12)
    operator: 'strictEqual'

ℹ tests 3 / pass 2 / fail 1
```

Line 141 is `assert.equal(job.runAt, runAt)`.

**Mutation B — `getJob` snapshot mapping (`api.js:486/491`)**

```js
runAt: row.created_at,
createdAt: row.run_at,
```

```
✖ a claimed job reports its own run_at, not its created_at
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  + actual - expected

  + 1788318131
  - 1788318031
           ^
      at TestContext.<anonymous> (test/typed_jobs_e2e.js:136:12)
    operator: 'strictEqual'
```

Line 136 is `assert.equal(snapshot.runAt, runAt)`.

Both sites restored; `git diff` on `api.js` is empty. No API, field name, or payload encoding changed — this is test coverage only.

### Local run

`tsc` green. `node --test` over `extension / api / basic / parity / watcher_backends / phase_mantle / typed_jobs_e2e`: **83 tests, 81 pass, 0 fail, 2 skipped** (the 2 are `watcherBackend=kernel` and `=shm`, which need Cargo features not built locally). The `cross_lang_*` files need a Python honker install this worktree does not have, so I left those to CI.

Findings 2, 3, 4 and 5 are untouched — this commit is scoped to finding 1.

russellromney:
## Adversarial review round 2 — #124

Deeper pass than round 1: is this complete, well designed, bug-free, comprehensively
tested, and does it make sense to a user? Every claim below is **RAN** (executed) or
**READ** (from source, not executed).

Pushed as `7360d38` on `feat/node-typed-jobs`. No API shape changed except one
nullability narrowing (below). Core untouched.

---

### No defect in the core RETURNING widening — READ

I went looking, since nine PRs sit on it. It holds up:

- The 12-column `RETURNING` order matches the 12-element tuple destructure exactly.
- Nullability is right against `BOOTSTRAP_HONKER_SQL`: `worker_id` and
  `claim_expires_at` are read as non-`Option` **only because the same statement
  `SET`s them**; `expires_at` is the one genuinely nullable column and is the one
  `Option<i64>`.
- `RETURNING` on an `UPDATE` yields post-update values, so `state` is always
  `'processing'` and `attempts` is already incremented. `Job.state: 'processing'`
  and "`attempts` ≥ 1" are honest.
- `_honker_live` only ever holds `'pending'`/`'processing'` (acked and dead rows
  leave the table), so `JobState = 'pending' | 'processing'` is complete.
- Extra keys are inert in every other binding: no `deny_unknown_fields` anywhere in
  the tree, and the consumers pass the whole row object through
  (`Job(self, row)` in Python, etc.) rather than splatting it into kwargs.

**No change to `honker-core`.**

---

## Findings

### 1. Twelve new fields, four of them timestamps, zero documentation on hover — MEDIUM, fixed

This is the finding I'd hold the PR on. #119 asked for job state to be *"officially
available"*; what shipped is twelve bare `readonly` lines. `wrapper.d.ts` and the
README contained **zero** occurrences of "unix", "epoch", or "seconds" — RAN
(`grep -n "unix\|epoch\|seconds" README.md wrapper.d.ts` → no output).

`runAt`, `createdAt`, `claimExpiresAt`, and `expiresAt` are Unix epoch **seconds**.
A JS user's reflex is `new Date(job.createdAt)`, which lands in 1970 and never
throws. The README enqueue example writes `Math.floor(Date.now() / 1000)` so the
convention is inferable — but only if you read the enqueue example, not the field.

Fixed: TSDoc on all twelve fields of both `JobSnapshot` and `Job`, plus a field table
in the README that calls out the unit and points at `new Date(job.createdAt * 1000)`.

### 2. `Job` / `JobSnapshot` consistency was unenforced — MEDIUM, fixed

The two shapes are meant to be the same twelve fields, one with methods. Nothing
checked that. Add a field to one and not the other and everything stays green.

Fixed two ways:

- `export class Job<TPayload> implements JobSnapshot<TPayload>` in the `.d.ts`. This
  is checked (`skipLibCheck` is off in `test/types/tsconfig.json`).
- A runtime assertion that the claimed `Job`'s own non-underscore keys and
  `Object.keys(snapshot)` both equal the same explicit twelve-name list.

**Mutation proof — RAN.** Deleted `readonly createdAt: number` from the `Job` class,
ran `tsc -p test/types/tsconfig.json`:

```
test/types/typed-jobs.ts(54,26): error TS2345: Argument of type 'Job<EmailPayload>' is not
  assignable to parameter of type 'JobSnapshot<EmailPayload>'.
  Property 'createdAt' is missing in type 'Job<EmailPayload>' but required in type 'JobSnapshot<EmailPayload>'.
wrapper.d.ts(177,14): error TS2420: Class 'Job<TPayload>' incorrectly implements interface 'JobSnapshot<TPayload>'.
  Property 'createdAt' is missing in type 'Job<TPayload>' but required in type 'JobSnapshot<TPayload>'.
EXIT=2
```

Restored from a scratchpad copy; `diff` against the backup is clean (no
`git checkout --`).

This also answers a question round 1 left open — *is `Job<T>` vs `JobSnapshot<T>` a
distinction a reader will understand?* The relationship is now stated in one line at
the top of each doc block ("a claimed job is a snapshot you hold a claim on") and
enforced by the compiler, and `describe(job: JobSnapshot<T>)` accepting a `Job<T>` is
a compile test.

### 3. `Job.claimExpiresAt` was declared nullable but can never be null — LOW, fixed (type narrowing)

`workerId` was correctly narrowed to `string` on the claimed `Job`; `claimExpiresAt`
was left `number | null`. It is set by the same `UPDATE` and read as a non-`Option`
`i64`, so on a claimed job it is always a number. Every worker doing lease
arithmetic was writing a null check that could never fire.

Narrowed to `number`. `JobSnapshot` keeps `number | null` for pending jobs.

**Mutation proof — RAN.** Widened it back to `number | null`:
`test/types/typed-jobs.ts(42,3): error TS18047: 'claimed.claimExpiresAt' is possibly 'null'. EXIT=2`.
Restored, `diff` clean.

The negative direction is also pinned now: two `@ts-expect-error` lines assert that
`pending.workerId` and `pending.claimExpiresAt` still *do* require a null check on a
snapshot. Unused `@ts-expect-error` is itself an error, so these cannot go vacuous.

### 4. `claimExpiresAt` on a claimed job was asserted by nothing, anywhere — MEDIUM, fixed

Same class of hole as round 1's `run_at`/`created_at`. RAN
(`grep -rn "claimExpiresAt" test/ *.js`): the only assertion in the whole package was
`assert.equal(row.claimExpiresAt, null)` in `phase_mantle.js` — the *pending* case.
The non-null case, i.e. the lease deadline a worker actually needs, had no coverage
on either the claim path or the snapshot path.

It matters more than it looks: `claim_expires_at` sits at tuple index 7 in the Rust
destructure between `worker_id` and `attempts`, and it is one of four `i64`
timestamps. `claim_batch_returns_the_complete_job_snapshot` asserts only `> 0` for
it, which any of the other three timestamps also satisfies.

Fixed in `typed_jobs_e2e.js`: the consumer now claims with `visibilityTimeoutS: 77`,
so the lease lands at `createdAt + 77` while `runAt`/`createdAt` sit at `+0` and
`expiresAt` at `+600`. The window assertion rejects all three substitutions and
rejects `undefined`. Also added the previously-unchecked `claimed.queue`.

Not mutation-proven locally — see the note at the bottom.

### 5. `payload: TPayload` is dishonest, and I left the type alone deliberately — MEDIUM, documented + tested

Round 1 flagged this. My judgment: **do not change the declared type.**

Widening to `TPayload | null | string` would put a cast or a narrow at every single
payload access, which is exactly the boilerplate #119 asked us to remove. The honest
framing is that `TPayload` is an *unchecked assertion* — the same contract as
`JSON.parse(x) as T`, which is what every typed-queue library in this space ships.

What was actually missing is that this was nowhere near the user. Fixed: the TSDoc on
`JobSnapshot.payload` now says the generic is unchecked, names who can violate it
(another process, another binding, a raw `honker_enqueue` on your own connection),
and calls out JSON `null` specifically as the violation reachable with no raw SQL at
all. The README repeats it.

Added a test pinning that a `null` payload reads back as `null` on both the claim and
snapshot paths, so the doc cannot drift from behavior. I deliberately did *not* test
the non-JSON-text case: #152/#153 are about to make it unreachable through
`honker_enqueue`, and a test that has to poke `_honker_live` directly to stay alive is
worse than no test. `"null"` is valid JSON, so the null test survives #153.

Not implemented here per scope: the core JSON contract (#152/#153) and the
`parseJson` silent fallback, which is a fail-quietly violation but belongs with the
core contract work, not in a binding.

### 6. The two breaking changes were documented where a maintainer looks, not where a user does — MEDIUM, fixed

CHANGELOG had them. The `getJob` TSDoc had the *scoping* break but not the
camelCase/decoding one, and the README had neither. The camelCase break is the one
that silently no-ops for a plain-JS user: `row.run_at` becomes `undefined` with no
throw, and an existing `JSON.parse(row.payload)` now throws or double-decodes.

Fixed: an "Upgrading `getJob()` from 0.5.1" section in the README, the same migration
on the `getJob` TSDoc so it shows on hover, and the CHANGELOG split into the two
breaks with the concrete before/after.

### 7. The scoping break removed the only by-id lookup with no replacement — LOW, documented not fixed

Round 1's finding 3, still true: `db.queue(anything).getJob(id)` used to be a global
lookup and now returns `null`. Per scope I did not add `Database.getJob` (#134).

But "no replacement" was overstated and that was the fixable part — `honker_get_job(?)`
is still global and unchanged. The `getJob` TSDoc, README, and CHANGELOG now name
`SELECT honker_get_job(?)` on your own connection as the interim escape hatch and
point at #134. That turns a silent capability removal into a documented migration.

---

## Left deliberately

| | why |
|---|---|
| `cancel` stays global while `getJob` is scoped | #134, coordinated across bindings. The TSDoc already explains the race that makes a JS-side check wrong. |
| No `Database.getJob` | #134. Escape hatch documented instead. |
| `parseJson` falls back silently to the raw string | #152/#153 in core. Documented on the type. |
| `db.queue<A>('x')` and `db.queue<B>('x')` still cross | Not fixable without a queue-name→type registry. Scoping never claimed to solve it. |
| Node's snapshot `payload` is decoded, Python's/Ruby's is raw JSON text | Cross-binding divergence, #136's call, not this PR's to make. |
| `(payload, job) => any \| Promise<any>` on `outbox` collapses to `any` | Cosmetic, pre-existing, and the permissive behavior is what's wanted. |

---

## Verification

**RAN**
- `tsc -p test/types/tsconfig.json` — clean before and after, exit 0.
- Both type-level mutations above, each failing exactly the intended guard, each
  restored from a scratchpad copy and `diff`-verified.
- `node --check test/typed_jobs_e2e.js`, plus the sorted-key arithmetic of the
  twelve-field assertion checked against the real construction order in `api.js`.
- `grep` sweeps for the unit docs, `claimExpiresAt` coverage, `deny_unknown_fields`,
  and parallel job declarations in `extension.d.ts` / `index.d.ts` (there are none).

**Not run — and why**
The runtime assertions in finding 4 are not mutation-proven locally. Proving them
needs `napi build --release`, and the only prebuilt `.node` on this machine is from
`feat/json-payload-contract`, which still has the **6-column** `RETURNING` — READ
from that checkout's `honker_ops.rs`. It would fail my new test for the wrong reason.
Per the build-discipline constraint I did not compile; CI runs
`test/typed_jobs_e2e.js` on all four node matrix entries and `npm run test:types` on
all four as well, so pass/fail is covered there. The assertions are plain value
comparisons with no timing dependence beyond a 120 s upper slack, so the analytic
argument for each substitution (`runAt` → fails the lower bound, `expiresAt` → fails
the upper, dropped field → `undefined` fails the `typeof`) is in the test's comment.

## Verdict

**Merge, after CI.** The design answers #119: the generic flows through
`enqueue` → `claimOne`/`claimBatch`/`claim`/`claimWaker` → `getJob` and through
`Outbox` without a cast anywhere, which the type test now demonstrates on a real
`for await` worker loop. The remaining sharp edges — global `cancel`, no
`Database.getJob`, an unvalidated payload contract — are real, are tracked in #134
and #152/#153, and are now visible to a user on hover instead of only in this thread.

---

**CI on `7360d38`: 19 pass, 2 skipping (the two benches, as before).** All four node
matrix entries green, which is where `npm run test:types` and `test/typed_jobs_e2e.js`
run. Verdict stands: **merge.** Not merging — leaving that to you.

PR #125, opened 2026-08-27
feat(node): add transactional queue lifecycle events

russellromney:
## Summary

- add an opt-in Rust-core queue lifecycle feed backed by Honker streams
- append `enqueued`, `claimed`, `completed`, `retry_scheduled`, `dead_lettered`, and `cancelled` events in the same SQLite transaction as each successful state change
- expose both a typed Node async iterator and an EventEmitter-style listener over the same durable cross-process feed
- make retention topic-specific and globally coordinated, detect stale replay checkpoints, and attach stable typed reasons to terminal events
- keep the public binding in this draft TypeScript/Node only; the implementation and SQL ABI live in `honker-core`

This PR is stacked on #124 so the typed job API is reviewed first. Refs #123.

## Deliberate scope

This is a bounded observability feed for dashboards, metrics, and debugging—not a permanent audit log or exactly-once business event bus. It does not add named consumers, delivery acknowledgements, progress events, worker telemetry, lease-expiry polling, per-queue configuration, or bindings beyond Node.

Configuration is persisted once per database. Queue events are opt-in and payload capture is off by default. `includePayload` is feed-wide, which is documented explicitly.

## Retention and replay

The public option is `retentionTarget`, not `maxEvents`, because cleanup is intentionally amortized. The Rust core counts only the reserved queue-event topic, trims periodically in bounded chunks, and never lets unrelated `db.stream()` traffic consume queue-event retention.

Trim scheduling is persisted in the singleton configuration row and updated in the same queue transaction, so short-lived connections and multiple processes share one rollback-safe counter. Actual threshold scans and deletes remain periodic; they do not run on every queue state change. Lowering the target trims existing events immediately.

The core persists `trimmed_through_offset`. An explicit stale `fromOffset` throws `QueueEventOffsetExpiredError` with the requested offset, trim watermark, and oldest available offset. Omitting `fromOffset` starts at the oldest retained event. The EventEmitter adapter defaults to the latest offset for familiar live-listener behavior and throws `QueueEventsDisabledError` when the feed has not been enabled.

Terminal events separate stable machine-readable `reason` values (`explicit_failure`, `attempts_exhausted`, and `job_expired`) from human-readable error text.

## Performance

- disabled producers retain the original lean SQL paths and do not serialize event JSON, parse/copy payloads, insert stream rows, or update retention accounting
- enabled producers add one small transactional counter update per successful queue mutation; batch transitions update it once for the batch
- configuration is cached per connection for 100 ms; payload JSON is parsed only when `includePayload` is enabled
- a steady-state performance floor fills the feed to its target before measuring, specifically guarding against per-event cleanup regressions
- local 10,000-job transactional measurement: disabled ~40.8k enqueues/s, enabled fill ~17.3k/s, and enabled at target ~16.8k/s; reaching retention does not produce a cleanup cliff

## Verification

- `cargo test -p honker-core --release` — 71 passed, 1 intentional soak test ignored
- feature-enabled bundled-SQLite Rust suite — 84 passed, 6 intentional watcher/soak ignores
- `make lint` — rustfmt and Clippy pass with default and watcher feature sets
- `npm test` — 97 passed, 18 feature-gated skips
- `.venv/bin/pytest -q` — 300 passed, 19 platform/feature skips on macOS
- queue-event performance floors — 2 passed; 10,000-job benchmark run recorded above
- TypeScript declaration compile test for payloads, reasons, iterators, and listeners
- Node integration coverage for rollback, topic-specific retention amid unrelated stream traffic, replay gaps, redaction, filtering, terminal reasons, disable persistence, listener validation/cleanup, reserved-topic protection, retention across fresh connections, and typed errors and configure-driven self-healing for pre-event schemas
- cross-process Node producer/worker E2E through the EventEmitter listener and durable replay iterator
- automated SQLite loadable-extension E2E for transaction rollback, topic-specific retention, stale-checkpoint detection, payloads, unrelated stream preservation, and reserved-topic rejection
- Rust coverage for schema migration, cross-connection refresh and retention churn, rollback-safe accounting, immediate reconfiguration trimming, pre-event schemas, scheduler enqueues, batch transitions, and chunked production retention

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -469,7 +469,7 @@ jobs:
       - name: Test (basic — node only)
         working-directory: packages/honker-node
         shell: bash
-        run: node --test test/api.test.js test/basic.js test/typed_jobs_e2e.js
+        run: node --test test/api.test.js test/basic.js test/typed_jobs_e2e.js test/queue_events.test.js test/queue_events_e2e.js
       - name: Test (parity — node API surface)
         working-directory: packages/honker-node
         shell: bash
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,9 @@
 - `configureQueueEvents({ retentionTarget, includePayload })` persists one
   feed-wide policy. `retentionTarget` is approximate because topic-specific
   cleanup runs in bounded chunks rather than adding a delete to every job
-  transaction; `includePayload` applies to every queue and defaults to false.
+  transaction; its scheduling is coordinated in SQLite across short-lived
+  connections and processes. `includePayload` applies to every queue and
+  defaults to false.
 - Explicit replay checkpoints that predate retained history now throw
   `QueueEventOffsetExpiredError`. Omitting `fromOffset` starts at the oldest
   retained event, while `queueEventListener()` defaults to live events at the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,27 @@
 - Regression test claims a job at a deadline with one claim forced to
   return empty: it fails on the unpatched `api.js` and passes on the fix.
 
+## Unreleased — transactional Node queue lifecycle events
+
+- New opt-in, Rust-core queue lifecycle feed for `enqueued`, `claimed`,
+  `completed`, `retry_scheduled`, `dead_lettered`, and `cancelled` transitions.
+  Events commit atomically with the queue mutation and are observable across
+  processes through the Node async iterator or EventEmitter-style listener.
+- `configureQueueEvents({ retentionTarget, includePayload })` persists one
+  feed-wide policy. `retentionTarget` is approximate because topic-specific
+  cleanup runs in bounded chunks rather than adding a delete to every job
+  transaction; `includePayload` applies to every queue and defaults to false.
+- Explicit replay checkpoints that predate retained history now throw
+  `QueueEventOffsetExpiredError`. Omitting `fromOffset` starts at the oldest
+  retained event, while `queueEventListener()` defaults to live events at the
+  latest offset and errors clearly when the feed is disabled.
+- Terminal events include a stable `reason` (`explicit_failure`,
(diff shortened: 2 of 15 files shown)


russellromney:
This merge into `feat/node-typed-jobs` was a mistake. Queue lifecycle events and
typed job payloads are separate features closing separate issues (#123 and #119),
and folding them together meant one squash-merge would land both on `main` as a
single commit named after typed jobs, losing all eight queue-event commits.

Being re-split: #124 goes back to typed jobs only, and this branch reopens as its
own PR to `main` once #124 lands. No work is lost — the branch still points at
`f9b4ba2`.

FoxxMD:
This looks great!

One question I wasn't able to determine from looking at the changes: Along with being able to track these events in real time I think a common scenario will be getting the initial numbers on application startup. Such that I can start my app, read from honker, and know that I have _100 pending jobs in `queueA`_ before I start running any workers on the queue.

Is it possible to do this by doing a raw query to something like `SELECT honker_queue_events_read_since`? Or some other way I could get these numbers?

Minimally, I think I'd only need to do know how many jobs are queued per queue. It would be nice if I could filter, though, so I could know `20 pending jobs with > 0 attempts` , `10 jobs with priority 5 for queueA`,  `100 dead letters jobs for queueA`, etc...

If this is too out of scope, that's understandable, since you're trying to keep the design simple. I can easily track this in an auxiliary table I handle myself.

EDIT: as a visual aid referring back to my original example scenarios in #119, this is what my users see in my app in terms of queues

image

At startup: **20** queued (first attempt), **3** queued (retrying), **10** dead letter.

image

When a job is processed or enqueued they get a visual indicator showing which queue (type) changed.

The first screenshot totals are calculated from real time changes when the app is running, but I also need a baseline total count when the app first starts.

EDIT 2: Actually thinking about it some more this is probably not related to lifecycle events. I should be able to get this by simply querying `_honker_live` and `_honker_dead` tables, yeah?

PR #138, opened 2026-09-01
fix(core): propagate SQLite errors from five lookups; roll back fail()'s DELETE

### LINKED ISSUE #133: Core: .ok() swallows SQLite errors in five lookups; fail() can lose the job
`honker-core/src/honker_ops.rs` ends five lookups with `.ok()`, which discards every error, not just `QueryReturnedNoRows`.

Four are `SELECT` queries and fail closed. A real SQLite error becomes “not found,” the operation silently does nothing, and the caller never sees the database error:

- `retry_with_cache`
- `get_job`
- `lock_acquire` — returns 0, so a leader cannot acquire the lock; it does not create two leaders
- `result_get`

`fail_with_cache` is different and worse. It runs `DELETE FROM _honker_live ... RETURNING`, so the row is already deleted when Rust reads the returned values. If that read fails, `.ok()` hides the error, the function returns `Ok(0)`, and the outer statement commits. The job is left in neither `_honker_live` nor `_honker_dead`.

## Reproduction

1. Enqueue and claim a job.
2. Write non-numeric text into its `attempts` column.
3. Call `fail()`.
4. `fail()` returns false.
5. The job is missing from both live and dead tables.

This can silently lose a job when the schema or stored value has the wrong type.

## Fix

Use `rusqlite::OptionalExtension::optional()?` at all five sites. This maps only `QueryReturnedNoRows` to `None` and returns every other error to the caller.

## Tests

- At each site, test a real missing row and a broken schema or bad stored type.
- For `fail()`, check that the error reaches the caller and the live row is still present after rollback.

russellromney:
Closes #133.

Five lookups in `honker-core/src/honker_ops.rs` ended in `.ok()`, which discards every error and not just `QueryReturnedNoRows`. All five now use `rusqlite::OptionalExtension::optional()?`, so only a genuine missing row maps to `None` and every other SQLite error reaches the caller.

- `retry`
- `fail`
- `get_job`
- `lock_acquire`
- `result_get`

## Step 1 finding: the DELETE does NOT roll back

The issue's fix section assumes propagating the error is enough. It is not. I measured it before changing anything else.

A probe test enqueued a job, claimed it, wrote non-numeric text into `attempts` (SQLite is dynamically typed, so the write sticks), then called `fail()`.

With `.optional()?` alone — error propagated, no savepoint:

```
PROBE RESULT: fail() -> Err(SqliteFailure(Error { code: Unknown, extended_code: 1 },
  Some("Invalid column type Text at index: 6, name: attempts")));
  live rows = 0; dead rows = 0
```

**The row is gone.** `fail()` runs `DELETE FROM _honker_live ... RETURNING`, so the row is already deleted when the row mapper decodes it. Returning an error from the mapper does not undo that DELETE. The caller now sees the error instead of a silent `Ok(0)`, but the job is still destroyed — present in neither `_honker_live` nor `_honker_dead`.

So propagation alone is **not sufficient**, and this PR adds the savepoint the issue did not ask for.

With `SAVEPOINT honker_fail` wrapping the delete-decode-insert, same probe:

```
PROBE RESULT: fail() -> Err(...same error...); live rows = 1; dead rows = 0
```

The job survives and stays claimable. SAVEPOINT rather than BEGIN/COMMIT so it nests cleanly when the caller already holds a transaction, and so the rollback undoes only `fail()`'s own work.

## Tests

12 tests in `honker_ops::optional_error_tests`. Each site gets a genuine miss (must still be a miss) and a broken stored type (must surface the error).

```
retry_on_a_missing_job_is_still_a_miss ... ok
retry_surfaces_a_decode_error ... ok
fail_on_a_missing_job_is_still_a_miss ... ok
fail_surfaces_a_decode_error_and_rolls_the_delete_back ... ok
fail_still_moves_a_healthy_job_to_dead ... ok
fail_rolls_back_only_its_own_work_inside_an_outer_transaction ... ok
get_job_on_a_missing_job_is_still_a_miss ... ok
get_job_surfaces_a_decode_error ... ok
lock_acquire_grants_then_refuses_without_erroring ... ok
lock_acquire_surfaces_a_decode_error ... ok
result_get_on_a_missing_result_is_still_a_miss ... ok
result_get_surfaces_a_decode_error ... ok

test result: ok. 67 passed; 0 failed; 1 ignored
```

## Proof the tests are load-bearing

**Break 1 — remove the savepoint from `fail()`, keep `.optional()?`.** Exactly 2 tests fail, both on the row-survival assertion and not on the error text:

```
fail_surfaces_a_decode_error_and_rolls_the_delete_back ... FAILED
  assertion `left == right` failed: the DELETE must roll back: the job has
  to stay in _honker_live, not vanish from both tables
    left: 0
   right: 1

fail_rolls_back_only_its_own_work_inside_an_outer_transaction ... FAILED
  assertion `left == right` failed: fail()'s DELETE must be rolled back
  inside an outer transaction too
    left: 0
   right: 1

test result: FAILED. 10 passed; 2 failed
```

**Break 2 — revert all five `.optional()?` back to `.ok()`, savepoint restored.** All 6 error-propagation tests fail; the 6 genuine-miss tests still pass, which is the point:

```
retry_surfaces_a_decode_error ... FAILED
  a non-integer attempts column must reach the caller, not read as a miss: 0
fail_surfaces_a_decode_error_and_rolls_the_delete_back ... FAILED
fail_rolls_back_only_its_own_work_inside_an_outer_transaction ... FAILED
get_job_surfaces_a_decode_error ... FAILED
lock_acquire_surfaces_a_decode_error ... FAILED
result_get_surfaces_a_decode_error ... FAILED

test result: FAILED. 6 passed; 6 failed
```

Both breaks were restored and the suite is green again.

## Scope

`honker-core/src/honker_ops.rs` and `CHANGELOG.md` only. No bindings touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,41 @@
   failed `fail()` leaves the job claimable instead of destroyed.
   SAVEPOINT rather than BEGIN/COMMIT so it nests under a caller's
   transaction and rolls back only its own work.
-- 12 regression tests: a genuine miss and a broken stored type at each
-  of the five sites, plus the live-row-survives assertion for `fail()`
-  standalone and inside an outer transaction.
+- The same delete-then-more-work pattern was live at three more sites,
+  and all three were measured losing the job the same way (`live=0,
+  dead=0`) before the fix:
+  - `dead_letter_exhausted_claimable`, reached by **every ordinary
+    claim** — no `fail()` call needed. It deletes the whole matching set
+    before decoding any of it, so one bad `attempts` value lost every
+    job beside it too.
+  - `retry()`'s dead-letter branch — `DELETE` then `INSERT` as two
+    statements. A failing `_honker_dead` INSERT destroyed the job.
+  - `sweep_expired` — same `DELETE ... RETURNING` then decode then
+    INSERT shape.
+- The savepoint wrapper is now one shared helper, `in_savepoint`, used
+  at all four sites instead of copied four times. Two defects in the
+  original copy are fixed in it:
+  - The rollback result is no longer discarded. `let _ =
+    conn.execute_batch("ROLLBACK TO SAVEPOINT ...")` hid exactly the
+    kind of error this change exists to surface. A failed undo is now
+    reported, with the original cause kept intact in the message.
+  - A failed `RELEASE` no longer strands the caller inside the
+    transaction the savepoint opened. `RELEASE` of the outermost
+    savepoint is the COMMIT, so it can fail — measured on a
+    rollback-journal database with a concurrent reader: the caller got
+    `database is locked` **and** a connection with
+    `is_autocommit() == false` whose own reads claimed the job had been
+    dead-lettered. Bindings hold long-lived connections, so every later
+    call joined that transaction. The helper now ends it.
+- `ack_batch` and `claim_batch`'s claiming UPDATE deliberately stay
+  savepoint-free; both now carry a comment saying why.
+- 27 regression tests. At each of the five `.ok()` sites: a genuine miss
+  and a broken stored type. At each of the four savepoint sites: the
+  live row survives a decode failure and a failing `_honker_dead`
+  INSERT, and a genuine miss still returns 0/None. Plus the
+  `SELECT honker_*(...)` scalar-function path — the only one bindings
+  and ORM users take, and previously untested — for `fail`,
+  `claim_batch`, and `sweep_expired`.
 
 ## 2026-08-27 — Node 0.5.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 # CHANGELOG
 
+## Unreleased — Core SQLite error propagation
+
+- `honker-core` no longer discards SQLite errors in five lookups.
+  `retry`, `fail`, `get_job`, `lock_acquire`, and `result_get` ended in
+  `.ok()`, which mapped every error to "no row", not just
+  `QueryReturnedNoRows`. A broken stored value turned into a silent
+  no-op: `retry`/`fail` reported "not our claim", `get_job` reported a
+  missing job, `lock_acquire` reported the lock as held so no leader
+  could ever start. All five now use `OptionalExtension::optional()?`.
+- `fail()` could lose a job outright. It runs
+  `DELETE FROM _honker_live ... RETURNING`, so the row is already gone
+  when the row mapper decodes it. Measured: propagating the mapper's
+  error does **not** undo that DELETE — the job ended up in neither
+  `_honker_live` nor `_honker_dead`. The delete-decode-insert now runs
+  inside a `SAVEPOINT honker_fail` that rolls back on any error, so a
+  failed `fail()` leaves the job claimable instead of destroyed.
+  SAVEPOINT rather than BEGIN/COMMIT so it nests under a caller's
+  transaction and rolls back only its own work.
+- 12 regression tests: a genuine miss and a broken stored type at each
+  of the five sites, plus the live-row-survives assertion for `fail()`
+  standalone and inside an outer transaction.
+
 ## 2026-08-27 — Node 0.5.1
 
 - Node `@russellthehippo/honker-node`: 0.5.1, with the four
--- a/honker-core/src/honker_ops.rs
+++ b/honker-core/src/honker_ops.rs
@@ -19,6 +19,7 @@
 //! tested once and inherited by every consumer.
 
 use rusqlite::Connection;
+use rusqlite::OptionalExtension;
 use rusqlite::functions::{Context, FunctionFlags};
 use rusqlite::types::ValueRef;
 use serde_json::{Value, json};
@@ -863,7 +864,7 @@ pub fn retry(
                 ))
             },
         )
-        .ok();
+        .optional()?;
     let Some((id, queue, payload, priority, run_at, max_attempts, attempts, created_at)) = row
     else {
         return Ok(0);
@@ -908,7 +909,33 @@ pub fn retry(
 
 /// Unconditionally move the claim to `_honker_dead` with the given
 /// error. Returns 1 if moved, 0 if not our claim.
+///
+/// The DELETE uses RETURNING, so the row is already gone by the time
+/// the row mapper decodes it. A decode failure there — a non-integer
+/// `attempts`, say — would otherwise leave the job in neither
+/// `_honker_live` nor `_honker_dead`: silent job loss. Verified: an
+/// error out of the mapper alone does NOT undo the DELETE. So the
+/// delete-decode-insert runs inside a SAVEPOINT and any error rolls
+/// it back before propagating. SAVEPOINT rather than BEGIN/COMMIT so
+/// we nest cleanly when the caller already holds a transaction.
 pub fn fail(conn: &Connection, job_id: i64, worker_id: &str, error: &str) -> rusqlite::Result<i64> {
+    conn.execute_batch("SAVEPOINT honker_fail")?;
+    let result = fail_inner(conn, job_id, worker_id, error);
+    if result.is_err() {
+        let _ =
+            conn.execute_batch("ROLLBACK TO SAVEPOINT honker_fail; RELEASE SAVEPOINT honker_fail");
+        return result;
+    }
+    conn.execute_batch("RELEASE SAVEPOINT honker_fail")?;
+    result
+}
+
+fn fail_inner(
+    conn: &Connection,
+    job_id: i64,
+    worker_id: &str,
+    error: &str,
+) -> rusqlite::Result<i64> {
     #[allow(clippy::type_complexity)]
     let row: Option<(i64, String, String, i64, i64, i64, i64, i64)> = conn
         .query_row(
@@ -931,7 +958,7 @@ pub fn fail(conn: &Connection, job_id: i64, worker_id: &str, error: &str) -> rus
                 ))
             },
         )
-        .ok();
+        .optional()?;
     let Some((id, queue, payload, priority, run_at, max_attempts, attempts, created_at)) = row
     else {
         return Ok(0);
@@ -1013,7 +1040,7 @@ pub fn get_job(conn: &Connection, job_id: i64) -> rusqlite::Result<String> {
                 ))
             },
         )
-        .ok();
+        .optional()?;
     let Some((
         id,
         queue,
@@ -1143,7 +1170,7 @@ pub fn lock_acquire(
             rusqlite::params![name],
             |r| r.get(0),
         )
-        .ok();
+        .optional()?;
     Ok(if current.as_deref() == Some(owner) {
         1
     } else {
@@ -1619,7 +1646,7 @@ pub fn result_get(conn: &Connection, job_id: i64) -> rusqlite::Result<Option<Str
             rusqlite::params![job_id],
             |r| Ok((r.get(0)?, r.get(1)?)),
         )
-        .ok();
+        .optional()?;
     match row {
         None => Ok(None),
         Some((_, Some(exp))) if exp <= now_unix(conn)? => Ok(None),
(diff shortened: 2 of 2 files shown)


russellromney:
Adversarial review. Throwaway worktree at 4881f27, probe module appended to `honker_ops.rs` and reverted after. Everything below marked "measured" was run, not read.

**Confirmed (measured):** the savepoint does what you claim. Decode error -> `live=1, dead=0`. A forced INSERT failure (`BEFORE INSERT ON _honker_dead ... RAISE(ABORT)`, standing in for a constraint/IO failure) -> also `live=1, dead=0`. Error path leaves `is_autocommit()==true` and no stale savepoint (a following `ROLLBACK TO SAVEPOINT honker_fail` returns "no such savepoint"). `SELECT honker_fail(...)` — the path every binding and ORM user takes, and the one no new test uses — works too: SAVEPOINT/RELEASE from inside the UDF is fine on the healthy path (`Ok(1)`, live 0/dead 1) and on the rollback path inside an explicit `BEGIN` (error, `COMMIT` clean, live 1/dead 0). Genuine-miss behavior at all five sites is unchanged.

### 1. (high) Three other sites still lose the job the same way; only `fail()` got the savepoint

All measured on this branch, same corruption the issue describes (`UPDATE _honker_live SET attempts='not-a-number'`; in SQLite text > integer, so the row matches `attempts >= max_attempts`):

- **`claim_batch` -> `dead_letter_exhausted_claimable`**: an ordinary claim on that queue returns `Err("Invalid column type Text at index: 6, name: attempts")` and leaves **live=0, dead=0**. No `fail()` call involved — this is the hot path, so it is strictly more reachable than the bug this PR fixes. It also deletes the whole matching batch before decoding, so it can lose several jobs at once.
- **`sweep_expired`**: same `DELETE ... RETURNING` -> decode shape, same result: **live=0, dead=0**.
- **`retry`'s dead-letter branch**: `DELETE` then `INSERT` as two statements, no savepoint. With the INSERT forced to fail: **live=0, dead=0**. `fail()` under the identical trigger keeps the row. So after this PR, `retry()` can destroy a job in a way `fail()` no longer can.

The PR body reads as if silent job loss is closed. It is closed at one of four sites. Fine as a follow-up issue, but it should be stated, not implied away.

### 2. (medium) The savepoint adds a new failure mode: a failed `RELEASE` leaves the connection inside a transaction it opened

`conn.execute_batch("RELEASE SAVEPOINT honker_fail")?` — if the RELEASE (which is the COMMIT for the outermost savepoint) fails, `fail()` returns that error with the transaction still open.

Measured, file DB, `journal_mode=delete`, `busy_timeout=0`, a second connection holding `BEGIN; SELECT count(*) FROM _honker_live`:

| | result | `is_autocommit()` after | writer-visible rows |
|---|---|---|---|
| this branch | `Err("database is locked")` | **false** | live=0, dead=1 |
| main (no savepoint) | `Err("database is locked")` | true | live=1, dead=0 |

So the caller gets an error *and* a connection that is silently mid-transaction, whose own reads say the job was dead-lettered. Bindings hold long-lived connections: every later honker call on it joins that transaction, and the `honker_fail` savepoint stays on the stack (a second `fail()` then pushes a same-named frame over it). Closing the connection rolls it all back (`live=1, dead=0` from a fresh reader), so the work is not lost — but the state is invisible until then. WAL narrows the window; honker explicitly documents supporting DELETE-journal, and ORM users often run it.

Related, same function: `let _ = conn.execute_batch("ROLLBACK TO ...; RELEASE ...")` discards the result. If that fails you have a connection in an unknown state and no signal. Repo rule is fail loudly.

### 3. (low) No test drives `SELECT honker_fail(...)`

Both nesting tests call the Rust `fail()` directly. The savepoint runs `execute_batch` on the connection while the caller's `SELECT` is mid-step in the extension path; I checked it by hand and it is fine, but nothing pins it, and that is the only path bindings use.

**Would I merge:** yes, after (2) — it is a regression this PR introduces — with (1) filed as an issue. (3) is cheap to add while you are in there.

russellromney:
Follow-up on the review. All three remaining sites confirmed and fixed in 3ab43aa. CI green.

## Measurements BEFORE the fix

Probe module appended to `honker_ops.rs` at 4881f27 and reverted after, same method as the review. Two forcing modes: the issue's corruption (`UPDATE _honker_live SET attempts='not-a-number'` — text > integer in SQLite, so the row also matches `attempts >= max_attempts`), and a `BEFORE INSERT ON _honker_dead ... RAISE(ABORT)` trigger standing in for a constraint or I/O failure on the dead-table INSERT.

| site | forcing | result | live | dead |
|---|---|---|---|---|
| 1. `claim_batch` → `dead_letter_exhausted_claimable` | decode | `Invalid column type Text at index: 6, name: attempts` | **0** | **0** |
| 1. same, two jobs, one corrupt | decode | same error | **0, 0** | **0, 0** |
| 1. same | dead INSERT aborts | `dead insert blocked` | **0** | **0** |
| 2. `retry()` dead-letter branch | dead INSERT aborts | `dead insert blocked` | **0** | **0** |
| 3. `sweep_expired` | decode | `Invalid column type Text at index: 6, name: attempts` | **0** | **0** |
| 3. same | dead INSERT aborts | `dead insert blocked` | **0** | **0** |

No site turned out to be already safe. #3 does have the same shape — `DELETE ... RETURNING`, then decode, then INSERT — it just reads differently from `fail()` because the select/insert are two `prepare_cached` handles rather than one `query_row`.

The batch row is the one that bothers me most: site 1 deletes the whole matching set before decoding any of it, so one bad `attempts` value takes every job beside it, on the ordinary claim path.

Defect 2 from the review reproduced too, file DB, `journal_mode=delete`, `busy_timeout=0`, second connection holding `BEGIN; SELECT count(*)`:

```
result=Err("database is locked")  is_autocommit()=false  writer-visible live=0 dead=1
```

## What changed

One shared helper, `in_savepoint(conn, name, body)`, replacing the copy in `fail()` and now used at all four sites. Both defects fixed inside it:

- **(4) the swallowed rollback** — the undo result is no longer discarded. On a failed undo the caller gets the original cause with the undo failure appended, so the cause's text (which callers and tests match on) survives.
- **(5) the stranded transaction** — `RELEASE` of the outermost savepoint is the COMMIT and can fail. The helper records whether *it* opened the transaction (`is_autocommit()` before `SAVEPOINT`); if the undo's `RELEASE` also cannot run and we own the transaction, it issues a plain `ROLLBACK`. `is_autocommit()` is back to true on every error path.

Ordering matters and is deliberate: the undo is always `ROLLBACK TO SAVEPOINT x; RELEASE SAVEPOINT x` first, never a bare `COMMIT`/`ROLLBACK`, because every one of these runs as `SELECT honker_*(...)` with the outer statement still stepping. The plain `ROLLBACK` is the last step only, and only when we own the transaction.

There is also a guard for the case where SQLite already rolled the whole transaction back itself (ABORT / SQLITE_FULL / I/O error): `is_autocommit()` is checked first, and the cause is returned untouched rather than dressed up with a bogus "no such savepoint".

Sites that deliberately stay savepoint-free now say why in a comment:
- `ack_batch` — the DELETE is the whole operation; `RETURNING id` is counted, never decoded, and nothing runs after it.
- `claim_batch`'s claiming UPDATE — an UPDATE, not a DELETE. A decode failure leaves the row in `_honker_live` with the claim held, so it is reclaimable when the visibility timeout expires. Nothing to undo.

## Review point 3

Added. `SELECT honker_fail(...)`, `SELECT honker_claim_batch(...)` and `SELECT honker_sweep_expired(...)` now have rollback tests on the scalar-function path, plus a healthy-path one for `honker_fail`. Four of the five savepoints are entered from inside a UDF while the caller's statement is stepping, so that assumption is load-bearing now and needed pinning.

## Proof the new guards are load-bearing

Each guard broken, the failing assertion recorded, then restored. Not just "the suite passes".

**A — drop the savepoint from `dead_letter_exhausted_claimable`** (4 failed):
```
claim_batch_rolls_back_a_dead_letter_decode_failure ... FAILED
  assertion `left == right` failed: the dead-letter DELETE must roll back:
  the job has to stay in _honker_live, not vanish from both tables
  left: 0
claim_batch_dead_letter_keeps_the_whole_batch ... FAILED
  assertion `left == right` failed: one undecodable row must not take the
  rest of the batch with it
  left: (0, 0)
claim_batch_rolls_back_when_the_dead_insert_fails ... FAILED
  assertion `left == right` failed: the DELETE must roll back when the
  dead-table INSERT fails
  left: 0
sql_honker_claim_batch_rolls_back_the_dead_letter_sweep ... FAILED
```

**B — drop the savepoint from `retry()`'s dead-letter branch**:
```
retry_rolls_back_the_dead_letter_when_the_insert_fails ... FAILED
  assertion `left == right` failed: retry() must not destroy the job in a
  way fail() no longer can
  left: 0
```

**C — drop the savepoint from `sweep_expired`** (3 failed):
```
sweep_expired_rolls_back_a_decode_failure ... FAILED
  assertion `left == right` failed: the sweep's DELETE must roll back, not
  lose the job
  left: 0
sweep_expired_rolls_back_when_the_dead_insert_fails ... FAILED
sql_honker_sweep_expired_rolls_back ... FAILED
```

**D — restore the swallowed undo result (`let _ = ...`)**:
```
in_savepoint_reports_a_rollback_failure_without_losing_the_cause ... FAILED
  a failed rollback must be reported, not discarded, got: original cause
```

**E — skip the plain ROLLBACK after a failed RELEASE**:
```
fail_leaves_no_open_transaction_when_release_fails ... FAILED
  a failed RELEASE must not leave the caller inside the transaction fail()
  opened; the connection is long-lived and every later call would silently
  join it
```

## Genuine-miss behavior

Unchanged, and pinned: `savepoint_sites_still_report_a_genuine_miss` asserts `sweep_expired` → 0, `claim_batch` → `"[]"`, `retry` → 0, `fail` → 0 on rows that were never enqueued, and that `is_autocommit()` is still true afterwards (no savepoint left on the stack). The existing per-site miss tests are untouched.

`optional_error_tests` is 27 tests, up from 12. Full `honker-core` lib suite: 82 passed, 0 failed.

## Noted, not touched

`scheduler_update()` has its own pre-existing `SAVEPOINT honker_sched_update` with the same `let _ = conn.execute_batch("ROLLBACK TO SAVEPOINT ...")` swallow. It predates this PR and is a different operation (field UPDATEs, no delete-then-insert, so no job-loss path), so I left it alone rather than widen this change. It should get `in_savepoint` too — worth a small follow-up issue.

russellromney:
Second adversarial review, different reviewer from round one. Round one asked "is it correct". This round asked: is it complete, are the tests load-bearing, is the savepoint right under nesting, does it fail loudly enough, and is `in_savepoint` the right abstraction. Everything marked RUN was executed against `3ab43aa` in a throwaway worktree; INFERRED means I read it and reasoned, without running it. One defect found and fixed in `753f6d0`.

## What I attacked and could not break

**Completeness (RUN).** Swept the whole crate rather than trusting the list. Every `DELETE FROM` in `honker_ops.rs` (12 of them), every `RETURNING` (7), every `.ok()`, every `let _ =` on a fallible connection call.

- The four wrapped sites are the four that need wrapping. `ack_batch` (`RETURNING id`, counted not decoded, nothing after) and `claim_batch`'s claiming UPDATE (a decode failure leaves the row claimed and reclaimable, nothing lost) are correctly savepoint-free, and now say why.
- No `.ok()` on a rusqlite result remains anywhere in `honker_ops.rs`. The two in `lib.rs` are a watcher poll and a `HONKER_SOAK_DURATION_SECS` env read — neither is a data path.
- **`scheduler_update` is the only remaining `let _ =` swallow on a fallible connection call in the crate.** Confirmed by grep, and #156 is open and tracks it. Not touched here. Worth noting for #156: `scheduler_update` has *both* round-one defects, not just the swallow — its `RELEASE SAVEPOINT honker_sched_update` on the success path can fail and strand the connection exactly the way `fail()` did. Moving it onto `in_savepoint` fixes both at once.
- No dead-letter requeue path exists, so there is no job-loss path in the other direction.

**Every guard is load-bearing (RUN).** I broke each one, ran the suite, recorded the failure, and restored. Independently, not by re-reading round one's table.

| mutation | tests that failed |
|---|---|
| `fail()` calls `fail_inner` directly (no savepoint) | `fail_surfaces_a_decode_error_and_rolls_the_delete_back`, `fail_rolls_back_only_its_own_work_inside_an_outer_transaction`, `sql_honker_fail_rolls_back_through_the_scalar_function` |
| `dead_letter_exhausted_claimable` calls its `_inner` directly | `claim_batch_rolls_back_a_dead_letter_decode_failure`, `claim_batch_dead_letter_keeps_the_whole_batch`, `claim_batch_rolls_back_when_the_dead_insert_fails`, `sql_honker_claim_batch_rolls_back_the_dead_letter_sweep` |
| `sweep_expired` calls its `_inner` directly | `sweep_expired_rolls_back_a_decode_failure`, `sweep_expired_rolls_back_when_the_dead_insert_fails`, `sql_honker_sweep_expired_rolls_back` |
| `retry()`'s `in_savepoint(...)` replaced with a plain closure | `retry_rolls_back_the_dead_letter_when_the_insert_fails` |
| all five `.optional()?` back to `.ok()` | 7 tests, one per site plus the two rollback assertions |
| `let _ =` restored on the undo | `in_savepoint_reports_a_rollback_failure_without_losing_the_cause` |
| plain `ROLLBACK` after a failed RELEASE skipped | `fail_leaves_no_open_transaction_when_release_fails` |
| `owns_transaction` hardcoded `true` | `fail_rolls_back_only_its_own_work_inside_an_outer_transaction`, `in_savepoint_reports_a_rollback_failure_without_losing_the_cause` |

That last one matters and nothing in round one covered it explicitly: getting `owns_transaction` wrong means issuing a bare `ROLLBACK` inside a caller's transaction and destroying the caller's uncommitted work. The existing nesting test catches it.

Nothing here is a test that only checks the error and not the row. Every rollback test asserts `live_count == 1` and `dead_count == 0`.

**Nesting, both cases (RUN).** Probed directly rather than inferring from the tests.

- No outer transaction: error names the column, `is_autocommit()` back to `true`, and a following `ROLLBACK TO SAVEPOINT honker_fail` returns `no such savepoint` — the frame is popped, not leaked.
- Caller already inside `BEGIN`: error propagates, the caller stays inside *its own* transaction (correct — closing it would be the bug), the frame is popped, and the caller's prior INSERT survives.

The second case is the one people get wrong and it is right here. `SAVEPOINT` / `ROLLBACK TO SAVEPOINT` / `RELEASE` never issues a bare `COMMIT`/`ROLLBACK` while the outer statement is stepping, except the one deliberate last-resort `ROLLBACK` that only fires when we own the transaction. That ordering is correct.

## Findings

### 1. (high, fixed in `753f6d0`) A panicking body skipped the undo entirely, and the next call returned a silent miss

`in_savepoint` hung its undo off the two `Err` arms. A panic unwinds past both. Measured on `3ab43aa`:

```
PROBE[panic]     caught=true  autocommit_after=false  live_after=0
PROBE[panic]     later fail() => Ok(0), autocommit=false
PROBE[udf-panic] result=Err("unwinding panic")  autocommit_after=false
PROBE[udf-panic] stale savepoint still usable? Ok("yes")
```

Read the second line. The body's DELETE stayed applied, the connection was left mid-transaction, and **the next `fail()` on that connection returned `Ok(0)`** — a miss, not an error, from inside the leaked transaction. That is the exact silent-job-loss shape this PR exists to close, entering through a different door.

The UDF line is why it is not theoretical. rusqlite catches a panic raised inside a scalar function and hands the caller an `unwinding panic` error, so the connection *survives to be reused* — and through `SELECT honker_*(...)` the frame was still on the stack afterwards, on the long-lived connection every binding holds. Every later honker call on it silently joins that transaction.

Reachability, honestly: I found no panic path in the four current bodies (no `unwrap`, no indexing, no arithmetic; rusqlite returns `Err` rather than panicking on decode). So this is latent, not live. It still gets fixed, for two reasons. `in_savepoint` is now a shared helper whose whole job is "the connection is never left stranded", and it did not deliver that; and #156 proposes putting a fifth body under it, which would inherit the hole.

Fix: the undo also hangs off a drop guard, disarmed on every ordinary path. `undo_savepoint` is split so the guard and the error path share one implementation — `undo_savepoint_frame` returns the list of things that went wrong, `undo_savepoint` dresses the caller's error with it, and the guard, which has no error to return while unwinding, reports a failed undo on stderr. That is the only channel left inside `Drop`; the panic itself is the loud part.

Two new tests, both broken and watched to fail before landing:

```
in_savepoint_undoes_its_work_when_the_body_panics ... FAILED
  a panicking body must not leave the connection inside the transaction
  in_savepoint opened
a_panicking_savepoint_body_leaves_no_frame_on_the_scalar_path ... FAILED
```

### 2. (medium, fixed in `753f6d0`) The `is_autocommit()` short-circuit had no test

`undo_savepoint` opens with a guard for the case where SQLite already rolled the whole transaction back itself, so the cause is returned untouched rather than dressed up with a bogus "no such savepoint". Good defensive reasoning — and no test reached it. Replacing `if conn.is_autocommit()` with `if false` left all 27 tests green (RUN).

Added `a_self_rolled_back_transaction_returns_the_cause_unchanged`: a `RAISE(ROLLBACK)` trigger on `_honker_dead` makes SQLite roll the transaction back on its own. It asserts the cause reaches the caller, that the message does *not* claim an unknown state, and that the job survived. It FAILS with the short-circuit removed (RUN).

Note the shape of the gap: the branch was written for a real failure mode (`ABORT` / `SQLITE_FULL` / I/O error) but the existing `block_dead_inserts` helper uses `RAISE(ABORT)`, which undoes only the statement and leaves the transaction open — so it never enters this branch. Easy to miss.

### 3. (medium, NOT fixed — needs a product call) The error does not tell the user their job was preserved

This is the round-two question the PR does not answer. Measured error text at all four sites, through `SELECT honker_*(...)`:

```
PROBE[fail]   => Invalid column type Text at index: 6, name: attempts
PROBE[retry]  => Invalid column type Text at index: 6, name: attempts
PROBE[sweep]  => Invalid column type Text at index: 6, name: attempts
PROBE[claim]  => Invalid column type Text at index: 6, name: attempts
```

An operator whose `claim_batch` starts throwing `Invalid column type Text at index: 6` learns nothing about which job, which operation, or — the thing they most need at 3am — that the dead-letter move was rolled back and the job is still sitting in `_honker_live`. Four different operations produce a byte-identical message. The savepoint does the right thing and then says nothing about having done it.

I did not fix this, deliberately. The PR made "keep the cause's text intact — callers and tests match on it" an explicit contract, and wrapping the message breaks it. Choosing between that contract and a more useful message is a product decision, not a review one. Proposed as a follow-up issue: add the operation name and job id, and state that the row was preserved, without disturbing the cause substring callers match on.

### 4. (low, not fixed) `claim_batch`'s comment about sharing the caller's transaction is wrong in the common case

> Same outer SQL statement / connection, so this shares the caller's transaction with the claim UPDATE below.

In autocommit — which is what `SELECT honker_claim_batch(...)` runs in, since a read statement does not clear autocommit — `in_savepoint`'s `SAVEPOINT` *opens* a transaction and its `RELEASE` commits it. So the dead-letter move commits before the claiming UPDATE begins, in two commits rather than one. Not a regression (before this PR each statement autocommitted separately anyway) and not a correctness problem (the jobs land in `_honker_dead`, which is where they belong). The comment just no longer describes what happens. INFERRED from the code plus the measured autocommit state; I did not instrument the commit boundary.

### 5. (low, out of scope) `scheduler_tick` can fire a schedule twice

`scheduler_tick` enqueues, then persists `next_fire_at` as a separate statement. In autocommit the enqueue is already committed when the UPDATE runs, so a failure between them re-fires the same boundary on the next tick. Duplicate, not loss, and honker is at-least-once — so this is a note, not a defect. Pre-existing, untouched by this PR. Mentioning it so the next person sweeping this file does not think it was missed.

Also noted, not touched: the binding crates (`packages/honker-rs`, `packages/honker-node`, `packages/honker`) carry several `let _ = ... "ROLLBACK"` swallows on their own transaction wrappers. Pre-existing, outside `honker-core`, different in kind — they sit on an already-failing path where the error is being returned anyway. A separate sweep if you want one.

## Design: is `in_savepoint` the right abstraction?

Yes, with the drop guard. Without it, no — an abstraction that promises "the connection is never left stranded" and delivers that on two of three exit paths is worse than the inline copy it replaced, because the remaining hole is invisible at every call site. The guard closes it, and splitting `undo_savepoint_frame` out means the panic path and the error path cannot drift apart.

The two design questions worth stating plainly:

- **Body panics**: handled now (finding 1).
- **RELEASE fails**: handled before this review, and I confirmed the fix is load-bearing (mutation table). The `owns_transaction` bookkeeping is the subtle part and it is right: a plain `ROLLBACK` only ever fires when this savepoint opened the transaction, only as a last resort, and never while an outer statement could still be stepping in a state that would refuse it.

One thing I checked and am satisfied with: `Drop` cannot return an error, so a failed undo during unwind goes to stderr with a `honker: error:` prefix rather than being swallowed. That is the loudest channel available mid-unwind, and the panic is propagating regardless.

## Would I merge

Yes, with `753f6d0`. The four sites are the right four, all five `.optional()?` conversions and all four savepoints are load-bearing under mutation, both nesting cases are proven rather than assumed, and the one hole I found is closed with tests that fail when the guard is removed.

Finding 3 is the one I would not let sit. The code now preserves the job correctly and tells the user nothing about it, which is a worse operator experience than it deserves. That is a follow-up issue, not a merge blocker.

`honker-core` lib: 85 passed, 0 failed. `cargo fmt --check` and `cargo clippy --all-targets` clean.

Issue #136, opened 2026-09-01
Add full job details and typed payloads to every binding


russellromney:
## Problem

Node #124 adds a full public job shape. The shared core `honker_claim_batch()` call now returns all job details too.

This JSON change is safe for old bindings because they ignore fields they do not use. However, most bindings still throw away some of those fields when they build a claimed `Job`.

Each binding can be updated on its own. We do not need one large breaking release.

## Fields every job should have

Claimed jobs and read-only job snapshots should include:

- `id`
- `queue`
- `payload`
- `state`
- `priority`
- `run_at`
- `worker_id`
- `claim_expires_at`
- `attempts`
- `max_attempts`
- `created_at`
- `expires_at`

The core JSON stays snake_case. Each binding should use the normal naming style for its language.

A claimed `Job` also has methods such as ack, retry, fail, and heartbeat. A `JobSnapshot` or `JobRow` is only data.

`claimed_at` is not included yet. That work is tracked in #135.

## Typed payloads

Payload types only help the binding and the user's code. Honker does not check the payload shape in the database. All apps that write to the same queue must agree on the JSON shape.

Use a design that fits each language:

- TypeScript and Bun: generic `Queue<T>`, `Job<T>`, and snapshots.
- Python: optional generic type hints. Runtime values stay normal decoded JSON.
- Rust: keep generic serialization and `payload_as<T>()`-style decoding.
- Go: keep raw JSON and add a generic decode helper or `TypedQueue[T]`.
- .NET and JVM/Kotlin: use a serializer or codec with typed queue wrappers.
- Ruby, Elixir, and C++: keep dynamic values or raw JSON unless a small codec helper is useful.

## Queue ownership

Queue-level `getJob()` should only return jobs from that queue. Database-level lookup can stay global. #134 tracks that change.

## Binding checklist

- [ ] Node: merge #124.
- [ ] Bun: add typed queues, jobs, and snapshots.
- [ ] Python: keep all claim fields, add `expires_at`, and fix the old comment that says claim results are narrow.
- [ ] Rust: expose all fields already present in `RawJob`.
- [ ] Go: expose all fields on claimed `Job`.
- [ ] .NET: expose all claimed-job fields and consider a typed queue wrapper.
- [ ] JVM/Kotlin: add all fields to the existing typed job API and add typed `getJob()` snapshots.
- [ ] Ruby: add readers for all claim fields.
- [ ] Elixir: add all fields to `%Honker.Job{}`.
[...]

Issue #135, opened 2026-09-01
No claimed_at, so job running time is unanswerable


russellromney:
`_honker_live` has no `claimed_at` column, so Honker cannot report how long the current job attempt has been running.

- `created_at` is the total job age and includes time spent waiting in the queue.
- `run_at` is when the job becomes ready to run.
- `claim_expires_at` changes on every heartbeat.

This came from #119, which asked how long a job has been running or queued. #124 adds the four fields named in that issue. A stable claim time needs a schema change.

## Change

Add a nullable `claimed_at INTEGER` column. Return it on claimed jobs and from `getJob()`.

## Rules

- It is `NULL` until the first claim.
- Set it to `unixepoch()` on every successful claim or reclaim. It measures the current attempt, not the first attempt.
- Clear it when `retry` moves the job back to pending.
- Heartbeats must not change it.

## Tests and rollout

- Add a migration test for a database created before the column existed.
- Test first claim, heartbeat, retry, and reclaim.
- Decide whether all bindings expose it in the same release or add it one binding at a time.
- Add it to the full job shape tracked in #136 after the schema work lands.

russellromney:
Note for whoever builds the per-binding exposure.

#140's migration adds `claimed_at` to existing databases and deliberately leaves
old rows NULL rather than backfilling a fake timestamp. That is the right call,
but it means `claimed_at` is NULL for two different reasons:

- the job is not currently claimed
- the row predates the column

A binding that computes running time as `now - claimed_at` cannot tell them
apart, so the first consumer will silently report pre-migration jobs as
never-claimed. Decide how bindings distinguish the two before exposing the field
— a claimed job with a NULL `claimed_at` is the case to reason about, since
`state = 'processing'` and no claim timestamp is only possible on a legacy row.

Found while documenting the field for honker-site.

Issue #134, opened 2026-09-01
Make getJob() and cancel() scope consistent


russellromney:
## Problem

Job IDs are unique across the whole database. The core functions `honker_get_job(job_id)` and `honker_cancel(job_id)` therefore work across all queues.

But bindings put these methods on a queue object:

```text
emails.cancel(smsId)  // deletes the SMS job
emails.getJob(smsId)  // used to return the SMS job
```

Node #124 now makes `Queue.getJob()` return null for a job from another queue. It needs this check because `Queue<EmailPayload>.getJob()` must not return an SMS payload.

Node #124 does not change `cancel()`. Checking the queue first and cancelling in a second query would have a race. Another worker could claim or finish the job between the two queries.

## API we want

- `Database.getJob(id)` looks up a job in any queue.
- `Queue.getJob(id)` only looks up jobs in that queue.
- `Database.cancel(id)` cancels a job in any queue.
- `Queue.cancel(id)` only cancels jobs in that queue.

This keeps database-wide tools available for admin and monitoring code. It also makes queue methods safe for typed payloads.

## Core changes

- Add one atomic core call for queue-scoped cancellation. It could be `honker_cancel(queue, job_id)` or a new function name.
- Keep the current one-argument global cancel function during the move.
- Decide whether queue-scoped `getJob()` should keep checking the returned row in each binding or use a shared core function.

## Binding changes

- Add global `getJob()` and `cancel()` methods to each database object.
- Make queue-level `getJob()` queue-scoped in every binding. Node #124 already does this.
- Make queue-level `cancel()` use the new atomic core call.
- If needed, mark the old queue behavior as deprecated before removing it.
- Use normal names for each language.

## Tests

- A queue cannot read a job from another queue.
- A queue cannot cancel a pending or processing job from another queue.
- Database-level lookup and cancel still work across all queues.
- Repeating cancel on a missing job returns false.
- Queue cancel uses one core statement. Do not read first and cancel second.
- Any old behavior kept for compatibility is documented and tested.

Related: #124 and #136.

russellromney:
Two ways to resolve, pick one:

- **Jobs are globally addressed.** Add `Database.cancel(id)`, deprecate cancel on the
  queue handle.
- **Jobs are queue-owned.** Add `honker_cancel(queue, job_id)` in core, propagate
  across bindings.

Core can carry both SQL arities during the transition if the old one needs to keep
working.

russellromney:
[...]

PR #139, opened 2026-09-01
feat(go): expose every claimed-job field and typed payload decoding

russellromney:
Part of #136 — the Go binding's slice of "full job details and typed payloads in every binding".

**Stacked on #124 (`feat/node-typed-jobs`). Merge #124 first.** The widened
`honker_claim_batch` RETURNING clause that exposes the full job snapshot lives
only on that branch; on `main` claim_batch still returns the old 8 fields, so
this PR targets `feat/node-typed-jobs` as its base.

## What changed

- `Job` now carries all twelve fields the core already returns: `ID`, `Queue`,
  `Payload`, `State`, `Priority`, `RunAt`, `WorkerID`, `ClaimExpiresAt`,
  `Attempts`, `MaxAttempts`, `CreatedAt`, `ExpiresAt`. Previously it kept five.
  A claimed `Job` and a `GetJob` snapshot (`JobRow`) now describe the same row;
  `Job` keeps `Ack` / `Retry` / `Fail` / `Heartbeat`, `JobRow` stays data only.
- New `honker.DecodePayload[T](payload []byte) (T, error)` and
  `JobRow.PayloadBytes()`, so a claimed job and a snapshot decode a typed
  payload through the same path. Payloads stay raw JSON, as the issue calls for
  in Go.
- README section documenting the field list and the decode helper.

No core or SQL ABI changes. `packages/honker-node` untouched. `claimed_at`
(#135) is deliberately not here.

## Not runtime validation

The type parameter on `DecodePayload[T]` is a compile-time contract between
producer and consumer. Honker does not inspect or validate payload shape in the
database — a mismatch surfaces as an unmarshal error at decode time, never as
an enqueue failure. The doc comment and README both say so.

## Tests

`packages/honker-go/typed_jobs_test.go`:

- `TestClaimedJobAndSnapshotCarryEveryField` — enqueues with a priority (7),
  TTL (600s), and max-attempts (5), then asserts every one of the twelve fields
  against its expected *value* on three reads: the pending snapshot, the
  claimed `Job`, and a second independent `Database` handle's processing
  snapshot. Checks `ExpiresAt - RunAt == 600` exactly, `ClaimExpiresAt` inside
  the claim-time + 120s visibility window, `Attempts` going 0 → 1, `WorkerID`
  and `ClaimExpiresAt` nil while pending and set once claimed, and `RunAt` /
  `CreatedAt` unchanged by the claim. After ack, the reader gets `nil`.
- `TestDelayedJobReportsItsRunAt` — a `Delay: 3600` job reports the right
  `RunAt` and is not claimable early.

```
$ go test -tags sqlite_load_extension -count=1 ./packages/honker-go
ok  	github.com/russellromney/honker-go	17.716s
```

### Load-bearing proof

Dropped `Priority: r.Priority` from the `ClaimBatch` row mapping (so the field
falls back to Go's zero value) and reran:

```
--- FAIL: TestClaimedJobAndSnapshotCarryEveryField (0.12s)
    typed_jobs_test.go:131: job.Priority = 0, want 7
FAIL	github.com/russellromney/honker-go	0.430s
```

Restored, full suite green again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,26 @@
 - Regression test claims a job at a deadline with one claim forced to
   return empty: it fails on the unpatched `api.js` and passes on the fix.
 
+## Unreleased — full job details in the Go binding
+
+- Go `Job` now carries every field `honker_claim_batch` returns: `State`,
+  `Priority`, `RunAt`, `ClaimExpiresAt`, `MaxAttempts`, `CreatedAt`, and
+  `ExpiresAt` join the existing `ID`, `Queue`, `Payload`, `WorkerID`, and
+  `Attempts`. A claimed `Job` and a `GetJob` snapshot (`JobRow`) now describe
+  the same twelve columns; `Job` keeps `Ack` / `Retry` / `Fail` / `Heartbeat`
+  and `JobRow` stays data only. Purely additive — no existing field changed
+  name, type, or meaning.
+- New `honker.DecodePayload[T]` unmarshals a payload into a caller-owned type,
+  plus `JobRow.PayloadBytes()` so a snapshot decodes through the same path as
+  a claimed job. The type parameter is a compile-time contract; Honker still
+  does not validate payload shape in the database.
+- `TestClaimedJobAndSnapshotCarryEveryField` enqueues with a priority, TTL, and
+  max-attempts, asserts each of the twelve fields on the pending snapshot, the
+  claimed job, and a second `Database` handle's processing snapshot, then
+  asserts the reader gets nothing after ack.
+  `TestDelayedJobReportsItsRunAt` asserts a delayed job's `RunAt` and that it
+  is not claimable early.
+
 ## Unreleased — typed Node jobs
 
 - Node `Queue`, `Job`, `JobSnapshot`, `ClaimWaker`, and `Outbox` APIs now carry
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,33 @@
 - `DecodePayload` now returns an error on an empty payload instead of the zero
   value with a nil error. The core never emits an empty payload, so empty means
   the bytes never arrived, and a caller could not tell that apart from a clean
-  decode.
+  decode. That error is the exported sentinel `honker.ErrEmptyPayload`, wrapped
+  with `%w`, so a caller can tell "the bytes never arrived" from "the bytes did
+  not match `T`" with `errors.Is` instead of matching message text.
+- Documented the claim-time snapshot semantics of `Job`: `Ack` / `Retry` /
+  `Fail` / `Heartbeat` change the row in the database and do not write back to
+  the struct, so `job.ClaimExpiresAt` stays at its claim-time value after a
+  heartbeat. Also documented that `Queue.GetJob` is not queue-scoped, which
+  matters now that the README points it at `DecodePayload`: another queue's
+  payload decodes into your type with zero-valued fields and no error.
+- Go README: the quick start, the delayed-job example, and the scheduler
+  example did not compile (`db.Queue("emails")`, `q.Enqueue(payload)`,
+  `honker.RunAt(...)` — which does not exist — and the old positional
+  `Scheduler.Add`). Rewritten against the real signatures. The new typed-payload
+  example also called `row.PayloadBytes()` on the result of `q.GetJob(id)`
+  without checking for `nil`, which panics on a miss; it now checks the error
+  and the nil row.
+- `TestClaimBatchKeepsEachRowsOwnFields` claims two jobs in one `ClaimBatch`
+  call and pins each returned `Job` to its own row — id, priority, payload,
+  `ExpiresAt` set vs nil — then acks each job through its own handle. Every
+  other test claims a single job, so a mapping that reads the same element for
+  every job was invisible: `ClaimBatch` could hand every job in a batch the
+  first row's fields and the whole suite stayed green.
+- `TestClaimedJobReportsBackDatedRunAt` and `TestClaimBatchKeepsEachRowsOwnFields`
+  assert `ExpiresAt` is `nil`, not `0`, on a job enqueued without a TTL — on
+  both the claimed `Job` and the `JobRow`. No test covered the null case before,
+  so decoding `expires_at: null` as `0` (a job that "expired in 1970") passed
+  the whole suite.
 - `TestClaimedJobAndSnapshotCarryEveryField` enqueues with a priority, TTL, and
   max-attempts, asserts each of the twelve fields on the pending snapshot, the
   claimed job, and a second `Database` handle's processing snapshot, then
@@ -47,7 +73,9 @@
   `RunAt` (`now - 100`) so the job stays claimable while `RunAt` and `CreatedAt`
   differ, then pins both on the claimed job. Without it the two `json:` tags in
   the claim decode can be transposed and the suite stays green.
-  `TestDecodePayloadRejectsEmptyInput` covers the empty-payload error.
+  `TestDecodePayloadRejectsEmptyInput` covers the empty-payload error, checks
+  it with `errors.Is(err, ErrEmptyPayload)`, and checks that a shape mismatch
+  does *not* match the sentinel.
 
 ## Unreleased — typed Node jobs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,12 +33,17 @@
   plus `JobRow.PayloadBytes()` so a snapshot decodes through the same path as
   a claimed job. The type parameter is a compile-time contract; Honker still
   does not validate payload shape in the database.
+- `DecodePayload` now returns an error on an empty payload instead of the zero
+  value with a nil error. The core never emits an empty payload, so empty means
+  the bytes never arrived, and a caller could not tell that apart from a clean
+  decode.
 - `TestClaimedJobAndSnapshotCarryEveryField` enqueues with a priority, TTL, and
   max-attempts, asserts each of the twelve fields on the pending snapshot, the
   claimed job, and a second `Database` handle's processing snapshot, then
   asserts the reader gets nothing after ack.
   `TestDelayedJobReportsItsRunAt` asserts a delayed job's `RunAt` and that it
   is not claimable early.
+  `TestDecodePayloadRejectsEmptyInput` covers the empty-payload error.
 
 ## Unreleased — typed Node jobs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,10 @@
   asserts the reader gets nothing after ack.
   `TestDelayedJobReportsItsRunAt` asserts a delayed job's `RunAt` and that it
   is not claimable early.
+  `TestClaimedJobReportsBackDatedRunAt` enqueues with a back-dated absolute
+  `RunAt` (`now - 100`) so the job stays claimable while `RunAt` and `CreatedAt`
+  differ, then pins both on the claimed job. Without it the two `json:` tags in
+  the claim decode can be transposed and the suite stays green.
   `TestDecodePayloadRejectsEmptyInput` covers the empty-payload error.
 
 ## Unreleased — typed Node jobs
--- a/packages/honker-go/README.md
+++ b/packages/honker-go/README.md
@@ -29,10 +29,11 @@ db, err := honker.Open("app.db", "./libhonker_ext.dylib")
 if err != nil {
     panic(err)
 }
+defer db.Close()
 
-q := db.Queue("emails")
+q := db.Queue("emails", honker.QueueOptions{})
 
-if _, err := q.Enqueue(map[string]any{"to": "alice@example.com"}); err != nil {
+if _, err := q.Enqueue(map[string]any{"to": "alice@example.com"}, honker.EnqueueOptions{}); err != nil {
     panic(err)
 }
 
@@ -42,7 +43,7 @@ if err != nil {
 }
 if job != nil {
     sendEmail(job.Payload)
-    _ = job.Ack()
+    _, _ = job.Ack()
 }
 ```
 
@@ -54,6 +55,15 @@ shape: `ID`, `Queue`, `Payload`, `State`, `Priority`, `RunAt`, `WorkerID`,
 A `Job` also has the claim methods (`Ack`, `Retry`, `Fail`, `Heartbeat`);
 a `JobRow` is data only.
 
+`ExpiresAt` is `*int64` on both types and is `nil` — never `0` — when the job
+was enqueued without a TTL. On a `JobRow`, `WorkerID` and `ClaimExpiresAt` are
+`nil` until a worker claims the job. On a claimed `Job` both are always set,
+so they are plain values.
+
+Every field on a `Job` is a snapshot taken at claim time. `Heartbeat` extends
+the claim in the database but does not refresh `job.ClaimExpiresAt`; re-read
+with `GetJob` if you need the new deadline.
+
 Payloads stay raw JSON. `DecodePayload[T]` unmarshals one into your own type:
 
 ```go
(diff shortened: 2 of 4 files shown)


russellromney:
Adversarial review (I did not write this code). Everything below was RUN unless marked INFERRED, in a detached worktree at `bc642ba` with a prebuilt `target/release/libhonker_ext.dylib`.

Baseline: both new tests actually executed — `TestClaimedJobAndSnapshotCarryEveryField (0.37s)`, `TestDelayedJobReportsItsRunAt (0.11s)`. Not the silent `findExtension` skip.

### 1. HIGH — the claimed `Job` can have `RunAt` and `CreatedAt` swapped and both new tests still pass. RAN.

`ClaimBatch`'s decode struct is the only place the claim JSON is mapped, and nothing in the suite can distinguish the two keys there:

- `TestClaimedJobAndSnapshotCarryEveryField` enqueues with no delay and no `RunAt`, so `run_at == created_at` to the second. It asserts `job.RunAt == pending.RunAt` and `job.CreatedAt == pending.CreatedAt` — both true under a swap.
- `TestDelayedJobReportsItsRunAt` is the only test with `run_at != created_at`, and it never claims. It asserts the `JobRow` snapshot and then that the job is *not* claimable.

Mutation applied to `packages/honker-go/honker.go` (the `var raw []struct` in `ClaimBatch`):

```
-		RunAt          int64  `json:"run_at"`
+		RunAt          int64  `json:"created_at"`
-		CreatedAt      int64  `json:"created_at"`
+		CreatedAt      int64  `json:"run_at"`
```

```
$ go test -run 'TestClaimedJobAndSnapshotCarryEveryField|TestDelayedJobReportsItsRunAt' .
ok  	github.com/russellromney/honker-go	0.561s
```

Still green with the two fields transposed on every claimed job.

Failure a user hits: a worker on a delayed or scheduled queue computes lateness as `time.Now().Unix() - job.RunAt`, or drives a retry/deadline decision off `RunAt`, and silently gets the enqueue time instead. Wrong by the whole delay, no error, green CI. Job-age metrics off `CreatedAt` report the future scheduled time.

Cheap fix: `EnqueueOptions.RunAt` already exists (honker.go:619). Back-date one job — `RunAt: &past` — claim it, and assert `job.RunAt == past` and `job.RunAt != job.CreatedAt`. That is exactly the shape #142's untyped tests use and it catches this in one line.

### 2. LOW — `DecodePayload` returns the zero value with a nil error on empty input.

```go
if len(payload) == 0 {
    return out, nil
}
```

A `JobRow` whose `Payload` failed to populate (a future core rename, a tag typo, a truncated blob) decodes to a zero-valued `T` and no error. That is a silent default on the data the caller actually cares about. Returning an error there costs nothing — core never emits an empty payload, so no legitimate call site is affected.

### 3. LOW — pre-existing `Queue.GetJob` doc becomes wrong under #134. Not introduced here.

`Queue.GetJob`'s comment still says a miss means "ack'd, dead'd, or never existed". Once #134 scopes it, that list needs "or belongs to a different queue". Compare the reference `packages/honker-node/wrapper.d.ts`. Nothing in this PR **conflicts** with adding the scoping — no test asserts global lookup, and `Queue.GetJob` is a single method body to filter — but the new README example (`row, _ := q.GetJob(id)` → `honker.DecodePayload[Email](row.PayloadBytes())`) is exactly the pattern wrapper.d.ts calls out as unsound while the lookup is global: a foreign queue's row decodes into `Email` with zero-valued fields and no error (see finding 2). Worth a sentence in the README until #134 lands.

- All twelve JSON keys on both `Job` and `JobRow` read against `honker_ops.rs` `claim_batch` / `get_job`. All correct.
- `attempts` vs `max_attempts` (1 vs 5), `claim_expires_at` vs `expires_at` (claim+120 vs run_at+600), `state`, `priority`, `queue`: fixture values are distinct, all swaps would be caught.
- `JobRow` run_at/created_at is covered — `TestDelayedJobReportsItsRunAt` pins `row.RunAt` to `before+3600`.
- Nulls: `pending.WorkerID`, `pending.ClaimExpiresAt` asserted `nil` (not `0`/`""`); `ExpiresAt` asserted `nil`-vs-set per row. Correct.
- `Job.ClaimExpiresAt` being a plain `int64` while `JobRow`'s is `*int64` is right — the claim UPDATE always writes it — and it is documented.

Would not merge as-is: finding 1 is a one-line fixture change and the test is currently not load-bearing for the field it advertises.

russellromney:
Both findings fixed. CI green on `adcd18a` (run 33585612331).

### Finding 1 — claimed `Job.RunAt` / `CreatedAt` transposable

`test(go): pin a claimed job's RunAt apart from its CreatedAt` — commit `adcd18a`.

`TestClaimedJobReportsBackDatedRunAt` enqueues with `EnqueueOptions{RunAt: &past}` where `past = now - 100`. The job stays claimable — a delayed job is not, which is exactly why the old fixture proved nothing — while `RunAt` and `CreatedAt` differ by 100 seconds. It pins `RunAt` to that exact timestamp on both the `JobRow` snapshot and the claimed `Job`, keeps `CreatedAt` in the enqueue window, and asserts the two are not equal.

Proof it is load-bearing. Your exact mutation, the two `json:` tags swapped in `ClaimBatch`'s decode struct (honker.go:710,715):

```
$ go test -run 'TestClaimedJobReportsBackDatedRunAt|TestClaimedJobAndSnapshotCarryEveryField|TestDelayedJobReportsItsRunAt' -v .
--- PASS: TestClaimedJobAndSnapshotCarryEveryField (0.11s)
--- PASS: TestDelayedJobReportsItsRunAt (0.11s)
    typed_jobs_test.go:276: job.RunAt = 1788318155, want 1788318055 (the back-dated absolute run_at)
    typed_jobs_test.go:279: job.CreatedAt = 1788318055, want within [1788318155, 1788318155]
    typed_jobs_test.go:286: claimed gap CreatedAt-RunAt = -100, snapshot gap = 100, want equal
--- FAIL: TestClaimedJobReportsBackDatedRunAt (0.11s)
FAIL	github.com/russellromney/honker-go	0.661s
```

The two pre-existing tests still pass under the mutation, as you found. The new one catches it.

### Finding 2 — `DecodePayload` on empty input

`fix(go): reject an empty payload in DecodePayload` — commit `e5f49bf`.

`return out, nil` became `return out, errors.New("decode payload: empty payload")`, with the reason in the doc comment.

Proof, with the guard reverted to `return out, nil`:

```
$ go test -run 'TestDecodePayloadRejectsEmptyInput' -v .
    typed_jobs_test.go:308: DecodePayload(nil) = {Recipient: Template: Version:0}, nil — want an error
    typed_jobs_test.go:308: DecodePayload(empty) = {Recipient: Template: Version:0}, nil — want an error
--- FAIL: TestDecodePayloadRejectsEmptyInput (0.00s)
```

Baseline confirmation the tests actually execute rather than hitting `findExtension`'s silent skip: `-v` shows `RUN`/`PASS` lines with real runtimes (0.16s for the claim test), against a `libhonker_ext.dylib` built in the worktree and checked with `strings … | grep -c "RETURNING id, queue, payload, state"` → 1.

Full suite: `ok github.com/russellromney/honker-go 16.984s`.

Finding 3 (the `GetJob` doc under #134) left alone — it is not introduced by this PR.

russellromney:
Second-round adversarial review — completeness, design, and "does this make sense to a Go user", not just correctness. I did not write this code.

Everything below is marked **RAN** or **READ**. RAN work used the prebuilt `target/debug/libhonker_ext.dylib` symlinked into a worktree; I confirmed it carries the widened clause (`strings … | grep -c "RETURNING id, queue, payload, state, priority, run_at, worker_id"` → 1) and that the tests actually execute rather than hitting `findExtension`'s silent skip (`-v` shows real per-test durations, 0.11–0.22s). No Rust was built. Fixes pushed to `feat/go-job-parity` as `b93249e`, `97633b1`, `286024e`, `b632072`, `b234473`.

I also confirmed the Go tests really run in CI, which is not obvious from `gh pr checks`: there is no `go` job, Go runs inside `binding smoke · ubuntu aggregate`, and run 33585612331 logs `ok github.com/russellromney/honker-go 29.653s`. **READ** (CI log).

---

## HIGH — two mutations that pass the entire suite. RAN.

### 1. `expires_at: null` can decode as `0` on both the claimed `Job` and the `JobRow`, and nothing fails

Every test that touched `ExpiresAt` enqueued *with* a TTL. Nothing anywhere asserted the NULL case. The first review said "`ExpiresAt` asserted nil-vs-set per row" — that is not so; it is only ever asserted set.

Mutation applied to `packages/honker-go/honker.go`: claim decode `ExpiresAt *int64` → `int64` with `ExpiresAt: &r.ExpiresAt`, plus `GetJob` filling a nil `row.ExpiresAt` with a pointer to 0.

```
$ go test -tags sqlite_load_extension -count=1 ./packages/honker-go
ok  	github.com/russellromney/honker-go	17.672s
```

Full suite green with every TTL-less job reporting "expires_at = 0" — a job that expired in 1970. That is exactly the nullability class this issue is about (`worker_id`, `claim_expires_at`, `expires_at` must be nil, never 0 or `""`).

**Fixed** in `97633b1`. `TestClaimedJobReportsBackDatedRunAt` (no TTL) now asserts `row.ExpiresAt == nil` and `job.ExpiresAt == nil`; the new batch test below asserts it too. Under the same mutation:

```
--- FAIL: TestClaimedJobReportsBackDatedRunAt (0.13s)
    typed_jobs_test.go:271: row.ExpiresAt = 0, want nil (no TTL was set)
    typed_jobs_test.go:302: job.ExpiresAt = 0, want nil (no TTL was set)
--- FAIL: TestClaimBatchKeepsEachRowsOwnFields (0.13s)
    typed_jobs_test.go:373: low: job.ExpiresAt = 0, want nil (no TTL)
```

### 2. `ClaimBatch` can give every job in a batch the first row's fields, and nothing fails

`ClaimBatch` is the only place the twelve fields are mapped, and it maps them in a loop — but every test in `typed_jobs_test.go` claims exactly one job, and `TestAckBatch` acks by the ids `Enqueue` returned, never by `job.ID`. So the per-row mapping was untested.

Mutation: `for i, r := range raw` → `for i := range raw { r := raw[0]`.

```
$ go test -tags sqlite_load_extension -count=1 ./packages/honker-go
ok  	github.com/russellromney/honker-go	17.564s
```

The user-visible damage is `Ack`: `job.Ack()` uses the `Job`'s own `ID`, so `jobs[1].Ack()` acks row 0 a second time, returns false, and `jobs[1]`'s real claim is left to expire and be retried.

A narrower single-field version is just as invisible — `Payload: []byte(raw[0].Payload)` alone also passed everything.

**Fixed** in `97633b1`. `TestClaimBatchKeepsEachRowsOwnFields` enqueues two jobs at priority 9 and 1 (priority DESC makes the claim order deterministic, no timing race), claims both in one `ClaimBatch`, pins each returned `Job` to its own row (id, priority, payload, `ExpiresAt` set vs nil, worker, attempts, max attempts), then acks each through its own handle and checks both rows are gone. Under the two mutations:

```
--- FAIL: TestClaimBatchKeepsEachRowsOwnFields (0.12s)
    typed_jobs_test.go:345: both claimed jobs carry id 1 — the batch mapping collapsed the rows
--- FAIL: TestClaimBatchKeepsEachRowsOwnFields (0.12s)
    typed_jobs_test.go:369: low: payload = {Recipient:high@example.com …}, want {Recipient:low@example.com …}
```

---

## HIGH — the README does not compile, and the new section teaches a panic. RAN.

Lens: would a Go developer reading only the README and godoc get this right on the first try? No. I compiled every Go snippet in `packages/honker-go/README.md` verbatim in a scratch module with a `replace` onto this package:

```
./old.go:15:16: not enough arguments in call to db.Queue
	have (string); want (string, honker.QueueOptions)
./old.go:17:25: not enough arguments in call to q.Enqueue
	have (map[string]any); want (any, honker.EnqueueOptions)
./old.go:21:69: undefined: honker.RunAt
./old.go:24:20: too many arguments in call to s.Add
	have (string, string, string, map[string]any); want (honker.ScheduledTask)
```

Three of those four are pre-existing (quick start, delayed jobs, scheduler) — I am fixing them anyway because this PR inserts its new section into the middle of that page and the question the review asks is exactly whether the page works. The package-level godoc at the top of `honker.go` and `examples/basic` were already correct; only the README had drifted.

The fourth problem is introduced here:

```go
row, _ := q.GetJob(id)
email, err := honker.DecodePayload[Email](row.PayloadBytes())
```

`GetJob` returns `(nil, nil)` on a miss, so this nil-derefs and panics the worker the first time a job is acked concurrently. The same shape was in `DecodePayload`'s godoc.

**Fixed** in `b632072` (README) and `286024e` (godoc). All snippets now compile (`go build` exit 0) and the `GetJob` example checks the error and the nil row.

---

## MEDIUM — `job.ClaimExpiresAt` is silently stale after `Heartbeat`. READ.

`Heartbeat` extends the claim in the database; the `Job` struct is never written back. A worker doing the obvious thing — heartbeat on a long job, then check `time.Now().Unix() < job.ClaimExpiresAt` — gets the pre-heartbeat deadline and concludes it is out of time. Plausible-looking wrong value, no error. The same applies to `Attempts` and `State` after `Retry`.

Node has the identical behavior (`api.js:286` just forwards to the core), so this is not a Go divergence — it is undocumented in both. Per the Go-only constraint I documented it here only.

**Fixed** in `286024e`: the `Job` doc says every field is a claim-time snapshot that the claim methods do not write back, `Heartbeat`'s doc says it does not refresh `ClaimExpiresAt` and to re-read with `GetJob`, and the README repeats it. I did not make `Heartbeat` write back a new deadline: `honker_heartbeat` returns a row count, not the new `claim_expires_at`, so that would need a core change.

## MEDIUM — the empty-payload error could not be matched programmatically. RAN.

Round one's fix returned `errors.New("decode payload: empty payload")`. A caller who wants to treat "the bytes never arrived" differently from "the bytes did not match `T`" — the whole reason the guard exists — had to match on message text. Not idiomatic Go for a library.

**Fixed** in `b93249e`: exported `honker.ErrEmptyPayload`, wrapped with `%w`. The test now asserts `errors.Is(err, ErrEmptyPayload)` on the empty cases *and* asserts a shape mismatch does **not** match the sentinel, so it is load-bearing both ways. With the wrap reverted to `errors.New`:

```
--- FAIL: TestDecodePayloadRejectsEmptyInput/nil
    typed_jobs_test.go:434: DecodePayload(nil) err = decode payload: empty payload, want errors.Is(err, ErrEmptyPayload)
```

## MEDIUM — `Queue.GetJob` is not queue-scoped and the new README aims it at `DecodePayload`. READ.

Round one raised this and it was left as "not introduced here". The lookup being global is indeed pre-existing (`honker_get_job` is `WHERE id = ?1`, no queue filter) and #134 owns the fix — but this PR is the one that adds `honker.DecodePayload[Email](row.PayloadBytes())` to the README, which turns a foreign-queue hit into a zero-valued `Email` with a nil error. Node's `wrapper.d.ts` documents its scoping explicitly; Go documented nothing.

**Fixed by documenting only** (`286024e`, `b632072`). No behavior change, no queue filter added — that stays #134.

---

## Attacked and could not break

- **All twelve JSON tags on both types**, re-read against `honker_ops.rs` `claim_batch` (line 652) and `get_job` (line 1006). Every key matches, and the Go nullability matches the Rust column types exactly (`Option<String>` / `Option<i64>` → `*string` / `*int64` on `JobRow`; the claim's non-optional `String` / `i64` → plain `string` / `int64` on `Job`). **READ**
- `Job.ClaimExpiresAt` as a plain `int64` while `JobRow`'s is `*int64` is right and is narrower than Node's `number | null`. The claim UPDATE always writes it, so a pointer would be a lie. Justified divergence. **READ**
- `DecodePayload[T]` as a package-level generic function is the only option — Go methods cannot have type parameters, so `Queue[T]` the way Node does it would force the type onto the handle. This is the right Go shape and matches what the issue asks for. **READ**
- `run_at` / `created_at` on the claimed job — round one's `TestClaimedJobReportsBackDatedRunAt` holds up. I re-ran the tag transposition and it still fails correctly. **RAN**
- Priority 7 vs max-attempts 5 vs attempts 1, `claim_expires_at` (claim+120) vs `expires_at` (run_at+600), `queue` vs `state`: all fixture values distinct, all swaps caught. **READ**
- `gofmt -l` clean on both changed files; `go vet` clean. The two files gofmt already flagged (`examples/atomic/main.go`, `parity_test.go`) are pre-existing and untouched. **RAN**
- Full suite after all fixes: `ok github.com/russellromney/honker-go 18.127s`. **RAN**

## Left alone, deliberately

- **`DecodePayload` on a JSON `null` payload** returns the zero value of `T` with a nil error. That is `json.Unmarshal`'s own behavior, `q.Enqueue(nil, …)` is a legitimate call, and rejecting `null` would break `DecodePayload[*Email]` and `DecodePayload[map[string]any]`. Documented rather than changed.
- **`Job.Payload []byte` vs `JobRow.Payload string`** — the asymmetry forces `row.PayloadBytes()` at one call site. It exists because `encoding/json` base64s a `[]byte` struct field, so `JobRow.Payload` cannot be `[]byte`. Widening `DecodePayload` to `[T any, P ~[]byte | ~string]` would remove the wart at the cost of a two-parameter signature in godoc. Not worth it; the reason is now in the `JobRow` doc.
- **`Outbox`'s delivery callback still takes only the payload**, so an outbox handler cannot see any of the twelve new fields. Node passes `(payload, job)`. Changing `OutboxDelivery` is a breaking signature change and is outside #136's scope — worth its own issue if anyone wants it.
- **`Queue.Cancel` is also unscoped** and also undocumented as such. Same #134 family; I only documented `GetJob` because that is what the new typed-decode example depends on.
- **`claimed_at`** (#135), **queue scoping** (#134), **payload encoding** (#146), **core and other bindings** — untouched, per the brief.
- **Cross-binding "one binding writes a versioned payload, another reads it"** from #136's test list — the Go↔Python interop test already round-trips queue payloads, and that checklist item is satisfied once across the issue rather than per binding. Not added here.

## Verdict

Merge, after CI. The twelve fields, their nullability, and the decode helper are now all pinned by tests I verified can fail, the two silent-wrong-value holes are closed, and the README a Go user actually reads compiles.

Issue #12, opened 2026-04-29
Testing regime: comprehensive cross-platform e2e gaps


russellromney:
This started as the broad audit of Honker's test and CI gaps. A lot of it is now done. This issue is now the tracker for the **remaining** work instead of the original wall of missing things.

## Completed since the issue was filed

- [x] Rust core runs on Linux, macOS, and Windows
- [x] Rust extension build + load path runs on Linux, macOS, and Windows
- [x] Python CI runs on Linux, macOS, and Windows
- [x] Node CI runs on Linux, macOS, and Windows
- [x] Smoke build coverage exists for the submodule bindings in the main pipeline:
  - [x] `packages/honker-bun`
  - [x] `packages/honker-cpp`
  - [x] `packages/honker-ex`
  - [x] `packages/honker-go`
  - [x] `packages/honker-rs`
  - [x] `packages/honker-ruby`
- [x] Crash-recovery / durability tests exist and run in CI
- [x] Journal-mode coverage exists for the watcher path
- [x] Debug-build CI coverage exists
- [x] Performance regression coverage exists in CI (`bench/wal_index_methods`)
- [x] Cross-binding proof is no longer one mixed test blob; direct proof and supporting proof are split
- [x] Ruby ↔ Python interop proof exists in CI

## Remaining gaps

- [ ] **Windows reverse listener parity**
  Node writes -> Python `listen()` on Windows is still not working in direct cross-language proof.
  Tracked in **#23**.

- [ ] **Broader cross-binding interop**
  We now prove more than one pairing, but we still do not have convincing coverage for the broader "any binding can share the same DB honestly" claim.

- [ ] **Concurrency / stress coverage**
  We still need higher-pressure multi-writer / multi-reader / many-subscriber stress beyond the current targeted tests.

- [ ] **Long-running soak / leak detection**
  A nightly or manual soak workflow is still missing.

- [ ] **SQLite version matrix**
  CI still mostly proves "the versions on the GitHub runners work" rather than a wider SQLite compatibility surface.

- [ ] **Coverage measurement**
  We still do not publish real line / branch coverage numbers.

- [ ] **Windows watcher timing behavior**
  The watcher cadence / timer-resolution difference on Windows is still worth tightening if it keeps shaping behavior or test fragility.

## What this issue is for now

Use this as the umbrella for the remaining test-regime work.

The immediate next phase is **#23** because it is the sharpest remaining user-visible parity gap:

- direct proof already passes on Linux and macOS
[...]

PR #140, opened 2026-09-01
feat(core): add claimed_at to _honker_live so attempt runtime is answerable

### LINKED ISSUE #135: No claimed_at, so job running time is unanswerable

russellromney:
Closes #135.

> **Stacked PR — must merge second.** Base is `fix/propagate-sqlite-errors` (#138), not `main`. Both branches touch `honker_ops.rs` and conflict if merged out of order. Merge #138 first, then this.

`_honker_live` had no column recording when the current attempt started, so Honker could not report how long a job had been running. None of the existing timestamps answer it:

- `created_at` is total job age and includes queue wait
- `run_at` is when the job became ready to run
- `claim_expires_at` moves on every heartbeat

## Change

New nullable `claimed_at INTEGER` on `_honker_live`.

| transition | `claimed_at` |
| --- | --- |
| enqueued, never claimed | `NULL` |
| first claim | `unixepoch()` |
| heartbeat | **unchanged** |
| retry back to pending | `NULL` |
| reclaim after visibility timeout | `unixepoch()` (new attempt) |

Set in `claim_batch`'s UPDATE, which is the single statement handling both first claim and reclaim, so both are covered by one change. Cleared in `retry`'s pending-flip. `heartbeat()` is untouched by design and now says so in its doc comment — refreshing `claimed_at` on a heartbeat would make a long-running job look like it just started, which is the exact blind spot this column exists to fix.

Exposed in `claim_batch`'s `RETURNING` and JSON, and in `get_job`'s JSON.

## Migration

Existing databases get the column through `ALTER TABLE _honker_live ADD COLUMN claimed_at INTEGER`, using the same `pragma_table_info` check and `"duplicate column"` tolerance as the existing `enabled` and `max_attempts` migrations in `lib.rs`. `CREATE TABLE IF NOT EXISTS` does not add columns to a table that already exists, so the ALTER is the only path for an existing database.

Nullable with no default on purpose: a default would date every pre-existing row to migration time and read as a claim that never happened.

The column is declared last in `BOOTSTRAP_HONKER_SQL` so a freshly created table and an ALTER-migrated one agree on column order.

## Binding impact

None required, and none made — core only. Both JSON changes are additive, and no binding declares `serde(deny_unknown_fields)` (checked across `packages/` and `honker-extension/`), so existing consumers ignore the new field until they opt in. Issue #135 leaves the per-binding rollout open; that is follow-up work.

## Tests

```
claimed_at_is_null_until_the_first_claim ... ok
first_claim_sets_claimed_at_and_returns_it ... ok
heartbeat_moves_the_deadline_but_not_claimed_at ... ok
retry_clears_claimed_at_when_the_job_goes_back_to_pending ... ok
reclaim_resets_claimed_at_to_the_new_attempt ... ok
ack_and_fail_do_not_need_claimed_at_but_still_work ... ok
bootstrap_pre_claimed_at_database_gains_the_column_and_keeps_its_rows ... ok

test result: ok. 74 passed; 0 failed; 1 ignored
```

`bootstrap_honker_schema_creates_tables_and_index` updated from 12 to 13 `_honker_live` columns.

## Proof the tests are load-bearing

**Break A — add `claimed_at = unixepoch()` to `heartbeat()`'s UPDATE** (the forbidden change):

```
heartbeat_moves_the_deadline_but_not_claimed_at ... FAILED
  assertion `left == right` failed: a heartbeat must NOT refresh claimed_at:
  it extends the claim, it does not start a new attempt
    left: Some(1788254358)
   right: Some(1788250758)
```

Worth calling out: **the first version of this test passed under Break A.** The claim and the heartbeat landed in the same second, so a heartbeat that did refresh `claimed_at` wrote back an identical value and the equality assertion could not see it. The test now backdates `claimed_at` by an hour before the heartbeat, and a comment in the test records why. Break A was re-run after the fix and fails as shown.

**Break B — drop `claimed_at = NULL` from `retry`'s pending-flip:**

```
retry_clears_claimed_at_when_the_job_goes_back_to_pending ... FAILED
  assertion `left == right` failed: a job waiting in the queue is not running,
  so claimed_at must be cleared
```

**Break C — remove the ALTER TABLE migration, leaving only `CREATE TABLE IF NOT EXISTS`:**

```
bootstrap_pre_claimed_at_database_gains_the_column_and_keeps_its_rows ... FAILED
  claimed_at must be added to a table that already existed;
  CREATE TABLE IF NOT EXISTS cannot do it
```

**Break D — drop `claimed_at = unixepoch()` from `claim_batch`'s UPDATE:**

```
test result: FAILED. 2 passed; 5 failed
```

All four breaks were restored and the suite is green.

## Scope

`honker-core/src/lib.rs`, `honker-core/src/honker_ops.rs`, `CHANGELOG.md`. No bindings touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,31 @@
 # CHANGELOG
 
+## Unreleased — `claimed_at` on `_honker_live`
+
+- New nullable `claimed_at INTEGER` column on `_honker_live`: when the
+  CURRENT attempt started. Nothing else could answer that.
+  `created_at` includes queue wait, `run_at` is when the job became
+  ready, and `claim_expires_at` moves on every heartbeat.
+- NULL until the first claim. Set to `unixepoch()` on every successful
+  claim and reclaim, so it tracks the current attempt and not the first
+  one. Cleared when `retry` returns the job to pending — a job waiting
+  in the queue is not running. `heartbeat()` deliberately leaves it
+  alone; refreshing it there would reintroduce exactly the blind spot
+  the column exists to fix.
+- Exposed in `claim_batch`'s RETURNING and JSON, and in `get_job`'s
+  JSON. Both are additive; no binding declares
+  `serde(deny_unknown_fields)`, so existing consumers ignore it until
+  they opt in.
+- Existing databases migrate with `ALTER TABLE ... ADD COLUMN`, matching
+  the `enabled` and `max_attempts` migrations, and tolerating the
+  "duplicate column" error when a concurrent bootstrap wins the race.
+  `CREATE TABLE IF NOT EXISTS` cannot add a column to a table that
+  already exists, which is what the migration test pins.
+- Tests: NULL before first claim, first claim, heartbeat-does-not-move,
+  retry clears, reclaim resets, plus a migration test proving a
+  pre-column database gains the column and keeps its rows with
+  `claimed_at` NULL rather than a backfilled timestamp.
+
 ## Unreleased — Core SQLite error propagation
 
 - `honker-core` no longer discards SQLite errors in five lookups.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@
   JSON. Both are additive; no binding declares
   `serde(deny_unknown_fields)`, so existing consumers ignore it until
   they opt in.
+- Validity window, documented in README under "How long has this job
+  been running": `claimed_at` is the start of the CURRENT claim and is
+  only meaningful while `claim_expires_at >= unixepoch()`. A claim that
+  lapses without a reclaim leaves `worker_id`, `claim_expires_at` and
+  `claimed_at` on the row, all stale together, until the next claim
+  overwrites all three. Nothing clears them, by design — there is no
+  expiry sweep for processing rows, and blanking `claimed_at` alone
+  would throw away the abandoned attempt's start time while leaving the
+  other two stale anyway.
 - Existing databases migrate with `ALTER TABLE ... ADD COLUMN`, matching
   the `enabled` and `max_attempts` migrations, and tolerating the
   "duplicate column" error when a concurrent bootstrap wins the race.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,10 +30,17 @@
   "duplicate column" error when a concurrent bootstrap wins the race.
   `CREATE TABLE IF NOT EXISTS` cannot add a column to a table that
   already exists, which is what the migration test pins.
+- No language binding maps `claimed_at` onto its job type yet, so for
+  now it is read in SQL. #136 tracks adding it to the full job shape
+  per binding.
 - Tests: NULL before first claim, first claim, heartbeat-does-not-move,
-  retry clears, reclaim resets, plus a migration test proving a
-  pre-column database gains the column and keeps its rows with
-  `claimed_at` NULL rather than a backfilled timestamp.
+  retry clears, reclaim resets to the reclaim time (bounded by a clock
+  read taken either side of the reclaim, not just "moved off the old
+  value"), ack and fail both remove the row, an expired claim keeps
+  `claimed_at` alongside the rest of the stale claim, plus a migration
+  test proving a pre-column database gains the column, keeps its rows
+  with `claimed_at` NULL rather than a backfilled timestamp, and ends
+  up with the same columns in the same order as a fresh database.
 
 ## Unreleased — Core SQLite error propagation
 
--- a/README.md
+++ b/README.md
@@ -182,6 +182,34 @@ The extension shares tables with the language bindings, so a Python
 worker can claim jobs written by SQL, Node, Ruby, Go, or any other
 binding.
 
+### How long has this job been running
+
+`_honker_live.claimed_at` is when the CURRENT claim started. Nothing
+else answers that: `created_at` includes queue wait, `run_at` is when
+the job became ready, and `claim_expires_at` moves on every heartbeat.
+No language binding exposes `claimed_at` yet, so read it in SQL:
+
+```sql
+SELECT id, queue, unixepoch() - claimed_at AS running_s
+  FROM _honker_live
+ WHERE state = 'processing'
+   AND claim_expires_at >= unixepoch();   -- required, see below
+```
+
+`claimed_at` is only meaningful while `claim_expires_at >= unixepoch()`.
+When a claim lapses and nobody reclaims the job — attempts exhausted, no
+worker on that queue, queue drained — honker does not touch the row.
+`worker_id`, `claim_expires_at` and `claimed_at` all stay put and all go
+stale together, and the next claim overwrites all three. Without the
+`claim_expires_at` filter, `unixepoch() - claimed_at` keeps counting up
+for an attempt nobody is running.
+
+`claimed_at` is NULL until the first claim, and stays NULL for jobs that
+were already in flight when an existing database was upgraded: the
+migration adds the column without backfilling, because a backfill would
+date those rows to upgrade time and read as a claim that never happened.
+They pick up a real value on their next claim.
+
 ## Architecture
 
 - One `PRAGMA data_version` watcher per `Database`; the default
--- a/honker-core/src/honker_ops.rs
+++ b/honker-core/src/honker_ops.rs
@@ -656,6 +656,16 @@ fn dead_letter_exhausted_claimable(conn: &Connection, queue: &str) -> rusqlite::
 /// first one. `heartbeat()` deliberately leaves it alone — that is the
 /// whole reason `claim_expires_at` cannot answer "how long has this
 /// been running".
+///
+/// Here it is always a number: the UPDATE just wrote it on every row
+/// this RETURNING sees. `get_job` is the one that can report
+/// `"claimed_at": null`, for a job that has never been claimed.
+///
+/// Validity window: `claimed_at` is the start of the CURRENT claim and
+/// is only meaningful while `claim_expires_at >= unixepoch()`. When a
+/// claim lapses without a reclaim, `worker_id`, `claim_expires_at` and
+/// `claimed_at` all stay on the row and all go stale together; the
+/// next claim overwrites the three of them.
 pub fn claim_batch(
     conn: &Connection,
     queue: &str,
@@ -1014,6 +1024,12 @@ pub fn cancel(conn: &Connection, job_id: i64) -> rusqlite::Result<i64> {
 /// Read a single job row by id. Returns a JSON object on success or
 /// the empty string on miss (job ack'd, dead'd, or never existed).
 /// Pure read — does not change state.
+///
+/// `claimed_at` is null for a job that has never been claimed, and for
+/// a job that was already in flight when an existing database migrated
+/// (the migration adds the column without backfilling). Otherwise it is
+/// the start of the current claim, and it is only meaningful while
+/// `claim_expires_at >= unixepoch()` — see `claim_batch`.
 pub fn get_job(conn: &Connection, job_id: i64) -> rusqlite::Result<String> {
     let row: Option<(
         i64,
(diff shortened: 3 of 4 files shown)


russellromney:
Adversarial review. Throwaway worktree at 6780cae, probes appended and reverted after.

**Attacked and could not break (all measured):**

- **Migration, full round trip.** Built a `_honker_live` with the pre-`claimed_at` 12 columns, bootstrapped, then ran enqueue -> claim -> `get_job` -> heartbeat -> retry on it. All work; `claimed_at` is returned by the claim and by `get_job`, and is NULL after retry.
- **Column order.** Fresh and migrated tables produce the identical 13-name list in the same `cid` order, `claimed_at` last. The claim in the PR body holds.
- **Mutation checks, re-run independently.** Adding `claimed_at = unixepoch()` to `heartbeat()`'s UPDATE fails exactly one test, `heartbeat_moves_the_deadline_but_not_claimed_at` (the backdate fix works — this is the one that used to false-pass). Dropping `claimed_at = unixepoch()` from `claim_batch` fails 5 of the 6 `claimed_at_tests`.
- **Binding blast radius.** No binding carries its own `_honker_live` DDL, none does `SELECT *` or positional column reads on it, and every payload write goes through the SQL functions. The added column and the two added JSON keys are additive.

### 1. (low) `claimed_at` is not cleared when a claim just expires

Measured: claim with a timeout already in the past, then `get_job` returns `claim_expires_at` in the past *and* `claimed_at` set. Nobody is running that attempt, but anything computing "running for N seconds" from `claimed_at` keeps counting up. Reclaim fixes it whenever it happens — which for a row nobody claims again (attempts exhausted, no worker on the queue, queue drained) is never.

Not a blocker and arguably consistent with `state='processing'`, but the column exists to answer one question and this is the case where it answers wrong. A doc line would do: `claimed_at` is the start of the current attempt and is only meaningful while `claim_expires_at >= now`.

### 2. (low) `ack_and_fail_do_not_need_claimed_at_but_still_work` never calls `fail()`

It enqueues, claims, acks. Rename it or add the fail leg. It does fail under Break D, but incidentally: `claim_batch`'s `RETURNING claimed_at` decodes as a non-null `i64`, so the helper panics before the assertion is reached.

### 3. (nit) `reclaim_resets_claimed_at_to_the_new_attempt` asserts a loose bound

`second > first - 3600` only proves the value moved off the backdated one. `second >= before_reclaim` (a `unixepoch()` read taken before the reclaim) pins the actual claim. It does catch Break D as written, so this is style.

**Would I merge:** yes, after #138 lands, as stacked. Nothing here blocks.

russellromney:
Round two. I did not write this code. Worked on `feat/claimed-at`, pushed `e9bb06c` and `fd063c9`. Every claim below is marked RUN (I executed it) or INFERRED (I read it).

## The three open round-one findings

### 1. `claimed_at` is not cleared when a claim just expires — keeping the behaviour, documenting it

I reproduced it first (RUN). Claim with a timeout already in the past, then read the row:

```
state='processing'  claimed_at=1788554826  worker_id='w1'  claim_expires_at=1788554766
```

So the report is accurate. But the important part is the other two columns. `worker_id` and `claim_expires_at` are stale on that row too, and they were stale before this PR. There is no expiry sweep for `processing` rows — `sweep_expired` only touches `state='pending'` (INFERRED, read the SQL). Nothing runs at expiry. The whole claim goes stale together and the next claim overwrites all three at once.

That makes clearing `claimed_at` the wrong fix, not just a bigger one:

- It needs machinery that does not exist. There is no expiry event to hang it on. It would have to be a new sweep, or a write inside `get_job` (which is documented as a pure read), or a `CASE WHEN claim_expires_at >= unixepoch()` in the read that lies about what is in the table — the "assert the return code, not the row" trap.
- It destroys the answer to the second question. On an abandoned attempt, "when did the attempt that died start" is exactly what you want, and blanking the column throws it away.
- It makes the row internally inconsistent. `state='processing'`, `worker_id='w1'`, `claim_expires_at` set, `claimed_at=NULL` reads as a row that was never claimed. Today `claimed_at IS NOT NULL` iff `state='processing'` (RUN across every transition, below); clearing on expiry breaks that invariant to fix a subset of the staleness.

So: doc line, as suggested. Placed where a raw-SQL user will actually hit it, since no binding maps the column yet — a new **README → SQL Extension → "How long has this job been running"** section with the query you should run:

```sql
SELECT id, queue, unixepoch() - claimed_at AS running_s
  FROM _honker_live
 WHERE state = 'processing'
   AND claim_expires_at >= unixepoch();   -- required
```

plus the reason the filter is required, plus the migration case (a job already in flight when the database upgrades has `claimed_at` NULL until its next claim — the migration does not backfill). Also on `claim_batch` and `get_job` doc comments, and in CHANGELOG.

And a doc line that nothing tests is a doc line that rots, so it is now pinned by `an_expired_claim_keeps_claimed_at_with_the_rest_of_the_stale_claim`, which asserts the expired row still reports `worker_id` **and** `claimed_at`. Mutation F below shows exactly one test catches an attempt to "fix" it.

### 2. `ack_and_fail_do_not_need_claimed_at_but_still_work` never called `fail()` — fixed

Added the fail leg. Both legs now assert row state rather than only the return code: ack removes the row from `_honker_live`, fail removes it from `_honker_live` **and** lands it in `_honker_dead`.

### 3. `reclaim_resets_claimed_at_to_the_new_attempt` asserted a loose bound — fixed, and the hole was real

Now reads `unixepoch()` either side of the reclaim and asserts `before_reclaim <= second <= after_reclaim`.

I proved the old bound was hollow rather than assuming it (RUN). With `claim_batch` writing `claimed_at = unixepoch() - 1000`:

| assertion | result |
| --- | --- |
| old `second > first - 3600` | **passes** |
| new `before_reclaim <= second <= after_reclaim` | **fails** |

## Round two

### Mutation testing — re-ran round one's two, added four (all RUN)

I did not take round one's word for A and B. Both reproduce, and B is bigger than reported.

| # | mutation | tests that fail |
| --- | --- | --- |
| A | add `claimed_at = unixepoch()` to `heartbeat()`'s UPDATE | 1 — `heartbeat_moves_the_deadline_but_not_claimed_at` |
| B | drop `claimed_at = unixepoch()` from `claim_batch` | **14** — 6 of the 7 `claimed_at_tests`, plus 8 unrelated |
| C | move `claimed_at` above `expires_at` in the CREATE TABLE | 1 — the migration test (new assertion) |
| D | `claim_batch` writes `claimed_at = unixepoch() - 1000` | 2 — `first_claim_...`, `reclaim_resets_...` |
| E | `fail()` deletes the live row but never writes `_honker_dead` | 2 — incl. the repaired `ack_and_fail_...` (it passed under E before the fix) |
| F | `get_job` returns null `claimed_at` once the claim expires | 1 — `an_expired_claim_keeps_claimed_at_...` |

A confirms round one exactly. B's extra 8 failures are worth knowing about, and they point at a design note: `claim_batch` decodes `RETURNING claimed_at` as a non-null `i64` while `get_job` decodes the same column as `Option<i64>`. Today that is fine — the UPDATE writes the column on every row the RETURNING sees, so it is provably never NULL there (INFERRED from the SQL, and B is what happens if that stops being true: `InvalidColumnType`, and the whole batch errors out *after* the UPDATE committed). I left the strict decode (an impossible state should be loud) but documented the asymmetry on `claim_batch`, because a binding author implementing #136 would otherwise model the field as non-null from `claim_batch` and break on `get_job`.

### Migration

- **Idempotent** — RUN, the existing test bootstraps twice.
- **Duplicate-column race** — the guard is `to_lowercase().contains("duplicate column")`. SQLite 3.51 emits `duplicate column name: claimed_at`, so it matches (RUN, forced the error directly). The actual two-process race is not tested — neither are the `enabled` and `max_attempts` migrations it copies. Not raised as a defect here; see the proposed follow-up.
- **Fresh vs migrated column list** — round one measured this but nothing pinned it. It does now: the migration test asserts a migrated `_honker_live` and a fresh one have the same names in the same `cid` order, and that `claimed_at` is last. Mutation C shows it is load-bearing — the pre-existing `assert_eq!(live_cols.len(), 13)` + `contains(...)` check does **not** catch a reordering.

### `claimed_at` across every state transition (RUN, throwaway probe, reverted)

| transition | result |
| --- | --- |
| enqueue | `claimed_at` NULL |
| first claim | set to `unixepoch()` |
| heartbeat | unchanged |
| retry → pending | cleared |
| retry, attempts exhausted | row deleted → `_honker_dead` |
| reclaim after expiry | reset to the reclaim time |
| ack | row deleted |
| **cancel** | row deleted, `get_job` returns empty |
| **fail** | row deleted → `_honker_dead` |
| **dead-letter on reclaim past max_attempts** | row deleted → `_honker_dead` |
| **sweep_expired** | never sees a non-NULL `claimed_at` — it only touches `state='pending'`, and pending always has `claimed_at` NULL |
| **claimed, then queue drained / nobody reclaims** | row keeps the full stale claim — finding 1 |

Invariant that falls out and holds everywhere: **`claimed_at IS NOT NULL` iff `state='processing'`**, with one deliberate exception — a row that was already `processing` when an existing database migrated, which the migration leaves NULL rather than backfilling a claim that never happened. That exception is now in `get_job`'s doc comment, because someone computing runtime has to handle NULL on a `processing` row right after an upgrade.

`_honker_dead` carries no `claimed_at`, which is right: it carries no `worker_id` or `claim_expires_at` either.

### Blast radius, checked independently

- No `SELECT *` on `_honker_live` and no positional reads of it anywhere in the repo; every `RETURNING` names its columns (RUN, grep).
- No binding maps `claimed_at`; the only repo hits are unrelated local variables in Python tests that time a claim with `time.time()` (RUN, grep).
- No binding uses strict deserialization and no test asserts an exact key set or key count on `claim_batch` or `get_job` output (INFERRED — swept all of `packages/` and `tests/`). Specifically checked for the modes that would reject an extra key: serde `deny_unknown_fields` (zero hits repo-wide), Go `DisallowUnknownFields` (zero hits — plain `json.Unmarshal`), .NET `JsonSerializerOptions` with `UnmappedMemberHandling.Disallow` (no `JsonSerializerOptions` constructed anywhere in the package, so the default Skip applies), Jackson `FAIL_ON_UNKNOWN_PROPERTIES` (no Jackson — the JVM binding hand-rolls a parser into `Map<String,Object>` and reads keys by name), pydantic `extra="forbid"` and `Job(**row)` (Python's `Job.__init__` takes `row: dict` and indexes by key), and nlohmann strict parse in C++ (`j.value(key, default)`). The Bun binding's `RawJob`/`JobRow` are TypeScript interfaces, so `JSON.parse(...) as JobRow` is an unchecked compile-time cast with no runtime validation.

### Things I looked at and decided were NOT defects

- **`.unwrap_or(false)` on the migration's `pragma_table_info` read.** It swallows every error from the read, which is the `.ok()` shape that #138 just removed from five lookups. Not the same defect: there the fallback was a *business answer* ("not your claim", "no such job"). Here the fallback only makes the code attempt an idempotent `ALTER TABLE`, whose own errors are checked — column missing → ALTER succeeds; column present → `duplicate column`, swallowed, correct; read failed for a real reason like SQLITE_BUSY → the ALTER hits it too and is returned. Every path ends correct or loud. Leaving it, and it matches the two migrations above it.
- **`_honker_dead` not gaining `claimed_at`.** Consistent with it carrying no other claim state.

## Proposed follow-ups (not done here, not in scope)

1. **Expose `claimed_at` per binding.** #136's field list explicitly excludes it and points at #135; #135 says add it to #136's shape *after* the schema lands. Merging this closes #135 and nobody updates #136, so the loop does not close: "how long has this been running" stays SQL-only with no issue tracking the gap. #136's checklist wants one more line, or a new issue.
2. **Test the concurrent-bootstrap migration race.** Three `ALTER TABLE ADD COLUMN` migrations tolerate `duplicate column` and none has a two-process test. Belongs on its own, covering all three.

## CI

Green on `fd063c9`: 19 pass, 2 skipped, 0 fail.

The first run failed one job — `python · honker (windows-latest, 3.12)`, `test_queue_1writer_many_workers_no_double_claim[None]`, `combined [] != expected [0..59]`. Not this branch. My two commits change no production code at all: every non-test line is a `///` doc comment, and the `lib.rs` change is entirely inside `mod tests` (RUN, diffed `6780cae..fd063c9`). The test spawns three worker subprocesses that exit after `idle_exit_s = 2.0` seconds with no job and with `idle_poll_s=None`, so they depend purely on cross-process wake delivery; all three returned `RESULT []`, which is a 2-second wake budget missed on a loaded Windows runner. Re-ran the failed job, passed. Flagging it because that test's own comment says a broken backend surfaces as "processed 0 jobs" — which is exactly what a slow runner also looks like, so the failure mode is ambiguous by construction. Separate file, separate concern; #12 is the closest existing home.

## Verdict

The three findings are resolved. Round two found no correctness defect. My changes are docs and tests only — no behaviour change, no binding touched, `claimed_at` still deliberately unmapped.

Not merging; that is yours.

PR #141, opened 2026-09-01
feat(jvm): carry every job field and add job snapshots

russellromney:
Part of #136 — the JVM half of the binding checklist.

**Stacked on #124. Merge #124 first.** This branch is based on
`feat/node-typed-jobs`, not `main`: the widened `honker_claim_batch`
RETURNING clause that exposes the full job snapshot exists only there.
The base of this PR is `feat/node-typed-jobs` for that reason.

## What changed

`packages/honker-jvm` only. No core, no SQL ABI, no other binding.

- `Job` and `TypedJob<T>` now expose all twelve fields core already
  returns: `id`, `queue`, `payloadJson`, `state`, `priority`, `runAt`,
  `workerId`, `claimExpiresAt`, `attempts`, `maxAttempts`, `createdAt`,
  `expiresAt`. Core JSON stays snake_case; the accessors are camelCase.
  Times are unix epoch seconds. The three columns that can be NULL
  (`workerId`, `claimExpiresAt`, `expiresAt`) return null. A missing
  required field throws instead of defaulting.
- New `JobSnapshot` record — the same twelve fields, data only, no
  ack/retry/fail/heartbeat. `Job.snapshot()` hands one out.
- New `Queue.getJob(id)` returns `Optional<JobSnapshot>` scoped to that
  queue. Job ids are globally unique, so an unscoped hit could return a
  payload that does not match the queue's declared type. `Database.getJob(id)`
  is the global lookup. Same split Node landed in #124.
- New `TypedJobSnapshot<T>` and `TypedQueue.getJob(id)` — the snapshot
  with the payload already decoded by the queue's `JsonCodec<T>`.

`claimed_at` is deliberately not here; that is #135.

## Typing is compile-time only

Payload types are a contract between the callers that share a queue.
Honker does not validate payload shape in the database and this PR adds
no runtime validation. The README and CHANGELOG both say so.

## Tests

Five new tests in `HonkerJvmTest`, all asserting values rather than
presence:

- `claimedJobCarriesEveryFieldCoreReturns` — enqueues with an explicit
  `runAt`, `priority(7)`, `maxAttempts(5)` and `expires(600s)`, claims,
  and checks all twelve fields. The clock-derived three are bounded by
  `unixepoch()` readings taken around the write and the claim.
- `jobSnapshotsFollowPendingProcessingAndAckedStates` — a second
  `Database` on the same file reads the job pending (no worker, no claim
  deadline, 0 attempts), then processing (worker id, attempts 1, claim
  deadline in the visibility window), then empty after ack.
- `delayedJobSnapshotReportsItsRunAt`
- `typedQueueCarriesEveryFieldOnJobsAndSnapshots`
- `getJobIsScopedToItsQueueWhileDatabaseLookupIsGlobal`

Local run: `mvn -B test` → 50/50 pass (45 existing + 5 new), extension
built `--release`.

### The tests are load-bearing

Broke the `run_at` mapping in `JobSnapshot.from` to read `created_at` —
the realistic column mixup — and reran:

```
[ERROR] Tests run: 2, Failures: 2, Errors: 0
HonkerJvmTest.delayedJobSnapshotReportsItsRunAt:182->assertBetween:1325
  runAt should be within [1788254339, 1788254339] but was 1788254309
  ==> expected: <true> but was: <false>
HonkerJvmTest.claimedJobCarriesEveryFieldCoreReturns:101
  expected: <1788254305> but was: <1788254310>
```

Restored, reran: 5/5 pass.

## Not touched

`honker-core`, the SQL ABI, `packages/honker-node`, every other binding.
The Kotlin half follows in a separate PR stacked on this one.

---

## Snapshot payload encoding (cross-binding note, no change requested)

Flagging for a later deliberate decision across all bindings, per the
#136 coordination: **the JVM binding returns snapshot payloads as raw
JSON text.** `JobSnapshot.payloadJson()` is exactly the string the row
holds; decoding is opt-in and explicit via `payload(JsonCodec<T>)` or
`TypedQueue.getJob()` → `TypedJobSnapshot<T>.payload()`.

That matches what claimed JVM jobs already did (`Job.payloadJson()`), so
nothing changed encoding here. It also matches Go's `JobRow.Payload` and
Python's `get_job()["payload"]`, and differs from Node #124, which
returns a decoded payload. Three bindings, two conventions. Not resolved
in this PR — changing it would be a breaking API change #136 did not ask
for.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -730,7 +730,7 @@ jobs:
         # differs is Linux-specific and needs its own investigation —
         # tracked in ROADMAP under Phase Kernel-JVM.
         #
-        # The other 49 tests gate, and so do the ORM recipes below.
+        # The other 52 tests gate, and so do the ORM recipes below.
         run: mvn -B test -Dtest='!HonkerJvmTest#childJvmKernelListenerWakesWhenParentNotifies'
       - name: ORM · JVM (JDBC, Hibernate, jOOQ)
         working-directory: scripts/proof/orm/jvm
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -730,7 +730,7 @@ jobs:
         # differs is Linux-specific and needs its own investigation —
         # tracked in ROADMAP under Phase Kernel-JVM.
         #
-        # The other 44 tests gate, and so do the ORM recipes below.
+        # The other 49 tests gate, and so do the ORM recipes below.
         run: mvn -B test -Dtest='!HonkerJvmTest#childJvmKernelListenerWakesWhenParentNotifies'
       - name: ORM · JVM (JDBC, Hibernate, jOOQ)
         working-directory: scripts/proof/orm/jvm
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,25 @@
 - Regression test claims a job at a deadline with one claim forced to
   return empty: it fails on the unpatched `api.js` and passes on the fix.
 
+## Unreleased — JVM job parity
+
+- JVM claimed `Job` and `TypedJob<T>` now carry every field the core
+  returns: `id`, `queue`, `payloadJson`, `state`, `priority`, `runAt`,
+  `workerId`, `claimExpiresAt`, `attempts`, `maxAttempts`, `createdAt`,
+  and `expiresAt`. The core JSON stays snake_case; the accessors are
+  camelCase. Times are unix epoch seconds.
+- New `JobSnapshot` record: the same twelve fields with no claim
+  operations. `Queue.getJob(id)` returns one scoped to that queue,
+  `Database.getJob(id)` looks up by id across every queue. Both are empty
+  once the job is ack'd or dead-lettered. `TypedQueue.getJob(id)` returns
+  a `TypedJobSnapshot<T>` with the payload already decoded.
+- Payload types stay a compile-time contract. Honker does not validate
+  payload shape in the database, and no runtime validation was added.
+- Tests enqueue with an explicit `runAt`, priority, `maxAttempts` and
+  `expires`, then claim and assert each field's value — including a
+  reader on a second connection watching one job go pending, processing,
+  then gone after ack.
+
 ## Unreleased — typed Node jobs
 
 - Node `Queue`, `Job`, `JobSnapshot`, `ClaimWaker`, and `Outbox` APIs now carry
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,10 +34,29 @@
   a `TypedJobSnapshot<T>` with the payload already decoded.
 - Payload types stay a compile-time contract. Honker does not validate
   payload shape in the database, and no runtime validation was added.
+- `TypedJob.snapshot()` returns a `TypedJobSnapshot<T>`, so the typed
+  path stays typed end to end. `raw()` steps down a layer:
+  `TypedJob.raw()` is the `Job`, `TypedJobSnapshot.raw()` is the
+  `JobSnapshot`.
+- The three nullable accessors (`workerId`, `claimExpiresAt`,
+  `expiresAt`) carry `org.jspecify.annotations.@Nullable`. The dependency
+  was already declared and unused. Kotlin callers see `String?` / `Long?`
+  instead of a platform type.
+- Building a claimed `Job` from a row with no `worker_id` or
+  `claim_expires_at` throws instead of unboxing `null` at
+  `claimExpiresAt()`.
 - Tests enqueue with an explicit `runAt`, priority, `maxAttempts` and
   `expires`, then claim and assert each field's value — including a
   reader on a second connection watching one job go pending, processing,
-  then gone after ack.
+  then gone after ack, a retry clearing `workerId` and `claimExpiresAt`
+  back to null while `attempts` survives, a `fail()` dead-letter leaving
+  no snapshot, and a narrow row from an older extension throwing rather
+  than filling in zeros.
+- `packages/honker-jvm/pom.xml` packages the extension from
+  `target/release`, falling back to `target/debug`, and fails the build
+  when neither exists. It copied `target/debug` only, with
+  `failonerror="false"`, so a release build produced a green
+  `mvn install` and a jar whose `runtimes/` directory was empty.
 
 ## Unreleased — typed Node jobs
 
--- a/packages/honker-jvm/README.md
+++ b/packages/honker-jvm/README.md
@@ -92,12 +92,17 @@ A claimed `Job` carries every field the core returns for the row:
 `id`, `queue`, `payloadJson`, `state`, `priority`, `runAt`, `workerId`,
 `claimExpiresAt`, `attempts`, `maxAttempts`, `createdAt`, and `expiresAt`.
 The core JSON is snake_case; the accessors are camelCase. Times are unix
-epoch seconds. `workerId`, `claimExpiresAt` and `expiresAt` are `null`
-when the row has no value for them.
+epoch seconds. A claimed row always names its worker and carries a claim
+deadline, so `workerId()` and `claimExpiresAt()` are never `null` on a
+`Job`. Only `expiresAt()` is, and only when the job was enqueued without
+`expires`.
 
 `JobSnapshot` is the same data with no claim operations. Read one with
 `Queue.getJob(id)` (scoped to that queue) or `Database.getJob(id)`
-(global). It is empty once the job is ack'd or dead-lettered.
+(global). It is empty once the job is ack'd or dead-lettered. A snapshot
+can describe a pending row, so all three of `workerId()`,
+`claimExpiresAt()` and `expiresAt()` are `null` when the row has no value
+for them.
 
 ```java
 Queue emails = db.queue("emails");
@@ -117,7 +122,10 @@ emails.getJob(id);                        // Optional.empty()
 ```
 
 `TypedQueue.getJob(id)` returns a `TypedJobSnapshot<T>` with the payload
-already decoded.
+already decoded, and `TypedJob.snapshot()` hands out the same shape for a
+job you already claimed. `raw()` steps down a layer either way:
+`TypedJob.raw()` is the `Job`, `TypedJobSnapshot.raw()` is the
+`JobSnapshot`.
 
 Payload types are a compile-time contract between the callers that share
 a queue. Honker does not validate payload shape in the database, so every
--- a/packages/honker-jvm/README.md
+++ b/packages/honker-jvm/README.md
@@ -86,6 +86,43 @@ sendEmail(job.payload());
 job.ack();
 ```
 
+## Job fields and snapshots
+
+A claimed `Job` carries every field the core returns for the row:
+`id`, `queue`, `payloadJson`, `state`, `priority`, `runAt`, `workerId`,
+`claimExpiresAt`, `attempts`, `maxAttempts`, `createdAt`, and `expiresAt`.
+The core JSON is snake_case; the accessors are camelCase. Times are unix
+epoch seconds. `workerId`, `claimExpiresAt` and `expiresAt` are `null`
+when the row has no value for them.
+
+`JobSnapshot` is the same data with no claim operations. Read one with
+`Queue.getJob(id)` (scoped to that queue) or `Database.getJob(id)`
+(global). It is empty once the job is ack'd or dead-lettered.
+
+```java
+Queue emails = db.queue("emails");
+long id = emails.enqueue("{\"to\":\"alice@example.com\"}");
+
+JobSnapshot pending = emails.getJob(id).orElseThrow();
+System.out.println(pending.state());     // "pending"
+System.out.println(pending.workerId());  // null
+
+Job job = emails.claimOne("worker-1").orElseThrow();
+System.out.println(job.state());          // "processing"
+System.out.println(job.attempts() + "/" + job.maxAttempts());
+System.out.println(job.claimExpiresAt());
+job.ack();
+
+emails.getJob(id);                        // Optional.empty()
+```
+
+`TypedQueue.getJob(id)` returns a `TypedJobSnapshot<T>` with the payload
+already decoded.
+
+Payload types are a compile-time contract between the callers that share
+a queue. Honker does not validate payload shape in the database, so every
+writer must agree on the JSON it produces.
+
 Result waits can use `CompletableFuture`:
 
 ```java
--- a/packages/honker-jvm/pom.xml
+++ b/packages/honker-jvm/pom.xml
@@ -51,11 +51,17 @@
             <phase>process-resources</phase>
             <configuration>
               <!--
-                Packages the extension the jar ships to users. This used to
-                copy `target/debug` only, with failonerror="false", so a
-                release build produced a green `mvn install` and a jar whose
-                runtimes/ directory was empty. Prefer release, fall back to
-                debug, and fail the build when neither exists.
+                Packages the extension the jar ships to users, which is the
+                only way it is found once the jar leaves this repo (see
+                scripts/test-jvm-consumer.sh, which unsets
+                HONKER_EXTENSION_PATH). This used to copy `target/debug`
+                only, with failonerror="false", so a release cargo build
+                followed by `mvn install` gave a green build and a jar whose
+                runtimes/ directory was empty.
+
+                Take whichever of debug and release was built last, so a
+                stale artifact never beats the one you just compiled, and
+                fail the build when there is neither.
               -->
               <target>
                 <property name="honker.native.release"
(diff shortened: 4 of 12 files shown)


russellromney:
Adversarial review (I did not write this code). Everything below was RUN unless marked INFERRED, in a detached worktree at `5cdac85` against a prebuilt extension.

**No defects found.** This is the strongest test suite in the #136 series and the one the other bindings should copy.

Raw results:
```
full suite:                       Tests run: 50, Failures: 1, Errors: 0, Skipped: 0, 31.54 s
CI's exclusion + env-broken python test excluded:
                                  Tests run: 48, Failures: 0, Errors: 0, Skipped: 0, 25.59 s  BUILD SUCCESS
```
The single failure is environment-only: `pythonStdlibSqliteInteropBothDirections` falls back to the macOS system `python3`, which lacks `enable_load_extension`; CI uses the repo venv. It fails loudly rather than skipping, which is correct.

The CI comment's arithmetic checks out: `@Test` count went 45 → 50, one is excluded by name, so "the other 49 tests gate" is exact.

### The named repo trap is defeated

The tests do **not** assert "runAt is roughly now" — they enqueue with an explicit `runAt = unixNow(db) - 5` and assert `assertEquals(runAt, job.runAt())` while bounding `createdAt` to `[enqueuedBefore, enqueuedAfter]`. `claimExpiresAt` (vt 300 s / 120 s / 60 s) and `expiresAt` (600 s / 900 s) are given deliberately different offsets. Every confusable pair is separated by construction.

Six mutations applied to the real source, the five new tests re-run each time, then `git checkout --`:

| Mutation | Result |
|---|---|
| `attempts` ↔ `max_attempts` | **caught** — 3 failures |
| `runAt` reads `created_at` (one-way) | **caught** — 4 failures |
| `createdAt` reads `run_at` (one-way) | **caught** — 3 failures |
| `claim_expires_at` ↔ `expires_at` | **caught** — 3 failures |
| nulls collapsed to `""` / `0L` | **caught** — 2 failures |
| `Queue.getJob` filter deleted | **caught** — 1 failure (`getJobIsScopedToItsQueueWhileDatabaseLookupIsGlobal`) |

Baseline before each: `Tests run: 5, Failures: 0`. Both directions of the run_at/created_at swap fail — the shape that survives in #139, #142 and #150.

### Also verified

- **Field wiring.** All twelve keys in `JobSnapshot.from()` checked against `honker_ops.rs` `claim_batch` (~line 720) and `get_job` (1006-1075). All correct. `Json.Parser` maps JSON `null` → Java `null` and integers → `Long`, so `nullableString`/`nullableLong` see real nulls; missing-key and JSON-null are handled identically. `requiredLong`/`requiredString` throw `HonkerException` on a missing field rather than defaulting — fail-loudly, and unlike #151 it is a single exception type for both missing and malformed.
- **No silent skips.** `NativeLoader.resolve()` throws `HonkerLoadException` on every miss; a set-but-wrong `HONKER_EXTENSION_PATH` fails rather than falling through. Proven: forcing a nonexistent path gives `Tests run: 10, Failures: 0, Errors: 10, Skipped: 0`. Never a skip. The only `assumeTrue` in the file guards python availability.
- **Scoping is real and load-bearing.** `Queue.java:81` is `db.getJob(jobId).filter(job -> name.equals(job.queue()))`, `Database.getJob` is global, and deleting the filter fails a test. Matches the Node reference (`api.js:470-494`, `wrapper.d.ts:217-225`).

### Low-severity notes, none blocking

1. **`pom.xml` silently ships a jar with no native. PRE-EXISTING, not in this diff — but it is the mechanism these tests depend on, so it belongs in the record.** The antrun step copies from `${project.basedir}/../../target/debug/${honker.native.name}` — **`target/debug` only, never `target/release`** — with `failonerror="false"`. Observed live during install:
   ```
   [ERROR] [copy] Warning: Could not find file .../target/debug/libhonker_ext.dylib to copy.
   ... install exit=0
   ```
   `unzip -l` on the resulting jar shows `runtimes/osx-arm64/native/` present but empty. A user who follows the README (`cargo build --release`), runs `mvn install`, gets a green build and a jar with no packaged extension, then has to discover `HONKER_EXTENSION_PATH` on their own. Warn-and-continue on a load-bearing artifact. RAN.
2. **README overstates nullability for a claimed `Job`.** `README.md:91-96`, under "A claimed `Job` carries every field…", says "`workerId`, `claimExpiresAt` and `expiresAt` are `null` when the row has no value for them". But `Job.claimExpiresAt()` and `TypedJob.claimExpiresAt()` return primitive `long` — they cannot return null, and a claimed row always has both. The javadoc on the method is correct ("Never null on a claimed job"); the README paragraph is loose. It reads correctly for `JobSnapshot`, which is probably where the sentence came from. INFERRED.
3. **Latent unbox NPE, unreachable today.** `Job.claimExpiresAt()` unboxes `JobSnapshot`'s `Long`. If a `Job` were ever built from a row with `claim_expires_at IS NULL` that is an NPE rather than a clear error. The claim `UPDATE` always sets the column, so it cannot happen on the current path. INFERRED.
4. **Deliberate divergences from the Node reference, on the record:** Node has no `Database.getJob`; #141 adds one (and the scoping design now depends on it — .NET and C++ have no equivalent yet). Node's `Job.claimExpiresAt` is `number | null` where JVM's is primitive `long`. Node's `Job.payload` is decoded where JVM splits `payloadJson()` / `payload(codec)`. All look intentional; none is wrong.

Would merge.

Environment note: the machine hit 100% disk mid-review and Bash was dead for ~10 minutes. I used the shared `~/.m2` in offline mode rather than an isolated repo, so `mvn install` of `dev.honker:honker:0.5.0` wrote there twice (once mutated for the scoping proof, once restored); the last install was the clean one and `packages/honker-jvm/src` is byte-identical between my two worktrees. Local JDK was openjdk@25 vs CI's temurin 17; `maven.compiler.release=17` so bytecode target matched, the running JVM did not. Worktree left clean.

russellromney:
Second adversarial round. I did not write this code and I am not the reviewer who found it clean. Round one asked "can I break the tests". This round asked a different question: is it complete, well designed, free of bugs, comprehensively tested, and does it make sense to a Java user.

I found seven things worth acting on and pushed six commits: `b001b34`, `5153921`, `6e70efe`, `bd58246`, `96e68fd`, `550e75f`. Everything below is **RAN** unless marked READ or INFERRED.

---

## Fixed

### 1. The README's nullability sentence is wrong for a claimed `Job` (medium, docs)

Round one flagged this as INFERRED. Confirmed by reading the source: `Job.claimExpiresAt()` and `TypedJob.claimExpiresAt()` return primitive `long`, and `Job.workerId()` on a claimed row is always set. The sentence is true of a `JobSnapshot`, which is where it belongs. Both paragraphs now say what is actually true of each type. READ.

### 2. `TypedJobSnapshot<T>` was unreachable from a job you already claimed (medium, completeness)

`TypedJob.snapshot()` returned the undecoded `JobSnapshot`, and `TypedJobSnapshot.of` is package-private. So there was no public way to get a `TypedJobSnapshot<T>` out of a `TypedJob<T>` — only `TypedQueue.getJob()` could produce one, and only for a job you had not claimed. `Job.snapshot()` exists precisely to hand the data across a boundary without the claim operations; the typed path could not do that.

`TypedJob.snapshot()` now returns `TypedJobSnapshot<T>`. That also makes the two layers read the same way:

| | untyped | typed |
|---|---|---|
| `claimOne` | `Job` | `TypedJob<T>` |
| `getJob` | `JobSnapshot` | `TypedJobSnapshot<T>` |
| `snapshot()` | `JobSnapshot` | `TypedJobSnapshot<T>` |
| `raw()` | — | one layer down |

Nothing on this branch called the old signature — `.snapshot()` appears only in `Job.java` and two test assertions, and neither `packages/honker-kotlin/src` nor `scripts/proof/orm/jvm/src` mentions `TypedJob`, `JobSnapshot` or the changed accessors at all. It is unreleased API, so there is no compatibility question. **#144 should check it:** if the Kotlin half wraps `TypedJob.snapshot()`, it now gets `TypedJobSnapshot<T>` rather than `JobSnapshot`.

### 3. `jspecify` is declared in the pom and used nowhere (medium — and this one lands under #144)

`org.jspecify:jspecify:1.0.0` has been an `<optional>` dependency with zero references in `src/`. This PR adds three genuinely nullable accessors. `honker-kotlin` compiles against this jar (`packages/honker-kotlin/pom.xml` depends on `dev.honker:honker`), so without annotations a Kotlin caller sees `String!` / `Long!` and gets no compiler help on exactly the three fields that can be null.

`workerId`, `claimExpiresAt` and `expiresAt` now carry `org.jspecify.annotations.@Nullable` on `JobSnapshot`, `Job`, `TypedJob` and `TypedJobSnapshot`. I did **not** add `@NullMarked` to the package — that would restate the nullability of every type in `dev.honker` and change what #144 sees across the whole surface. That is a decision for its own change.

### 4. `Job.claimExpiresAt()` unboxes a `Long` the class never checked (low, latent)

Round one called this unreachable and was right — `Job.from` is only reached from `claimBatch` (READ: `Queue.java:66` is the only caller). But the class asserted an invariant it never verified, and the failure mode was a bare NPE. `Job.from` now rejects a row with no `worker_id` or no `claim_expires_at`, naming the fields.

### 5. The `pom.xml` packaging defect — fixed here, with a wrinkle round one did not have (medium)

Round one's finding reproduces: the antrun step copied `target/debug` only with `failonerror="false"`, so `cargo build --release` plus `mvn install` gave a green build and a jar with an empty `runtimes/` directory. I fixed it here rather than filing it separately: it is eight lines, it is warn-and-continue on a load-bearing artifact, and #144 stacks on this branch so it would sit unfixed under another PR. (I cannot open an issue on a public repo without approval, so "its own issue" was not available to me.)

**The wrinkle: "prefer release" is wrong.** `scripts/test-jvm-consumer.sh` is the only proof of the packaged native. It runs `cargo build -p honker-extension` (**debug**), installs, then runs `mvn test` from a `mktemp -d` with `HONKER_EXTENSION_PATH` unset, so the jar's copy is the only one `NativeLoader` can find. A release-first rule would silently package a stale `target/release` from another branch over the debug build the script just made — the same class of bug.

So the rule is **whichever of the two was built last**. Proven by putting distinguishable files in `target/debug` and `target/release` and reading back what landed in `target/classes`:

```
both exist, debug newer   -> picked: I-AM-DEBUG
both exist, release newer -> picked: I-AM-RELEASE
release only              -> BUILD SUCCESS, copied
debug only                -> BUILD SUCCESS, copied
neither                   -> BUILD FAILURE
```

and the failure is actionable:

```
Honker extension not built for osx-arm64. Looked for <repo>/target/release/libhonker_ext.dylib
and <repo>/target/debug/libhonker_ext.dylib. Run: cargo build --release -p honker-extension
```

The jar now actually carries it (it did not before, on this machine or in CI):

```
1064736  runtimes/osx-arm64/native/libhonker_ext.dylib
```

My first attempt at this had a literal `--` inside the XML comment, which makes the POM unparseable. Caught by running Maven; fixed in `96e68fd`. Worth noting as evidence that the build file needed execution to settle, not just reading.

### 6. Three behaviors had no test at all (medium)

Round one could not break the five tests. It is what they do not reach that mattered.

- **`retriedJobSnapshotGoesBackToPendingAndKeepsItsAttempts`.** Every existing snapshot test walks null → value. None walked the clearing direction. A retry puts the row back to `pending` with `workerId` and `claimExpiresAt` null *again*, `attempts` still 1 (not reset — that is what `maxAttempts` counts against), `createdAt` unchanged, and `runAt` pushed out by the delay. It is also the only test with a nonzero `attempts` on a pending row.
- **`failedJobIsDeadLetteredAndHasNoSnapshot`.** The README and CHANGELOG both claim a snapshot is empty once the job is "ack'd **or dead-lettered**". Only ack was proven. This claims `fail()`, checks both `Queue.getJob` and `Database.getJob` are empty, and counts `_honker_dead` so "empty" cannot be confused with "never existed".
- **`jobRowsMissingRequiredFieldsFailLoudly`.** The fail-loudly guards had no test. Drives `JobSnapshot.from` and `Job.from` with a row map: a full row decodes, a row missing `state` (what an extension older than the widened `honker_claim_batch` RETURNING sends) throws naming the field, a non-numeric `run_at` throws naming the field, and a row with no claim is a valid snapshot but not a valid `Job`.

Each was checked against a mutation of the real source, restored from a file copy rather than `git checkout --`:

| Mutation | Test | Result |
|---|---|---|
| `JobSnapshot.required()` returns `0L` instead of throwing | `jobRowsMissing…` | **caught** — "Expected HonkerException to be thrown, but nothing was thrown" |
| `Job.from` claim guard deleted | `jobRowsMissing…` | **caught** — same |
| `Job.fail(error)` calls `queueRef.ack(...)` | `failedJobIsDeadLettered…` | **caught** — `_honker_dead` count `expected: <1> but was: <0>` |
| `Job.retry(delay, …)` drops the delay | `retriedJobSnapshot…` | **caught** — `runAt should be within [1788340136, 1788340136] but was 1788340091` |

Baseline before each: 8/8 pass. All eight job tests pass restored.

### 7. A narrow row from an old extension now says why

`JobSnapshot.from` hard-requires `state`, `priority`, `run_at`, `max_attempts` and `created_at`. Against an extension older than #124 every `claimOne` throws. Throwing is right, but `missing required field state` leaves someone with a vendored old dylib nowhere to go. The message now names the likely cause.

---

## Left, and why

**JVM has no `cancel` at all.** This branch is the de-facto #134 reference for `getJob`, but it demonstrates only half of the decision — the other ten bindings will find no `cancel` precedent here. #134's own research says JVM/Kotlin should implement cancel directly at the 2-arg form once core has `honker_cancel(queue, job_id)`. That is new API and new core, both outside #136. Not mine to add. READ.

**No connect-time capability probe.** There is no `honker_version()` and no version guard in any binding (grepped `Honker.java`, `NativeLoader.java`, `Database.java`, `honker-extension/src`). #134's decision already asks for the `pragma_function_list` probe. Adding it to one binding only would be worse than adding it to none. Flagged; item 7 is the cheap mitigation.

**`getJob` has no `Transaction` overload.** `Queue.enqueue` and `Queue.saveResult` both take one; `getJob` does not, so you cannot read your own uncommitted write. Real gap, but it is new API surface #136 did not ask for, and `Database.getJob` wrapping `db.transaction(...)` is the same shape `nextClaimAt` and `getResult` already use. READ.

**`Math.toIntExact` throws `ArithmeticException`, not `HonkerException`,** for `priority`, `attempts` and `max_attempts` — inconsistent with the `HonkerException` in the same method. Unreachable: `EnqueueOptions.priority` is an `int` and the other two are bounded by it. Not worth a change. READ.

**`TypedJobSnapshot`'s canonical constructor is public** (records have no choice), so `new TypedJobSnapshot<>(raw, payloadThatDoesNotMatch)` compiles. Hiding it means giving up the record. Not worth it. READ.

**`TypedJob.payload()` re-decodes on every call,** and `snapshot()` now decodes too. Pre-existing shape; a caching change is a design decision, not a review fix. READ.

**The packaged-native path is not in CI.** The jvm job always sets `HONKER_EXTENSION_PATH`, so it never exercises the jar's copy — only `make test-jvm-consumer` does, and that is not wired into any workflow. Item 5 means CI's installed jar now carries a native for the first time, but nothing checks it. Worth its own change. READ.

**Payload encoding (#146) and `claimed_at` (#135)** — out of scope by instruction, unchanged.

---

## Scoping conformance with #134

Matches the decision. `Queue.getJob` is `db.getJob(jobId).filter(job -> name.equals(job.queue()))`; `Database.getJob` is global. #134 requires atomicity only for `cancel` ("a read-then-cancel in a binding races a concurrent claim"), and its research says explicitly that a read is not racy that way, so a post-read filter is the right shape for `getJob`.

It matches Node too — `api.js:475-494` does the same `if (row.queue !== this.name) return null` — and it goes one step further: #134's decision says "`Database.getJob(id)` and `Database.cancel(id)` are the global forms", and JVM is the only binding that has `Database.getJob`. Node does not. On this axis JVM is the reference and Node is behind. READ.

---

## Full run

```
mvn -B -o test -Dtest='!HonkerJvmTest#childJvmKernelListenerWakesWhenParentNotifies'
Tests run: 52, Failures: 1, Errors: 0, Skipped: 0
```

52 = 53 `@Test` minus the one CI excludes by name; the workflow comment is updated from 49 to 52. The single failure is the same environment-only one round one hit: `pythonStdlibSqliteInteropBothDirections` falls back to the macOS system `python3`, whose `sqlite3.Connection` has no `enable_load_extension`. CI uses the repo venv.

CI for these six commits is run [33612943924](https://github.com/russellromney/honker/actions/runs/33612943924); the runner queue is backed up and the `jvm` job had not started when I wrote this, so the numbers above are local. The run before my commits was fully green.

Environment: JDK 26 (Homebrew) with `maven.compiler.release=17`, offline `~/.m2`, extension built in-worktree at `--release` with `kernel-watcher,shm-fast-path`. I checked first and did **not** use the prebuilt `target/release/libhonker_ext.dylib` in the main checkout: `strings` shows it still carries `RETURNING id, queue, payload, worker_id, attempts, claim_expires_at`, the pre-#124 six-column form. Testing against it would have been meaningless. Worktree is clean; nothing outside `packages/honker-jvm`, `CHANGELOG.md` and one `ci.yml` comment line was touched.

---

## Verdict

**Merge**, after #124. The five original tests hold up and the fixes above close the gaps I could find. The two things I would keep on the list are both cross-binding and neither belongs here: JVM still has no `cancel`, so it is only half a #134 reference, and no binding has the capability probe #134's decision asks for.

PR #142, opened 2026-09-01
feat(dotnet): full job details and typed queues (#136)

russellromney:
Part of #136 — the .NET half. The Bun half is a separate PR.

**Stacked on #124 (`feat/node-typed-jobs`), which is the base of this PR. Merge after #124.**
It depends on the widened `honker_claim_batch` RETURNING clause that only exists there. No honker-core or SQL ABI change here; core already returns everything needed.

## What changed

- `Job` (claimed) now carries all twelve fields the core returns: `Id`, `QueueName`, payload, `State`, `Priority`, `RunAt`, `WorkerId`, `ClaimExpiresAt`, `Attempts`, `MaxAttempts`, `CreatedAt`, `ExpiresAt`. It keeps `Ack`/`Retry`/`Fail`/`Heartbeat`.
- `ClaimBatch` now decodes `honker_claim_batch` into the same `JobRow` record `GetJob()` returns (the two SQL functions return the same field set). The old `row.State ?? "processing"` fallback — against a field core did not previously return — is gone; `State` is a real core value now.
- A claimed row arriving without `worker_id` or `claim_expires_at` throws instead of being silently defaulted.
- New `db.Queue<TPayload>(name)` returns `TypedQueue<TPayload>`: `Enqueue(TPayload)`, `ClaimBatch`/`ClaimOne`/`ClaimAsync` yielding `Job<TPayload>`, and `GetJob()` yielding the data-only `JobSnapshot<TPayload>` record (no ack/retry/fail/heartbeat — reading a row does not claim it). `Untyped` gets you back to the plain handle.
- Named `TypedQueue<T>` rather than `Queue<T>` on purpose: a `Honker.Queue<T>` would collide with `System.Collections.Generic.Queue<T>` in any file that imports both (it broke this binding's own `Listener.cs`/`Stream.cs` first).
- Typing is **compile-time only**. Honker never validates payload shape in the database and this binding adds no runtime check; the README and XML docs say so.

Read-only snapshots needed no field work — `JobRow` already had all twelve.

### Snapshot payload encoding (flagged for a repo-wide decision)

No existing .NET payload encoding changed here. `Queue.GetJob()` still returns `JobRow` with `Payload` as raw JSON text, and `Job.Payload` is still a `JsonElement` with `GetPayload<T>()` for decoding. The new opt-in `JobSnapshot<TPayload>` (reachable only through `db.Queue<TPayload>()`) carries both: `Payload` decoded as `TPayload` and `PayloadRaw` as the original text.

Bindings currently disagree on snapshot payload encoding — Node (#124) decodes; Bun (#145), Go, and Python return raw text — and this PR does not try to settle it.

## Tests

New `packages/honker-dotnet/tests/Honker.Tests/JobDetailTests.cs`, 7 tests, each asserting values rather than presence:

- `ClaimedJobCarriesEveryCoreDetail` — explicit `runAt`/`priority`/`maxAttempts`/`expires`, then every field checked against the value that was enqueued (times bounded by `unixepoch()` reads taken around the call).
- `PendingSnapshotCarriesEveryCoreDetail` — pending snapshot: `state=pending`, null worker/claim expiry, exact payload JSON.
- `ReaderSeesProcessingDetailWhileWorkerHoldsTheClaimAndNothingAfterAck` — second connection reads the processing snapshot, then gets null after ack.
- `DelayedJobReportsItsRunAt`
- `TypedQueueDecodesClaimedPayloadsAndKeepsEveryDetail`
- `TypedSnapshotDecodesPayloadAndCarriesEveryDetail`
- `TypedAndUntypedHandlesShareOneQueue`

Local: `dotnet test ... -c Release` → 110 passed, 0 failed (whole suite, after building `Honker.QueueHelper`).

### Falsification

Broke `RunAt = row.RunAt` to `RunAt = row.CreatedAt` in the `Job` constructor:

```
Failed Honker.Tests.JobDetailTests.ClaimedJobCarriesEveryCoreDetail [9 ms]
  Assert.Equal() Failure: Values differ
  Expected: 1788254421
  Actual:   1788254426
  at Honker.Tests.JobDetailTests.ClaimedJobCarriesEveryCoreDetail() in JobDetailTests.cs:line 38
```

Restored, suite green again.

## Not in scope

- `claimed_at` (#135) — deliberately not added.
- Queue-scoping of `GetJob` (#134) — .NET `GetJob` is still global, unlike Node's now-scoped one.
- No changes to honker-core, the SQL ABI, honker-node, or any other binding.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,7 +73,21 @@
   fallback; loud is better than either. `worker_id` and
   `claim_expires_at` are in the old ABI, so their new throws did not
   catch it. Nothing in the binding constructs a `JobRow` — it is only a
-  deserialization target — so `required` costs callers nothing.
+  deserialization target.
+- `JobRow` is public, so `required` is a source-breaking change for a
+  downstream caller who constructs one: `new JobRow { Id = 1 }` compiled
+  before and is now `CS9035: Required member 'JobRow.Payload' must be
+  set`. Nothing outside a decoder has a reason to build a `JobRow`, and
+  the package is pre-1.0, so this is the accepted cost of the loud
+  decode.
+- Against a 0.5.x extension, `required` moves the failure from a wrong
+  `Job.State` to a `JsonException` thrown inside `ClaimBatch` — after
+  the rows are claimed in the database, which is the same stranding
+  shape the `Job<TPayload>.Payload` fix above removes. It is still the
+  right trade: a version-mismatched core fails on the very first claim
+  and every claim after, so the worker never appears to run, whereas one
+  producer's mismatched payload is an ordinary production event a
+  healthy worker has to survive.
 - The README now names the `TypedQueue<T>` vs `Queue<T>` collision, shows
   the poison-payload worker loop, separates the four spellings of
   "payload" across the typed and untyped pairs, and says `GetJob` is not
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,17 @@
 - Tests for the previously uncovered typed surface: `ClaimBatch`,
   `AckBatch`, `ClaimAsync`, `Cancel`, `Name`, `Job<T>.Untyped`,
   `Job<T>.Retry`, and `Job<T>.Heartbeat`.
+- Every `JobRow` member is now `required`, so a core that omits a field
+  fails the decode with a `JsonException` naming it. `ClaimBatch` decodes
+  into `JobRow` now, and `JobRow.State` defaulted to `""` — so this
+  binding run against a 0.5.x extension, whose `honker_claim_batch`
+  returns six columns and no `state`, would have reported
+  `Job.State == ""`. The `row.State ?? "processing"` fallback this PR
+  removed was right for that case. Wrong and silent is worse than the
+  fallback; loud is better than either. `worker_id` and
+  `claim_expires_at` are in the old ABI, so their new throws did not
+  catch it. Nothing in the binding constructs a `JobRow` — it is only a
+  deserialization target — so `required` costs callers nothing.
 - The README now names the `TypedQueue<T>` vs `Queue<T>` collision, shows
   the poison-payload worker loop, separates the four spellings of
   "payload" across the typed and untyped pairs, and says `GetJob` is not
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,25 @@
 - Regression test claims a job at a deadline with one claim forced to
   return empty: it fails on the unpatched `api.js` and passes on the fix.
 
+## Unreleased — .NET job details and typed queues
+
+- .NET claimed jobs (`Job`) now carry every field the core returns:
+  `State`, `Priority`, `RunAt`, `MaxAttempts`, `CreatedAt`, and
+  `ExpiresAt` join the id, queue, payload, worker, attempts, and
+  claim-expiry fields they already had. `ClaimBatch` decodes the claim
+  ABI into the same `JobRow` shape `GetJob()` returns, so the old
+  `row.State ?? "processing"` fallback is gone — `state` is now a real
+  value from the core.
+- New `db.Queue<TPayload>(name)` returns a `TypedQueue<TPayload>` whose
+  claims yield `Job<TPayload>` and whose `GetJob()` yields the
+  data-only `JobSnapshot<TPayload>` record. Named `TypedQueue` rather
+  than `Queue<T>` so it cannot collide with
+  `System.Collections.Generic.Queue<T>` in callers' files.
+- Payload typing is compile-time only. Honker never validates payload
+  shape in the database, and this binding adds no runtime check.
+- A claimed row missing `worker_id` or `claim_expires_at` now throws
+  instead of being silently defaulted.
+
 ## Unreleased — typed Node jobs
 
 - Node `Queue`, `Job`, `JobSnapshot`, `ClaimWaker`, and `Outbox` APIs now carry
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,14 @@
   shape in the database, and this binding adds no runtime check.
 - A claimed row missing `worker_id` or `claim_expires_at` now throws
   instead of being silently defaulted.
+- The typed job tests enqueue with a back-dated absolute `RunAtUnix`
+  (`now - 100`) so the job stays claimable while `RunAt` and `CreatedAt`
+  differ, then pin `RunAt` exactly and assert the two are not equal.
+  With an immediate enqueue the two are equal to the second, so all four
+  properties (`Job<T>.RunAt`/`CreatedAt` and `JobSnapshot<T>.FromRow`'s
+  two assignments) could be transposed and the suite stayed green.
+- `Job<TPayload>.PayloadRaw` is now asserted against the enqueued JSON.
+  Only `JobSnapshot<T>.PayloadRaw` was covered before.
 
 ## Unreleased — typed Node jobs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,27 @@
   two assignments) could be transposed and the suite stayed green.
 - `Job<TPayload>.PayloadRaw` is now asserted against the enqueued JSON.
   Only `JobSnapshot<T>.PayloadRaw` was covered before.
+- `Job<TPayload>.Payload` decodes on first read instead of inside the
+  constructor. It used to decode during the claim, so a payload that did
+  not match `TPayload` threw from inside `ClaimBatch`/`ClaimOne`/
+  `ClaimAsync` — after the rows were already claimed in the database.
+  The caller got an exception and no handle, so it could not `Ack`,
+  `Retry`, or `Fail` any job in that batch; they stayed invisible until
+  the visibility timeout and then poisoned the next claim. Honker never
+  validates payload shape, so one disagreeing producer was enough. A
+  claim now never decodes and never throws; the `JsonException` arrives
+  when you read `Payload`, with the job in hand to `Fail`.
+- `TypedQueue<T>.GetJob` still decodes eagerly and still throws on a
+  mismatched payload — a read holds no claim, so nothing is stranded.
+  `Untyped.GetJob` reads the row whatever shape it is. Both behaviours
+  are documented on the members and in the README.
+- Tests for the previously uncovered typed surface: `ClaimBatch`,
+  `AckBatch`, `ClaimAsync`, `Cancel`, `Name`, `Job<T>.Untyped`,
+  `Job<T>.Retry`, and `Job<T>.Heartbeat`.
+- The README now names the `TypedQueue<T>` vs `Queue<T>` collision, shows
+  the poison-payload worker loop, separates the four spellings of
+  "payload" across the typed and untyped pairs, and says `GetJob` is not
+  queue-scoped. `Queue.GetJob` says so in its XML doc too (#134).
 
 ## Unreleased — typed Node jobs
 
--- a/packages/honker-dotnet/README.md
+++ b/packages/honker-dotnet/README.md
@@ -64,10 +64,42 @@ if (job is not null)
 }
 ```
 
+The handle `db.Queue<TPayload>(name)` returns is called
+`TypedQueue<TPayload>`, not `Queue<TPayload>`. A `Honker.Queue<T>` would
+shadow `System.Collections.Generic.Queue<T>` in any file that uses both,
+so `var` or `TypedQueue<EmailPayload>` is the name to write down.
+
 The type parameter is a compile-time contract only. Honker stores
 payloads as opaque JSON and never validates their shape — in this
 binding or in the database — so every process writing to a queue has to
-agree on the payload type.
+agree on the payload type. When a producer disagrees, the row still
+lands in the queue and the decode is what fails:
+
+```csharp
+foreach (var job in emails.ClaimBatch("worker-1", 10))
+{
+    EmailPayload payload;
+    try
+    {
+        // Claiming never decodes, so the claim itself cannot throw and
+        // you always hold a handle. Payload decodes on first read.
+        payload = job.Payload!;
+    }
+    catch (JsonException e)
+    {
+        job.Fail(e.Message);   // dead-letter it; job.PayloadRaw has the text
+        continue;
+    }
+
+    Send(payload.To);
+    job.Ack();
+}
+```
+
+`GetJob()` is the one exception: it decodes eagerly and throws on a
+payload that does not match, because a read holds no claim and so
+strands nothing. `emails.Untyped.GetJob(id)` reads the row whatever
+shape it is.
 
 Claimed jobs (`Job`, `Job<TPayload>`) and read-only snapshots
 (`JobRow`, `JobSnapshot<TPayload>`) carry every field the core returns:
@@ -76,7 +108,19 @@ Claimed jobs (`Job`, `Job<TPayload>`) and read-only snapshots
 A claimed job additionally has `Ack`, `Retry`, `Fail`, and `Heartbeat`;
 a snapshot is data only, because reading a row does not claim it.
 `Untyped` on `TypedQueue<TPayload>` and `Job<TPayload>` gets you back
-to the untyped handle when an API needs one.
+to the untyped handle when an API needs one — results (`SaveResult`,
+`GetResult`, `WaitResult`) live there, since a result's type has
+nothing to do with `TPayload`.
+
+Two spellings of "payload" sit side by side, so watch which type you
+hold. On the typed pair (`Job<T>`, `JobSnapshot<T>`) `Payload` is the
+decoded `T` and `PayloadRaw` is the JSON text. On the untyped pair,
+`Job.Payload` is a `JsonElement` (`Job.GetPayload<T>()` decodes) and
+`JobRow.Payload` is the JSON text with no `PayloadRaw` beside it.
+
+`GetJob` is not scoped to the queue you called it on: job ids are
+globally unique and the lookup spans every queue, so pass only ids you
+know this queue owns.
 
 ## Native loading
 
--- a/packages/honker-dotnet/README.md
+++ b/packages/honker-dotnet/README.md
@@ -66,8 +66,8 @@ if (job is not null)
 
 The handle `db.Queue<TPayload>(name)` returns is called
 `TypedQueue<TPayload>`, not `Queue<TPayload>`. A `Honker.Queue<T>` would
-shadow `System.Collections.Generic.Queue<T>` in any file that uses both,
-so `var` or `TypedQueue<EmailPayload>` is the name to write down.
+shadow `System.Collections.Generic.Queue<T>` in any file that uses both.
+Write `var`, or `TypedQueue<EmailPayload>` where you need the name.
 
 The type parameter is a compile-time contract only. Honker stores
 payloads as opaque JSON and never validates their shape — in this
@@ -112,11 +112,11 @@ to the untyped handle when an API needs one — results (`SaveResult`,
 `GetResult`, `WaitResult`) live there, since a result's type has
 nothing to do with `TPayload`.
 
-Two spellings of "payload" sit side by side, so watch which type you
-hold. On the typed pair (`Job<T>`, `JobSnapshot<T>`) `Payload` is the
-decoded `T` and `PayloadRaw` is the JSON text. On the untyped pair,
-`Job.Payload` is a `JsonElement` (`Job.GetPayload<T>()` decodes) and
-`JobRow.Payload` is the JSON text with no `PayloadRaw` beside it.
+`Payload` means something different on each of these types, so check
+which one you hold. On the typed pair (`Job<T>`, `JobSnapshot<T>`) it is
+the decoded `T`, with `PayloadRaw` beside it for the JSON text. On `Job`
+it is a `JsonElement`, and `Job.GetPayload<T>()` decodes. On `JobRow` it
+is the JSON text itself, with no `PayloadRaw` beside it.
 
 `GetJob` is not scoped to the queue you called it on: job ids are
 globally unique and the lookup spans every queue, so pass only ids you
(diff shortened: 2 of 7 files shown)


russellromney:
Adversarial review (I did not write this code). Everything below was RUN unless marked INFERRED, in a detached worktree at `cee2a17` with `HONKER_EXTENSION_PATH` pointed at a prebuilt extension.

Baseline: `TestHarness.FindExtension` **throws** when the extension is absent — there is no skip path, so green means green.
`Passed! - Failed: 0, Passed: 7, Skipped: 0, Total: 7` for the JobDetail filter; `Failed: 3, Passed: 106, Skipped: 0, Total: 109` for the rest of the suite, where the 3 are pre-existing `BindingTests.QueueWatcherBackend*NoFallback(backend: null)` environmental failures that reproduce identically on a clean tree. CI runs `Honker.Tests.csproj` unfiltered (`ci.yml:366`), so the new tests are in the gate.

Positive control, to prove the suite is load-bearing at all: swapping `run_at`/`created_at` in the **untyped** `Job` constructor (Models.cs:170,180) fails `ClaimedJobCarriesEveryCoreDetail`. That path is well covered.

### 1. HIGH — `Job<TPayload>` can have `RunAt` and `CreatedAt` swapped and every test still passes. RAN.

`TypedQueueDecodesClaimedPayloadsAndKeepsEveryDetail` enqueues with **no** `RunAtUnix` and no `DelaySeconds`, so `run_at == created_at` to the second, then asserts the identical range for both (JobDetailTests.cs:147,152):

```csharp
Assert.InRange(job.RunAt,     enqueuedBefore, enqueuedAfter);
Assert.InRange(job.CreatedAt, enqueuedBefore, enqueuedAfter);
```

Mutation in `src/Honker/Models.cs`:
```
-    public long RunAt => _job.RunAt;         // line 252
+    public long RunAt => _job.CreatedAt;
-    public long CreatedAt => _job.CreatedAt; // line 257
+    public long CreatedAt => _job.RunAt;
```
```
Passed!  - Failed: 0, Passed: 7, Skipped: 0, Total: 7, Duration: 411 ms
```

Failure a user hits: a typed worker on a delayed or scheduled queue reads `job.RunAt` to log lateness or drive a retry deadline and gets the enqueue time instead; `job.CreatedAt` reports the future scheduled time, so queue-age metrics go negative. Silent, no exception.

The untyped tests already do this right (`runAt = Now(db) - 5`, `Assert.Equal(runAt, job.RunAt)` at line 38). The typed fixtures just did not copy the pattern.

### 2. HIGH — `JobSnapshot<TPayload>.FromRow` has the same uncaught swap. RAN.

Same shape in `TypedSnapshotDecodesPayloadAndCarriesEveryDetail` (lines 177, 182): enqueued with no delay, both fields asserted against the same `InRange`.

```
-        RunAt = row.RunAt,          // line 301
+        RunAt = row.CreatedAt,
-        CreatedAt = row.CreatedAt,  // line 306
+        CreatedAt = row.RunAt,
```
```
Passed!  - Failed: 0, Passed: 7, Skipped: 0, Total: 7, Duration: 301 ms
```
With that mutation still applied, the **full 109-test suite** produced exactly the 3 pre-existing failures and nothing more — no other test in the package covers it either.

Failure a user hits: an admin view over `JobSnapshot<T>` shows "scheduled for" and "created at" transposed for every delayed job.

Fix for both: give the typed fixtures a distinct `RunAtUnix` (or `DelaySeconds`), same as the untyped ones.

### 3. MEDIUM — `Job<TPayload>.PayloadRaw` has zero coverage. RAN.

Only `JobSnapshot<T>.PayloadRaw` is asserted (line 174). Mutation at Models.cs:248:
```
-    public string PayloadRaw => _job.PayloadRaw;
+    public string PayloadRaw => _job.QueueName;
```
```
Passed!  - Failed: 0, Passed: 7, Skipped: 0, Total: 7, Duration: 408 ms
```
This is the escape hatch a caller reaches for when the typed decode is not what they want (re-serialization, forwarding, debugging). Wiring it to the queue name would ship undetected.

### 4. LOW — new typed surface with no test at all. INFERRED (grep over JobDetailTests.cs).

`TypedQueue<T>.ClaimBatch`, `.ClaimAsync`, `.AckBatch`, `.Cancel`, `.NextClaimAt`, and `Job<T>.Retry` / `.Fail` / `.Heartbeat` / `.Untyped` are added or newly reachable here and none are exercised. Only `ClaimOne`, `Enqueue`, `GetJob`, `Ack`, and `TypedQueue.Untyped` are.

### 5. LOW — `GetJob` doc enumeration becomes incomplete under #134.

Both `Queue.GetJob` and `TypedQueue<T>.GetJob` say a miss means "ack'd, dead'd, or never existed". Once scoping lands that needs "or belongs to a different queue" — compare `packages/honker-node/wrapper.d.ts:219-224`.

Separately, worth flagging to whoever picks up #134: **.NET has no `Database.GetJob` at all.** The only definitions in the whole package are `Queue.GetJob` and `TypedQueue.GetJob`. The reference design keeps a global `Database.getJob` as the escape hatch; .NET will have to add one alongside the scoping or lose global lookup entirely. Not a defect of this PR.

Today, `TypedQueue<OrderPayload>.GetJob(idOwnedByAnotherQueue)` returns a `JobSnapshot<OrderPayload>` built from a foreign queue's JSON. `System.Text.Json` ignores unknown properties, so it most likely produces a record with default/null members rather than throwing — exactly the unsoundness wrapper.d.ts cites as the reason for scoping. INFERRED, not run.

- **Field wiring is correct.** All twelve `JsonPropertyName` attributes on `JobRow` (Models.cs:97-129) match the core's emitted keys in `honker_ops.rs` `claim_batch` and `get_job` exactly. `System.Text.Json` is case-sensitive by default and every name is explicit — no accidental fuzzy match.
- **Nulls are right and genuinely guarded.** `WorkerId` is `string?`, `ClaimExpiresAt`/`ExpiresAt` are `long?`; no non-nullable value type sits under a nullable column. `PendingSnapshotCarriesEveryCoreDetail` asserts all three are `null` (not `0`, not `""`), and that assertion is load-bearing — a `claim_expires_at`/`expires_at` swap makes it non-null and fails.
- **The confusable pairs are separated everywhere except finding 1/2.** `attempts` vs `max_attempts`: 1/5, 0/2, 1/4, 0/7 across the tests. `claim_expires_at` vs `expires_at`: now+120 vs now+600, now+90 vs now+300. Those swaps all fail.
- **The 27 lines deleted from Queue.cs are a net improvement, not a regression.** The private `ClaimedJobRow` was replaced by the shared `JobRow`. Three behavior changes, all in the fail-loudly direction: the `row.State ?? "processing"` fallback is gone (the core always emits `state`); `WorkerId` went from non-nullable defaulting to `""` to nullable with `throw new InvalidOperationException($"claimed job {row.Id} has no worker_id")`; `ClaimExpiresAt` went from defaulting to `0` to the same throw. No null check, error path, or throw was removed.
- **No conflict with #134.** `TypedQueue<T>.GetJob` is a thin delegate to `Queue.GetJob`; scoping touches one method body. No test asserts global lookup — every `GetJob` call site in `JobDetailTests.cs` and `PhaseMantleTests.cs` uses an id from the same queue handle, so nothing breaks when scoping lands.

Would not merge until findings 1 and 2 are fixed — they are a one-line fixture change each, and until then the typed job's `RunAt` is advertised but not proven.

Worktree left clean; all mutations reverted.

russellromney:
Follow-up: I independently reproduced findings 1 and 2 on a clean tree, because the first pass left an unreverted edit in `JobDetailTests.cs` that could have contaminated the result. It did not.

Reverted the test file to `cee2a17`, applied only the source mutation to `packages/honker-dotnet/src/Honker/Models.cs` (all four lines — both the `Job<TPayload>` properties and `JobSnapshot<TPayload>.FromRow`):

```
-    public long RunAt => _job.RunAt;            -        RunAt = row.RunAt,
+    public long RunAt => _job.CreatedAt;        +        RunAt = row.CreatedAt,
-    public long CreatedAt => _job.CreatedAt;    -        CreatedAt = row.CreatedAt,
+    public long CreatedAt => _job.RunAt;        +        CreatedAt = row.RunAt,
```
```
$ git diff --stat
 packages/honker-dotnet/src/Honker/Models.cs | 8 ++++----
$ dotnet test ... --filter FullyQualifiedName~JobDetail
Passed!  - Failed: 0, Passed: 7, Skipped: 0, Total: 7, Duration: 193 ms
```

Both fields transposed on both typed surfaces, all 7 tests green. Mutation reverted; tree clean.

russellromney:
Findings 1, 2, and 3 fixed. CI green on `1c57e43` (run 33585805328).

`test(dotnet): pin the typed job's RunAt apart from its CreatedAt` — commit `1c57e43`.

Both typed tests now enqueue with `RunAtUnix: Now(db) - 100`, the same pattern the untyped tests already use. The job stays claimable — a delayed job is not — while `RunAt` and `CreatedAt` differ by 100 seconds. Each test pins `RunAt` to that exact timestamp, keeps `CreatedAt` in the enqueue window, and adds `Assert.NotEqual(RunAt, CreatedAt)`. `TypedQueueDecodesClaimedPayloadsAndKeepsEveryDetail` also asserts `job.PayloadRaw`, finding 3.

### Findings 1 + 2 — proof

Your exact mutation, all four properties in `src/Honker/Models.cs` (`Job<TPayload>.RunAt`/`CreatedAt` and both assignments in `JobSnapshot<TPayload>.FromRow`):

```
  Failed Honker.Tests.JobDetailTests.TypedQueueDecodesClaimedPayloadsAndKeepsEveryDetail
   Assert.Equal() Failure: Values differ
   Expected: 1788318267
   Actual:   1788318367
     at JobDetailTests.cs:line 155
  Failed Honker.Tests.JobDetailTests.TypedSnapshotDecodesPayloadAndCarriesEveryDetail
   Assert.Equal() Failure: Values differ
   Expected: 1788318267
   Actual:   1788318367
     at JobDetailTests.cs:line 192
Failed!  - Failed: 2, Passed: 5, Skipped: 0, Total: 7
```

Line 155 is `Assert.Equal(runAt, job.RunAt)`; line 192 is `Assert.Equal(runAt, pending.RunAt)`. Both surfaces caught.

### Finding 3 — proof

`Job<TPayload>.PayloadRaw` wired to `QueueName` (Models.cs:248), everything else clean:

```
  Failed Honker.Tests.JobDetailTests.TypedQueueDecodesClaimedPayloadsAndKeepsEveryDetail
   Assert.Equal() Failure: Strings differ
           ↓ (pos 0)
   Expected: "{"Sku":"SKU-42","Quantity":3}"
   Actual:   "orders"
           ↑ (pos 0)
     at JobDetailTests.cs:line 152
```

### Runs

Clean tree, JobDetail filter: `Passed! - Failed: 0, Passed: 7, Skipped: 0, Total: 7`.
Clean tree, full suite: `Failed! - Failed: 3, Passed: 107, Skipped: 0, Total: 110` — the 3 are the pre-existing environmental `BindingTests.QueueWatcherBackend*NoFallback(backend: null)` you already identified. CI is green, so they do not reproduce there.

Extension verified fresh, not a stale dylib: `strings libhonker_ext.dylib | grep -c "RETURNING id, queue, payload, state"` → 1.

Findings 4 and 5 left alone — 4 is new surface with no test rather than a wrong test, and 5 lands with #134.

russellromney:
## Round-two adversarial review — verification write-up

I did not write this code. The round-two pass pushed five commits (`cacf9f5`..`0c67e9b`) but died before writing this up, so this re-derives each finding from the diff and proves each guard by breaking it. Every claim below is marked **RUN** (I executed it) or **INFERRED** (I read it).

Environment: macOS arm64, `dotnet 10.0.102`, extension built from this branch at `target/release/libhonker_ext.dylib`. Baseline before any mutation: **RUN** `dotnet test --filter JobDetailTests` → 12/12 passed; full `dotnet test` → 115/115 passed (the 3 `QueueWatcherBackend*` tests need `Honker.QueueHelper` built in Release first, which CI does and a bare local run does not — not a PR defect).

### Finding 1 — a typed claim decoded the payload it had just claimed (`cacf9f5`)

**INFERRED from the diff:** before the fix, `Job<TPayload>`'s constructor called `_job.GetPayload<TPayload>()`. `TypedQueue<T>.ClaimBatch` builds those inside a `.Select(...).ToList()`, so a payload that did not match `TPayload` threw `JsonException` out of `ClaimBatch`/`ClaimOne`/`ClaimAsync` — after `honker_claim_batch` had already flipped the rows to `processing` in the database. The caller got an exception and no handle, so it could not `Ack`, `Retry`, or `Fail` anything in that batch. Honker never validates payload shape, so a single disagreeing producer stalls a typed worker until the visibility timeout, then again on every claim after.

**RUN — guard broken.** I restored the eager decode (`_payload = job.GetPayload<TPayload>(); _decoded = true;` in the `Job<TPayload>` ctor) and re-ran:

```
Honker.Tests.JobDetailTests.TypedClaimHandsBackAJobWhosePayloadDoesNotDecode [FAIL]
  System.Text.Json.JsonException : The JSON value could not be converted to OrderPayload. Path: $.Quantity
   at Honker.Job.GetPayload[T]() in Models.cs:line 232
   at Honker.Job`1..ctor(Job job) in Models.cs:line 260
   at Honker.TypedQueue`1.ClaimOne(String workerId) in TypedQueue.cs:line 64
Honker.Tests.JobDetailTests.TypedClaimBatchDecodesEveryJobAndOneBadPayloadDoesNotSinkTheRest [FAIL]
   at Honker.TypedQueue`1.ClaimBatch(String workerId, Int32 batchSize) in TypedQueue.cs:line 56
Failed: 2, Passed: 10, Total: 12
```

The stack names `ClaimOne`/`ClaimBatch`, i.e. the throw is inside the claim — the exact defect. Restored → 12/12. Both tests are load-bearing, and the fix does what the message says.

### Finding 2 — a missing core field was silently defaulted (`1e455a2`)

**INFERRED from the diff:** this PR replaced the private `ClaimedJobRow` with the public `JobRow`, which dropped `row.State ?? "processing"`. `JobRow.State` defaulted to `""`. `honker_claim_batch` in 0.5.x and earlier returns six columns and no `state`, so this binding against a released extension would report `Job.State == ""` for every claimed job. The new `worker_id`/`claim_expires_at` throws do not catch it — both are in the old ABI. Making every member `required` turns that into a `JsonException` naming the field.

**RUN — guard broken.** Dropped `required` from `State` alone (`public string State { get; init; } = "";`):

```
Honker.Tests.JobDetailTests.AClaimRowFromAnOlderCoreIsRejectedRatherThanDefaulted [FAIL]
  Assert.Contains() Failure: Sub-string not found
Failed: 1, Passed: 11, Total: 12
```

Load-bearing, and specific to `State` — the other required members still throw, so the test only fails because the message stops naming `state`. Restored → 12/12.

**RUN — the "core emits nulls, never omits them" claim.** `PendingSnapshotCarriesEveryCoreDetail` reads a pending row through the real extension, where `worker_id` and `claim_expires_at` are SQL NULL. With `required` on both, an omitted key would throw. It passes, so the core really does emit them as JSON `null`. **INFERRED** confirmation from `honker-core/src/honker_ops.rs`: both `claim_batch` (line 719) and `get_job` (line 1061) build the object with `json!`, which always writes the key.

### Finding 3 — version number in the docs (`49295b6`)

**INFERRED.** Cosmetic and correct: the doc and test comment said "before 0.6", naming a release that does not exist. Now "0.5.x and earlier". No behaviour.

### Finding 4 — `ExpiresAt` was never asserted null (`b5e224e`) — the commit's stated proof is wrong

The guard itself is real. The proof written in the commit message is not.

**RUN — the mutation the commit message names does not do what it says.** The message claims "pointing `ExpiresAt`'s `JsonPropertyName` at `created_at` fails `DelayedJobReportsItsRunAt` and nothing else in the filter." I ran exactly that. It fails **all 12** tests, and not on the assertion:

```
System.InvalidOperationException : The JSON property name for 'Honker.JobRow.created_at'
  collides with another property.
Failed: 12, Passed: 0, Total: 12
```

Two properties cannot share a `JsonPropertyName`, so `System.Text.Json` throws while building the type's metadata and every test dies before reaching any assertion. That mutation proves nothing about `Assert.Null(row.ExpiresAt)`.

**RUN — the honest mutation.** The message names the real failure mode as "wired to another column, or defaulted to 0". I injected the second one, leaving the mapping alone:

```csharp
private readonly long? _expiresAt;
[JsonPropertyName("expires_at")]
public required long? ExpiresAt { get => _expiresAt ?? 0; init => _expiresAt = value; }
```

```
Honker.Tests.JobDetailTests.DelayedJobReportsItsRunAt [FAIL]
Honker.Tests.JobDetailTests.AClaimRowFromAnOlderCoreIsRejectedRatherThanDefaulted [FAIL]
Failed: 2, Passed: 10, Total: 12
```

So the assertion **is** load-bearing — it just fails alongside one other test, not alone as claimed. Net: keep the test, and treat that commit message's evidence line as wrong on both counts (which tests fail, and why). Nothing to change in the code; recording it here because the commit message is the ledger.

### Round-one trap re-checked: the `run_at`/`created_at` transposition

The trap: a "claimed job reports run_at" test that enqueues with no delay passes for the wrong reason, because `run_at` and `created_at` are then the same second.

**INFERRED:** `TypedQueueDecodesClaimedPayloadsAndKeepsEveryDetail` (line 140) and `TypedSnapshotDecodesPayloadAndCarriesEveryDetail` (line 180) both back-date an absolute `RunAtUnix: Now(db) - 100` and add `Assert.NotEqual(job.RunAt, job.CreatedAt)`. That is the right shape.

**RUN — guard broken.** I transposed the two `JsonPropertyName`s on `JobRow` (`run_at` ↔ `created_at`):

```
Failed ClaimedJobCarriesEveryCoreDetail
Failed DelayedJobReportsItsRunAt
Failed PendingSnapshotCarriesEveryCoreDetail
Failed TypedJobHeartbeatsRetriesAndCancels
Failed TypedQueueDecodesClaimedPayloadsAndKeepsEveryDetail
Failed TypedSnapshotDecodesPayloadAndCarriesEveryDetail
Failed: 6, Passed: 6, Total: 12
```

Caught six ways. Trap cleared.

### New finding — `required` on the public `JobRow` is a source break, and the CHANGELOG said it was not

`1e455a2`'s CHANGELOG entry ended "Nothing in the binding constructs a `JobRow` — it is only a deserialization target — so `required` costs callers nothing." True inside the binding, false outside it: `JobRow` is a `public sealed record` with `init` setters, so any downstream assembly can construct one, and before this PR every member had a default.

**RUN.** Compiled `new Honker.JobRow { Id = 1, Queue = "q" }` against this branch:

```
error CS9035: Required member 'JobRow.Payload' must be set in the object initializer
error CS9035: Required member 'JobRow.State' must be set ...
error CS9035: Required member 'JobRow.Priority' must be set ...
(and RunAt, WorkerId, ClaimExpiresAt, Attempts, MaxAttempts, CreatedAt, ExpiresAt)
```

Still the right trade for a pre-1.0 package — nothing outside a decoder has a reason to build a `JobRow` — but the ledger has to say it is a source break rather than say it costs nothing. Fixed in `cd43e5b`.

### New observation — `required` relocates the old-core failure rather than removing it

**INFERRED.** Against a 0.5.x extension, `JsonSerializer.Deserialize<List<JobRow>>` now throws inside `Queue.ClaimBatch` — after `honker_claim_batch` has claimed the rows. That is the same stranding shape Finding 1 removed for payloads, and the two CHANGELOG entries read as contradicting each other.

It is still the right call, and the reason is that the two cases differ in kind: a version-mismatched core fails on the very first claim and every claim after, so the worker never appears to run and an operator sees it immediately; one producer's mismatched payload is an ordinary production event a healthy worker has to survive. Recorded next to the entry in `cd43e5b` so the trade is written down instead of implied. No code change.

### Checked and clean

- **12-column contract.** **INFERRED.** This binding maps by JSON key, not by column position, so `claim_batch`'s `RETURNING` order cannot bite it. Each of the twelve keys the core emits has exactly one `JsonPropertyName` on `JobRow`, spelled identically; `JobSnapshot<T>.FromRow` assigns all twelve, and the transposition mutation above proves the two easily-swapped ones are pinned.
- **Null/absent handling.** **RUN.** `worker_id`/`claim_expires_at` null on a pending read (`PendingSnapshotCarriesEveryCoreDetail`), non-null and range-checked on a claimed one (`ClaimedJobCarriesEveryCoreDetail`), `expires_at` null with no expiry (`DelayedJobReportsItsRunAt`), and back to null after `Retry` (`TypedJobHeartbeatsRetriesAndCancels`). `JsonSerializer.Deserialize<TPayload>` returning null is documented as meaning the stored JSON is literally `null`.
- **No swallowed decode errors in the binding.** **RUN** `grep -rn catch packages/honker-dotnet/src/Honker/`: six sites, none near a payload or row decode. `Database.cs:171` rethrows; `Queue.cs:121` is a narrow `when (ex.Message.Contains("no such function"))` filter behind an existing `FunctionExists` probe. No catch-all sits between a `JsonException` and the caller.
- **Public API.** **INFERRED.** `Job` gained `State`, `Priority`, `RunAt`, `MaxAttempts`, `CreatedAt`, `ExpiresAt`; its constructor is `internal`, so the reshaped signature is not a break. `Database.Queue` keys its cache on `(name, QueueOptions)`, so `db.Queue<T>(name, opts)` and `db.Queue(name, opts)` really do share one underlying `Queue` (asserted by `Assert.Same` in `TypedAndUntypedHandlesShareOneQueue`) and differing options do not silently reuse a handle built with other options. `TypedQueue<T>.GetJob` and `Queue.GetJob` not being queue-scoped is documented on both members and tracked in #134.

### Out of scope, flagging loudly

`honker-core/src/honker_ops.rs:1043` — `get_job` ends its `conn.query_row(...)` with `.ok()`, which discards every error, not just `QueryReturnedNoRows`, and returns `Ok(String::new())`. Every binding then reports "job not found" for what may have been an I/O or schema failure. Four sibling sites do the same (lines 893, 961, 1173, 1649). **This is pre-existing on `main` and untouched by this PR**, and open PR #138 ("fix(core): propagate SQLite errors from five lookups") already covers exactly these five. Noting it so it is not mistaken for something #142 introduced; no core edit made here.

### Verdict

The round-two fixes are real and their guards are load-bearing — proven by breaking each one and watching the right tests fail. One commit message (`b5e224e`) states evidence that does not reproduce; the test it added is still worth keeping. One CHANGELOG claim was wrong and is corrected in `cd43e5b`, along with the stranding trade being written down. Nothing else outstanding from my pass. Not merged, `main` untouched, no force-push.

PR #143, opened 2026-09-01
feat(python): keep every claimed-job field and add payload type hints

russellromney:
Part of #136 — the Python binding's slice of "full job details and typed payloads in every binding". Covers the issue's Python checklist line: keep all claim fields, add `expires_at`, fix the old comment that says claim results are narrow.

**Stacked on #124 (`feat/node-typed-jobs`). Merge #124 first.** The widened
`honker_claim_batch` RETURNING clause that exposes the full job snapshot lives
only on that branch; on `main` claim_batch still returns the old 8 fields, so
this PR targets `feat/node-typed-jobs` as its base.

## What changed

- `Job` now carries all twelve fields: `expires_at` is new, and `state`,
  `priority`, `run_at`, `max_attempts`, `created_at` are read straight out of
  the claim result instead of `row.get(..., default)`. Those defaults were
  stale — they described a narrow claim path that no longer exists — and a
  default on a field the core does return would silently paper over ABI drift
  between the core and this binding.
- The "narrow claim path only returns hot-path fields" comment is replaced with
  what the ABI actually does.
- `Job.last_error` is now set to `None` explicitly, with a comment saying why:
  `honker_claim_batch` does not return it. It was already always None; the
  `row.get("last_error", None)` fallback just made that look accidental.
- New `honker.JobSnapshot` TypedDict names the twelve keys `get_job()` returns.
- `Queue` and `Job` take an optional payload type parameter (`Queue[T]`,
  `Job[T]`) that flows through `claim_one()`, `claim_batch()`, `claim()`, and
  `Job.payload`.
- README section documenting the field list and the payload hints.

No core or SQL ABI changes. `packages/honker-node` untouched. `claimed_at`
(#135) is deliberately not here.

## No behavior change to `get_job()`

`get_job()` still returns a plain dict with the core's snake_case keys and a
raw-JSON-text `payload`. `JobSnapshot` is a TypedDict — a type hint over the
same runtime value — so nothing existing breaks. Python's own naming style is
snake_case, so the core's keys already read naturally here.

## Not runtime validation

`Queue[T]` / `Job[T]` are compile-time hints and nothing more. Honker does not
inspect or validate payload shape in the database, so a payload that does not
match the annotation still enqueues and claims fine.
`test_typed_payload_hints_do_not_validate_at_runtime` pins that on purpose: it
must start failing if someone ever adds runtime schema checking.

## Tests

`tests/test_job_details.py` (5 tests):

- `test_claimed_job_and_snapshot_carry_every_field` — enqueues with a priority
  (7), TTL (600s), and max-attempts (5), then asserts every one of the twelve
  fields against its expected *value* on three reads: the pending snapshot, the
  claimed `Job`, and a second independent `honker.open()` handle's processing
  snapshot. Checks `expires_at - run_at == 600` exactly, `claim_expires_at`
  inside the claim-time + 120s visibility window, `attempts` going 0 → 1,
  `worker_id` / `claim_expires_at` None while pending and set once claimed, and
  `run_at` / `created_at` unchanged by the claim. Asserts the snapshot's key set
  is exactly the twelve. After ack, the reader gets `None`.
- `test_delayed_job_reports_its_run_at` — `delay=3600` reports the right
  `run_at` and is not claimable early.
- `test_job_without_expiry_reports_none` — no `expires=` means `expires_at` is
  None on both the snapshot and the claimed job.
- `test_retry_moves_the_job_back_to_pending_with_the_new_run_at` — a second
  claim reports the state the first claim's retry wrote (attempts 1 → 2,
  priority and created_at preserved, worker_id cleared).
- `test_typed_payload_hints_do_not_validate_at_runtime` — described above.

```
$ python -m pytest tests/ -q -n 4 -m 'not slow and not linux_only'
296 passed, 25 skipped in 24.86s
```

### Load-bearing proof

Replaced `self.expires_at = row["expires_at"]` with `self.expires_at = None`
in `Job.__init__` and reran:

```
FAILED tests/test_job_details.py::test_claimed_job_and_snapshot_carry_every_field
E   assert None == 1788255166
     +  where None = <honker._honker.Job object>.expires_at
tests/test_job_details.py:83
1 failed, 4 passed
```

Restored, full suite green again.


russellromney:
## Adversarial review — #143

Labelled **RAN** (executed) or **INFERRED** (read, not executed).

### 1. The suite cannot tell `run_at` from `created_at` on a claimed job — **RAN**

Both of these mutations leave `tests/test_job_details.py` at **5/5 passing**:

```python
self.run_at = row["created_at"]      # 5 passed
self.created_at = row["run_at"]      # 5 passed
```

Controls that *do* fail, so the rest of the mapping is genuinely covered:

```python
self.expires_at = row["claim_expires_at"]   # 2 failed
self.attempts   = row["max_attempts"]       # 2 failed
```

Cause: `test_delayed_job_reports_its_run_at` uses `delay=3600`, which makes the job **unclaimable**, so the claimed path is never exercised with `run_at != created_at`. Every other test enqueues with no delay, where the two are equal to the second.

**Fix — one line.** Backdate `run_at` instead of delaying it, so the job still claims:

```python
job_id = q.enqueue(payload, run_at=int(time.time()) - 100)
job = q.claim_one("w")
assert job.run_at != job.created_at
```

RAN: this yields `run_at=1788316141`, `created_at=1788316241`, and `claim_one` still returns the job. The same hole is mutation-proven in #124 (Node), #148 (Ruby) and #149 (Elixir); the same backdating fix works in all four.

### 2. `get_job` is not queue-scoped here, but it is in Node — **RAN**

`emails.get_job(sms_id)` returns the SMS row. That is deferred to #134 and the Ruby/Elixir PRs say so in their docstrings, but Python's docstring says only "or None if ack'd, dead'd, or never existed" — which reads as the complete miss list and is now wrong relative to Node's behaviour. Worth one sentence.

### 3. Cross-binding payload divergence — **RAN**

- A non-JSON payload makes Python's `Job.payload` raise `json.JSONDecodeError`; Node returns the raw string for the same row.
- Python's snapshot `payload` is raw JSON text; Node's is decoded.

Same field, three behaviours across bindings. #148/#149 document this explicitly; it would be good to say it here too.

### 4. The new type hints are ungated — low, **INFERRED**

#143 adds `Generic`/`TypedDict` hints but no mypy or pyright step in CI, so nothing verifies them (#124 did add `tsc`). `Database.queue()` is also not parameterised, so `honker.Queue[EmailPayload] = db.queue(...)` only type-checks through `Any`.

### What holds up — **RAN**

All twelve fields present (`set(pending) == set(JOB_FIELDS)`). `worker_id`, `claim_expires_at` and `expires_at` are `None` rather than `0`/`""`. Acked, dead-lettered and never-existed all return `None`. Dropping the `row.get(..., default)` fallbacks is the right call — no silent defaults on data that matters.

**Recommendation:** merge once the `run_at` test is tightened. The field mapping itself is correct; only the test's ability to detect a regression is missing.

russellromney:
## Finding 1 fixed — the `run_at` / `created_at` gap is now load-bearing

Commit: `e07572b` — `test(python): prove a claimed job reports run_at, not created_at`

Added `test_claimed_job_reports_its_own_run_at_not_created_at` to `tests/test_job_details.py`. It enqueues with `run_at=int(time.time()) - 100` — a **back-dated absolute run_at** rather than a delay — so the job stays claimable while `run_at != created_at`, and it pins `run_at` to the exact value enqueued instead of comparing it to another field.

### Mutation proof

Mutated `Job.__init__` in `packages/honker/python/honker/_honker.py:232/237`:

```python
self.run_at = row["created_at"]
self.created_at = row["run_at"]
```

```
    def test_claimed_job_reports_its_own_run_at_not_created_at(tmp_path):
        ...
        job = q.claim_one("worker-py")
        assert job is not None, "a back-dated job must still be claimable"
        assert job.id == jid
>       assert job.run_at == run_at
E       assert 1788318882 == 1788318782
E        +  where 1788318882 = <honker._honker.Job object at 0x10a849c60>.run_at

tests/test_job_details.py:193: AssertionError
=========================== short test summary info ============================
FAILED tests/test_job_details.py::test_claimed_job_reports_its_own_run_at_not_created_at
1 failed, 5 passed in 0.83s
```

**Only the new test fails** — the other 5 stay green, which is the blind spot you reported. Source restored; `git diff` on `packages/` is empty.

Python's `get_job` returns the raw row dict, so the `Job` constructor is the only mapping site to mutate here.

### Local run

Full suite: **297 passed, 25 skipped**, exit 0. `test_job_details.py` alone: 6 passed (was 5).

Findings 2, 3 and 4 are untouched — this commit is scoped to finding 1. No API or type-hint changes.

russellromney:
## Second adversarial review — #143 (Python)

Deeper pass than round one: is this complete, well designed, bug-free, comprehensively tested, and does it make sense to a Python user?

Labelled **RAN** (executed and observed) or **READ** (reasoned from source, not executed). Five commits pushed to this branch. No build: everything ran against a prebuilt native.

---

### 1. HIGH — the type hints did nothing. No `py.typed`. **RAN**

The PR's headline feature is `Queue[T]` / `Job[T]` / `JobSnapshot`, and **no type checker could see any of it.**

PEP 561: a checker must ignore inline annotations from an installed package with no `py.typed` marker. `packages/honker/python/honker/` had none. Installed the package into a clean venv and ran mypy on exactly the example the README shows:

```
error: Skipping analyzing "honker": module is installed, but missing
       library stubs or py.typed marker  [import-untyped]
note: Revealed type is "Any"    # emails: honker.Queue[EmailPayload] = db.queue(...)
note: Revealed type is "Any"    # job.payload
note: Revealed type is "Any"    # job.expires_at
note: Revealed type is "Any"    # emails.get_job(1)
```

Every hint in the PR resolved to `Any` in user code. Node ships `wrapper.d.ts` and points `"types"` at it; Python shipped annotations and no marker.

**Fixed** (`45a903d`): added the marker, added it to maturin's wheel `include`, and asserted it in `scripts/proof/check-python-wheel.py` beside the native and the bundled extension. Three independent gates now: the wheel proof, a pytest that checks `honker.__file__`'s directory on every OS in the matrix, and the mypy step in finding 4 (which fails with 25 `assert-type` errors the moment the marker goes missing — RAN, that mutation).

Also verified the marker is honoured through a `maturin develop`-style `.pth` editable install, not only a plain site-packages copy — RAN, both shapes.

### 2. HIGH — all twelve of `Job`'s fields were `Any`. **RAN**

With the marker in place, `JobSnapshot` typed cleanly but the claimed `Job` did not:

```
reveal_type(job.expires_at)   ->  Any
reveal_type(job.attempts)     ->  Any
...all twelve
```

`Job.__init__` reads them out of an untyped `row: dict`, and `__slots__` carries no types. So the exact field list this PR is about — the point of #136 — was invisible to a checker, while the read-only snapshot beside it was fully typed. Backwards.

**Fixed** (`6acccd1`): class-level annotations for all twelve. Checked at runtime that bare annotations alongside `__slots__` still leave `Job` `__dict__`-free and still reject unknown attributes — RAN.

### 3. MEDIUM — `Queue[T]` never reached the producer side. **RAN**

`enqueue(payload: Any)`, so `Queue[EmailPayload].enqueue({"wrong": "shape"})` type-checked fine. The producer is where a payload-shape hint actually catches something — consumers only ever see what producers wrote. Node types both sides (`enqueue(payload: TPayload, ...)`).

**Fixed** (`6acccd1`): `enqueue(payload: PayloadT)`. mypy now rejects both a wrong shape and a wrong type — RAN, and mutation-proven: reverting to `Any` makes the gate fail.

### 4. MEDIUM — no type checker in CI (round one's finding 4). **RAN**

Node gates `wrapper.d.ts` with `tsc`; Python's hints were ungated, so nothing would have caught findings 1, 2, or 3.

**Fixed** (`91d47a7`): `scripts/proof/typing/check_payload_hints.py`, checked by mypy on one matrix leg (Linux/3.12 — the hints do not vary by OS or interpreter). Nothing in it runs; it is type-level only.

Load-bearing by construction, three ways:

- It resolves `honker` through site-packages, the way a user's project does, so a missing `py.typed` fails it.
- `assert_type` fails against `Any`, so a field that decays back to `Any` fails it.
- `--warn-unused-ignores` turns each `# type: ignore[...]` into an assertion that mypy **still rejects** that line, so the negative cases (bad enqueue payload, unknown TypedDict key) cannot silently stop being errors.

Mutations RAN: delete `py.typed` → 25 errors. Rename the `expires_at` annotation → 3 errors including the `JobSnapshot` key. Revert `enqueue` to `Any` → 2 `unused-ignore` errors.

### 5. `Database.queue()` unparameterized (round one's finding 4b) — verified, judged, not "fixed". **RAN**

Verified: `db.queue("emails")` is `Queue[Any]`, and `emails: honker.Queue[EmailPayload] = db.queue(...)` type-checks through `Any`. Accurate, but not a defect — it is the normal Python idiom, the same one as `xs: list[int] = json.loads(text)`. A queue *name* carries no payload type, and Python has no call-site type-argument syntax. Once the annotation binds `T`, every later use of `emails` is genuinely checked, and now on both sides (finding 3).

I considered a runtime `payload_type=` parameter on `db.queue()` and rejected it: it would join the memoized-options equality check (`ValueError` on a mismatched re-open) for a value with no runtime meaning — real complexity buying a hint we already have. Instead `Database.queue()` now returns `"Queue[Any]"` explicitly, and the docstrings and README say plainly that the annotation is what binds the type and that the binding is *asserted*, not checked.

### 6. `get_job()` keeps raw-text `payload` — right call, now documented. **READ**

Judged: keeping it is correct. #146 exists precisely because Node decodes its snapshot while Python and Go return raw text, and it says the binding PRs deliberately preserve each binding's encoding since changing it is a breaking API change #136 did not ask for. Changing it here breaks every caller already calling `json.loads(row["payload"])` and forks the decision away from #146.

What was wrong was that a Python user had no way to know. `JobSnapshot`, `get_job()` and the README now say the snapshot's payload is raw text, that a claimed `Job.payload` is decoded, and that #146 tracks reconciling them.

### 7. `Job.payload` raising on non-JSON — right call, now documented. **READ**

Round one flagged that Python raises `json.JSONDecodeError` where Node returns the raw string. Python is right. Node's `parseJson` catches and hands back the text, turning an undecodable payload into a silent violation of `TPayload` that surfaces somewhere far away. Raising is the loud outcome. That is Node's divergence to close, and #153 makes it moot by rejecting non-JSON at every core write path.

Documented on the property, including what actually happens per caller — `Outbox.run_worker` turns it into a retry, the decorated-task worker loop lets it escape and stop that worker. I read both paths rather than assuming a uniform retry.

Deliberately **no test** pinning it: writing a non-JSON payload means calling `honker_enqueue` with raw text, which #153 is about to make impossible. A test there would exist only to be deleted.

### 8. `get_job()` is not queue-scoped (round one's finding 2) — documented, not changed. **READ**

Left per #134 as instructed. The docstring said "or None if ack'd, dead'd, or never existed", which reads as the complete miss list and is now wrong next to Node's scoped `getJob`. It now says outright that `emails.get_job(sms_id)` returns the SMS row, that the payload will not match this queue's `PayloadT`, and that #134 tracks pushing the filter into the core where it can stay atomic. Same in the README.

### 9. MEDIUM (test) — the retry test never checked the `run_at` it is named for. **RAN**

`test_retry_moves_the_job_back_to_pending_with_the_new_run_at` asserted state, attempts and `worker_id`, and never looked at `run_at`. It could not: the job was enqueued at *now* and retried with `delay_s=0`, and `retry` sets `run_at = unixepoch() + delay_s`, so old and new were equal to the second. The same shape of blind spot as the one fixed in `e07572b`.

**Fixed** (`13f5ff0`): back-date the enqueue by 100s. The rewrite becomes measurable while `delay_s=0` keeps the job claimable for the second claim. Now asserts the pending row's `run_at` moved to retry time and that the second claimed `Job` reports it.

Side effect, mutation-proven RAN: `self.run_at = row["created_at"]` now fails **two** tests instead of one.

### 10. LOW — the README named a field that does not exist on `Job`. **READ**

"A claimed `Job` and a `get_job()` snapshot both carry the full job shape: `id`, `queue`, …" — but `job.queue` is the `Queue` object the claim methods call. The column is `job.queue_name`, and no example showed it. Fixed in the README and the `Job` docstring.

---

### What holds up

Round one's "what holds up" still does: twelve fields present and asserted by value, `worker_id` / `claim_expires_at` / `expires_at` genuinely `None` and never `0` or `""`, acked / dead-lettered / never-existed all return `None`, and dropping the `row.get(..., default)` fallbacks is right — a default on a field the core does return would hide ABI drift. No correctness bug in the field mapping itself; the twelve columns were read correctly all along.

Full Python suite after all changes: **267 passed, 56 skipped**, exit 0 (RAN, against a prebuilt native — no cold build). `tests/test_job_details.py`: 7 passed, was 6.

### Left deliberately

- **`Outbox` is not generic** while Node has `Outbox<TPayload>`. It is the binding's other enqueue surface, so this is a real coherence gap — but it needs `delivery: Callable[[PayloadT], Any]`, hits the same "cannot infer from a name" problem on `db.outbox()`, and touches the memoized-delivery equality check. Outside #136's Python checklist line and outside what this PR claims. Worth its own change.
- **The `_honker_jobs` inspection view omits `expires_at`.** `_honker_dead` has no `expires_at` column, so the UNION's dead branch would have to fill it with a `NULL` that is a lie. Private underscore view, not the public job surface this PR is about.
- **`Job.state` is mutation-undetectable.** Hard-coding `self.state = "processing"` passes every test, because on the claimed path state *is* `'processing'` by construction — nothing can distinguish it. Harmless and unfixable from the claim side; the `get_job` path is covered (`pending["state"] == "pending"`).
- `claimed_at` (#135), queue-scoped `get_job` (#134), payload encoding (#146), core JSON contract (#153) — all out of scope, all now referenced from the code so a reader lands on the right issue.

Nothing touched outside the Python binding except `CHANGELOG.md`, `scripts/proof/`, and the Python job in `ci.yml`. No core, no other binding.

### Verdict

**Merge**, after #124. Findings 1 and 2 meant the PR's headline feature was inert for every user; both are fixed and gated so they cannot regress silently.

Not merged, per instructions.

PR #144, opened 2026-09-01
feat(kotlin): add a JobDetails data class over the JVM job fields

russellromney:
Part of #136 — the Kotlin half of the binding checklist.

**Stacked twice. Merge order: #124 → #141 → this PR.**
`honker-kotlin` is a thin wrapper over `dev.honker:honker`, so this
branch is based on #141 (the JVM half), which is itself based on
`feat/node-typed-jobs` (#124) because the widened `honker_claim_batch`
RETURNING clause exists only there. The diff below is Kotlin-only; the
JVM commits show up because of the stack.

## What changed

`packages/honker-kotlin` plus one CI step. No core, no SQL ABI, no other
binding.

- New `JobDetails<T>` data class in `dev.honker.kotlin`. It carries every
  field the core returns — `id`, `queue`, `payload`, `payloadJson`,
  `state`, `priority`, `runAt`, `workerId`, `claimExpiresAt`, `attempts`,
  `maxAttempts`, `createdAt`, `expiresAt` — restated with Kotlin
  nullability, so the compiler knows a pending job has no `workerId` or
  `claimExpiresAt` and an unexpiring job has no `expiresAt`. Times are
  unix epoch seconds.
- Data only. Ack, retry, fail and heartbeat stay on the claimed `Job`.
- `Job.details()`, `TypedJob<T>.details()`, `JobSnapshot.details()` and
  `TypedJobSnapshot<T>.details()` build one.
- `Queue.jobDetails(id)`, `TypedQueue<T>.jobDetails(id)` and
  `Database.jobDetails(id)` look one up and return Kotlin `null` rather
  than `Optional.empty()`. The queue lookups stay scoped to their queue;
  the database lookup is global.

It is named `JobDetails`, not `JobSnapshot`, on purpose: `dev.honker.JobSnapshot`
already exists and users routinely star-import both packages, so a
same-named Kotlin class would be an ambiguity error at every use site.

`claimed_at` is deliberately not here; that is #135.

## Typing is compile-time only

The payload type is the JVM generic. Honker does not validate payload
shape in the database and this PR adds no runtime validation. README and
CHANGELOG both say so.

## CI

The Kotlin binding had **no CI job at all** — `make test-kotlin` was the
only way to run it. Added a `Run honker-kotlin tests` step to the
existing `jvm` job, which already builds the extension and installs
`dev.honker:honker`. The stale dependabot comment claiming neither JVM
nor Kotlin is gated is updated to match.

## Tests

Five new tests in `HonkerKotlinTest`, all asserting values rather than
presence:

- `claimedJobDetailsCarryEveryCoreField` — enqueues with an explicit
  `runAt`, `priority(7)`, `maxAttempts(5)` and `expires(600s)`, claims,
  and checks all twelve fields. The clock-derived three are bounded by
  `unixepoch()` readings taken around the write and the claim.
- `jobDetailsFollowPendingProcessingAndAckedStates` — a second `Database`
  on the same file reads the job pending (`workerId` and
  `claimExpiresAt` null, 0 attempts), then processing (worker id,
  attempts 1, claim deadline in the visibility window), then null after
  ack.
- `delayedJobDetailsReportItsRunAt`
- `typedQueueDetailsDecodePayloadAndKeepEveryField`
- `jobDetailsAreScopedToTheirQueueWhileDatabaseLookupIsGlobal`

Local run: `mvn -B test` → 10/10 pass (5 existing + 5 new), extension
built `--release`, `dev.honker:honker` installed from this branch.

### The tests are load-bearing

Swapped `attempts` and `maxAttempts` in `JobSnapshot.toDetails` — the
classic mixup — and reran:

```
[ERROR] Tests run: 2, Failures: 2, Errors: 0
HonkerKotlinTest.claimedJobDetailsCarryEveryCoreField:144
  expected: <1> but was: <5>
HonkerKotlinTest.jobDetailsFollowPendingProcessingAndAckedStates:191
  expected: <0> but was: <4>
```

Restored, reran: 10/10 pass.

## Not touched

`honker-core`, the SQL ABI, `packages/honker-node`, `packages/honker-jvm`
(beyond what the stack carries), every other binding.

---

## Snapshot payload encoding (cross-binding note, no change requested)

Flagging for a later deliberate decision across all bindings, per the
#136 coordination: **the Kotlin binding returns snapshot payloads as raw
JSON text by default.** `JobDetails.payloadJson` is always the string the
row holds. `JobDetails.payload` is decoded only when the caller asked for
it — `jobDetails(id, codec)` or `TypedQueue<T>.jobDetails(id)`. The
untyped `Queue.jobDetails(id)` returns `JobDetails<String>` whose
`payload` is that same raw JSON.

That matches the JVM binding it wraps, and matches Go's `JobRow.Payload`
and Python's `get_job()["payload"]`. Node #124 returns a decoded payload
instead. Three bindings, two conventions. Not resolved here — changing it
would be a breaking API change #136 did not ask for.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -182,9 +182,8 @@ updates:
     commit-message:
       prefix: "deps"
 
-  # honker-jvm and honker-kotlin have no CI job yet, so Dependabot is the
-  # only thing watching them. Keep the PRs coming even though nothing
-  # gates them — they at least surface CVEs.
+  # The `jvm` CI job now runs both binding test suites, so Dependabot PRs
+  # here are gated. Keep them coming: they also surface CVEs.
   - package-ecosystem: maven
     directories:
       - "/packages/honker-jvm"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -694,7 +694,7 @@ jobs:
           python -m pytest tests/test_ruby_python_interop.py -q --tb=short
 
   jvm:
-    name: jvm · honker + documented ORM recipes
+    name: jvm · honker + kotlin + documented ORM recipes
     timeout-minutes: 20
     runs-on: ubuntu-latest
     steps:
@@ -732,6 +732,14 @@ jobs:
         #
         # The other 44 tests gate, and so do the ORM recipes below.
         run: mvn -B test -Dtest='!HonkerJvmTest#childJvmKernelListenerWakesWhenParentNotifies'
+      - name: Run honker-kotlin tests
+        # The Kotlin wrappers compile and run against the dev.honker:honker
+        # jar installed above, so this gates the Kotlin surface on the same
+        # extension build.
+        working-directory: packages/honker-kotlin
+        env:
+          HONKER_EXTENSION_PATH: ${{ github.workspace }}/target/release/libhonker_ext.so
+        run: mvn -B test
       - name: ORM · JVM (JDBC, Hibernate, jOOQ)
         working-directory: scripts/proof/orm/jvm
         env:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,14 @@
 - The `jvm` CI job now runs `packages/honker-kotlin` tests too, against
   the same extension build and the `dev.honker:honker` jar it installs.
   The Kotlin binding previously had no CI job at all.
+- Fix: `Queue.asFlow` and `Listener.asFlow` closed their channel only
+  when the collector cancelled. If the producer thread died on anything
+  other than `HonkerClosedException` it printed a stack trace and left
+  every collector parked on the flow forever. Both now end the flow, and
+  hand a real failure to the collector instead of swallowing it.
+- `HonkerKotlinTest` carries a 60-second per-test timeout so a wrapper
+  that stops producing fails by name instead of burning the CI job's
+  20-minute budget with no diagnosis.
 
 ## Unreleased — JVM job parity
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,24 @@
 - Regression test claims a job at a deadline with one claim forced to
   return empty: it fails on the unpatched `api.js` and passes on the fix.
 
+## Unreleased — Kotlin job parity
+
+- New `JobDetails<T>` data class in `dev.honker.kotlin`: every field the
+  core returns for a job row, restated with Kotlin nullability so the
+  compiler knows a pending job has no `workerId` or `claimExpiresAt` and
+  an unexpiring job has no `expiresAt`. Data only — the claim operations
+  stay on the claimed `Job`.
+- `Job.details()`, `TypedJob<T>.details()`, `JobSnapshot.details()` and
+  `TypedJobSnapshot<T>.details()` build one. `Queue.jobDetails(id)`,
+  `TypedQueue<T>.jobDetails(id)` and `Database.jobDetails(id)` look one
+  up and return Kotlin `null` instead of `Optional.empty()`. The queue
+  lookups are scoped to their queue; the database lookup is global.
+- Payload typing is the JVM generic, so it stays a compile-time contract.
+  No runtime payload validation was added.
+- The `jvm` CI job now runs `packages/honker-kotlin` tests too, against
+  the same extension build and the `dev.honker:honker` jar it installs.
+  The Kotlin binding previously had no CI job at all.
+
 ## Unreleased — JVM job parity
 
 - JVM claimed `Job` and `TypedJob<T>` now carry every field the core
--- a/packages/honker-kotlin/README.md
+++ b/packages/honker-kotlin/README.md
@@ -80,11 +80,17 @@ does not match the queue's type. `Database.jobDetails(id)` is the global
 lookup.
 
 Pass a `JsonCodec<T>` — or use a `TypedQueue<T>` — to get the payload
-decoded:
+decoded. Both give the same `JobDetails<T>`; `payloadJson` is always the
+raw row text either way:
 
 ```kotlin
 val typed = db.queue("emails").typed(strings)
 val details: JobDetails<String>? = typed.jobDetails(id)
+
+// Same thing without a typed queue. The codec overload exists on
+// `Queue.jobDetails`, `Database.jobDetails`, `Job.details` and
+// `JobSnapshot.details`.
+val same: JobDetails<String>? = db.queue("emails").jobDetails(id, strings)
 ```
 
 The payload type is a compile-time contract between the callers that
--- a/packages/honker-kotlin/README.md
+++ b/packages/honker-kotlin/README.md
@@ -12,6 +12,8 @@ Current shape:
 - adds coroutine `TaskResult.await(...)`
 - adds codec-based typed queue/task helpers without forcing a JSON
   library dependency
+- adds a `JobDetails<T>` data class over the JVM job fields, with Kotlin
+  nullability and nullable `jobDetails(id)` lookups instead of `Optional`
 
 ## Quick start
 
@@ -43,6 +45,50 @@ val job = db.queue("emails").asFlow("worker").first()
 println(job.decode(strings))
 job.ack()
 ```
+
+## Job fields
+
+`JobDetails<T>` is a Kotlin data class carrying every field the core
+returns for a job row: `id`, `queue`, `payload`, `payloadJson`, `state`,
+`priority`, `runAt`, `workerId`, `claimExpiresAt`, `attempts`,
+`maxAttempts`, `createdAt`, `expiresAt`. Times are unix epoch seconds.
+`workerId`, `claimExpiresAt` and `expiresAt` are Kotlin nullable, so the
+compiler knows a pending job has no worker.
+
+It is data only. Ack, retry, fail and heartbeat stay on the claimed
+`Job`, because only the worker holding the claim may call them.
+
+```kotlin
+val emails = db.queue("emails")
+val id = emails.enqueueJson("""{"to":"alice@example.com"}""")
+
+val pending = emails.jobDetails(id)!!
+println(pending.state)      // "pending"
+println(pending.workerId)   // null
+
+val claimed = emails.claimOne("worker-1").orElseThrow()
+val job = claimed.details()
+println("${job.state} ${job.attempts}/${job.maxAttempts} ${job.priority}")
+claimed.ack()
+
+emails.jobDetails(id)       // null
+```
+
+`jobDetails(id)` on a `Queue` only returns jobs from that queue; job ids
+are globally unique, so an unscoped hit could hand back a payload that
+does not match the queue's type. `Database.jobDetails(id)` is the global
+lookup.
+
+Pass a `JsonCodec<T>` — or use a `TypedQueue<T>` — to get the payload
+decoded:
+
+```kotlin
+val typed = db.queue("emails").typed(strings)
+val details: JobDetails<String>? = typed.jobDetails(id)
+```
+
+The payload type is a compile-time contract between the callers that
+share a queue. Honker does not validate payload shape in the database.
 
 ## Local test
 
--- a/packages/honker-kotlin/src/main/kotlin/dev/honker/kotlin/HonkerKotlin.kt
+++ b/packages/honker-kotlin/src/main/kotlin/dev/honker/kotlin/HonkerKotlin.kt
@@ -80,7 +80,15 @@ fun Listener.asFlow(timeout: Duration = Duration.ofSeconds(15)): Flow<Notificati
             while (!Thread.currentThread().isInterrupted) {
                 next(timeout).ifPresent { trySend(it).isSuccess }
             }
+            close()
         } catch (ignored: HonkerClosedException) {
+            // The database went away under us: the flow simply ends.
+            close()
+        } catch (failure: Throwable) {
+            // Anything else is a real failure. Hand it to the collector.
+            // Without this the producer thread dies and every collector
+            // parks on this flow forever.
+            close(failure)
         }
     }
     thread.isDaemon = true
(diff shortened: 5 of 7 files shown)


russellromney:
Adversarial review (I did not write this code). Everything below was RUN unless marked INFERRED, in a detached worktree at `3cb1403` against a prebuilt extension.

**No defects found.** Attacked hard because of the standing caveat that Kotlin had no CI job before this PR and its five pre-existing tests had never run anywhere — they run, and they pass.

```
mvn -B test (run twice):
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.612 s -- BUILD SUCCESS
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.448 s -- BUILD SUCCESS
```

### The five never-run tests

`flowWrapperUsesJavaStreamRuntime`, `listenerAndQueueFlowWrappersUseJavaRuntime`, `taskResultAwaitWrapsJavaResult`, `typedCodecHelpersWorkForQueuesAndTasks`, `dslHelpersForwardJavaRuntimeAndValidation` all compile and pass. No skips, no flakiness across two runs (2.6 s and 1.4 s). The whole file compiled cleanly under Kotlin 2.4.10, including the two things I expected to break: the synthetic-property access `Optional.isEmpty`, and `List.first()` coexisting with the `kotlinx.coroutines.flow.first` import.

### Mutation results — all five caught

Applied to `packages/honker-kotlin/src/main/kotlin/dev/honker/kotlin/Jobs.kt` (K5 to the inherited `Queue.java`, with a jar reinstall), each restored afterwards:

| Mutation | Result |
|---|---|
| `attempts` ↔ `maxAttempts` | **caught** — 3 failures |
| `runAt = createdAt()` | **caught** — 4 failures |
| `claimExpiresAt` ↔ `expiresAt` | **caught** — 3 failures |
| `workerId ?: ""`, `claimExpiresAt ?: 0L`, `expiresAt ?: 0L` | **caught** — 2 failures |
| `Queue.getJob` scoping filter deleted, jar rebuilt | **caught** — 1 failure (`jobDetailsAreScopedToTheirQueueWhileDatabaseLookupIsGlobal`) |

The tests inherit #141's discipline: `runAt = unixNow(db) - 5` with an exact `assertEquals`, `claimExpiresAt` at vt+300/120/60 vs `expiresAt` at +600/900. The run_at/created_at trap that survives in #139, #142 and #150 does not survive here. `JobDetails.toDetails` (Jobs.kt:45-59) maps all twelve fields 1:1 with no crossings.

### No silent skips

The Kotlin pom has no antrun step, so nothing is packaged and `NativeLoader` walks up from `packages/honker-kotlin` to the repo root's `target/release/libhonker_ext.dylib`. Forcing `HONKER_EXTENSION_PATH=/nonexistent/...` gives `Tests run: 10, Failures: 0, Errors: 10, Skipped: 0` — `HonkerLoadException` on every test, never a skip. There are no `assumeTrue` calls in `HonkerKotlinTest.kt`.

### No conflict with the #134 scoping decision

Kotlin defines no `getJob` of its own. `Queue.jobDetails` delegates to #141's scoped `Queue.getJob`; `Database.jobDetails` delegates to the global `Database.getJob`. `packages/honker-kotlin/README.md:77-80` already states the split correctly — "`jobDetails(id)` on a `Queue` only returns jobs from that queue; job ids are globally unique, so an unscoped hit could hand back a payload that does not match the queue's type. `Database.jobDetails(id)` is the global lookup." — so no doc goes stale. This PR is the only one in the series besides #141 that already documents the scoped behavior rather than the global one.

### CI wiring is honest

The new step runs `mvn -B test` with **no exclusions**, so all 10 tests gate. It sets `HONKER_EXTENSION_PATH` to the release `.so` the job already built, and it runs after `mvn -DskipTests install` of `dev.honker:honker`, so the dependency is present. The `dependabot.yml` comment was correctly flipped from "no CI job yet, nothing gates them" to "now gated". The `make test-kotlin` target the README cites exists (Makefile line 61).

Would merge.

Environment note: the machine hit 100% disk mid-review and Bash was dead for ~10 minutes. I used the shared `~/.m2` in offline mode rather than an isolated repo, so the K5 proof installed a mutated `dev.honker:honker:0.5.0` jar there once and a clean one after; the last install was the clean one. Local JDK was openjdk@25 vs CI's temurin 17 (`maven.compiler.release=17`, so bytecode target matched, running JVM did not). Worktree left clean.

russellromney:
Second adversarial round (I did not write this code, and I did not do the first review). Everything below is marked **RUN** or **READ**. Ran in a detached worktree at `3cb1403`; two commits pushed to this branch.

The first round's five mutations really were all caught. So I looked for what has *no* test rather than what is wrong, and ran the suite the way CI will. That turned up one genuine defect, one silent skip, and four uncovered public functions.

## High — `Queue.asFlow` / `Listener.asFlow` turn a producer failure into a permanent hang

**RUN.** First run of `mvn -B test` on this branch hung for **over ten minutes** on `listenerAndQueueFlowWrappersUseJavaRuntime`. `jstack` on the surefire fork:

```
"main" ... TIMED_WAITING (parking)
  at kotlinx.coroutines.BlockingCoroutine.joinBlocking
  at dev.honker.kotlin.HonkerKotlinTest.listenerAndQueueFlowWrappersUseJavaRuntime(HonkerKotlinTest.kt:48)
```

No `honker-kotlin-queue-flow` thread in the dump. stderr had:

```
Exception in thread "honker-kotlin-queue-flow" dev.honker.HonkerException: job row from Honker is missing required field state
  at dev.honker.Queue.claimBatch(Queue.java:66)
  at dev.honker.kotlin.HonkerKotlinKt$asFlow$3.invokeSuspend$lambda$0(HonkerKotlin.kt:104)
```

The *cause* was mine — a stale `libhonker_ext` without #124's widened `honker_claim_batch` RETURNING. With a matching extension the suite is green in 0.65 s. **RUN**, so that specific error will not happen in CI.

The *failure mode* is the defect, and it is pre-existing in `HonkerKotlin.kt`: the producer runs on a bare `Thread` that catches only `HonkerClosedException` and never closes the channel. Any other throwable kills the thread, prints a trace, and parks every collector on that flow **forever**. The clean path has the same hole — on `HonkerClosedException` the thread exits and the collector still waits. This is the "fail loudly" rule inverted: the error is swallowed into an infinite wait.

It matters *for this PR* because this PR is what puts these flows under CI for the first time. A hang here burns the `jvm` job's whole `timeout-minutes: 20` — taking the ORM proofs with it — and reports no test failure to read.

**Fixed** in 39f5a1d: `close()` when the loop ends or the database goes away, `close(failure)` on anything else.

Load-bearing, **RUN**: `queueFlowEndsWhenTheDatabaseClosesInsteadOfParkingTheCollector` opens a db, starts collecting, closes the db.

- old catch/finally restored → `FAILURE! Time elapsed: 15.81 s ... closing the database must end the queue flow ==> expected: not <null>`
- fix restored → passes in under a second.

Also added a class-level `@Timeout(60, threadMode = SEPARATE_THREAD)`. `SEPARATE_THREAD` is deliberate — the default mode does not preempt, so it cannot catch a hang. Suite is 0.65–4.0 s across four runs, so 60 s is ~15x headroom on a loaded box.

## High — a test in this file can be silently not run

**RUN.** I wrote the flow test as `= runBlocking { ... }` ending in `assertNotNull(...)`. That makes the method return `Result<Unit>`, JUnit never discovers it, and **surefire reported `Tests run: 13` with BUILD SUCCESS** — no skip, no warning. I only caught it because I counted `@Test` and got 14.

Every existing test here happens to end in a `Unit`-returning assertion, so they all run today. But this is a live trap in a file where every coroutine test is expression-bodied. Mine is now `runBlocking<Unit>` with a comment saying why. Worth promoting to the repo CLAUDE.md: **an expression-bodied `= runBlocking { … }` test whose last expression is not `Unit` compiles to a non-void method and is silently never run.**

## Medium — four public functions had zero coverage

**RUN.** `Queue.jobDetails(id, codec)`, `Database.jobDetails(id, codec)`, `Job.details(codec)` and `JobSnapshot.details(codec)` were untested. Deleting all four left the suite green. Concretely: mutating `JobSnapshot.details(codec)` to `codec.decode(state())` — the shared path under all four — **passed all 13 pre-existing tests**. That is the mutation the first round could not have found, because it was aimed at covered code.

**Fixed** in 8b7cc8b: `codecOverloadsDecodeOnEveryReceiver` exercises all four plus the scoping of the codec overload. With the mutation in place: `expected: <alice@example.com> but was: <pending>`, failing that test and only that test.

## Medium — nullability was never proven to round-trip

**RUN.** Pending and processing were covered; retry was not. Retry is the only path where `workerId` and `claimExpiresAt` go from set *back to* null on the same row (core `honker_ops.rs:925`: `state='pending', worker_id=NULL, claim_expires_at=NULL`, attempts kept) — which is exactly what the Kotlin nullable types promise.

**Fixed**: `retriedJobDetailsDropBackToPendingWithNoWorker` asserts state, both nulls, `attempts == 1`, stable `createdAt`, and `runAt` inside `[retriedBefore+45, retriedAfter+45]`. Proof it bites: `claimExpiresAt() ?: 0L` → `a retried job releases its claim deadline ==> expected: <null> but was: <0>`.

Also added `deadLetteredJobDetailsAreGoneFromBothLookups` — the KDoc on `Queue.jobDetails` claims a dead-lettered job returns null, and only the ack'd / unknown-id / wrong-queue cases were tested. Verified by running, not by mutation: breaking it would require changing core or JVM, which is out of bounds here.

Suite is now **14/14, stable over four runs**, 0.65–4.0 s.

## The five lenses

**1. Does it make sense to a Kotlin user? Yes.** **READ.** `details()` on the four job-shaped types and `jobDetails(id)` on the three container types is 11 functions, but the rule behind it is one line: *typed receiver, no codec; untyped receiver, optional codec.* `TypedQueue`/`TypedJob`/`TypedJobSnapshot` correctly have no codec overload. `details()` vs `jobDetails()` is right, not inconsistent — `queue.details()` would read as the queue's own details. Discoverable by autocomplete on the receiver you already hold.

One judgment call I did **not** change, because it is a semantics decision: Kotlin convention for "returns null instead of throwing/Optional" is the `OrNull` suffix (`getOrNull`, `firstOrNull`). `jobDetails(id)` does not signal nullability by name; only the `JobDetails?` return type does. `getJobOrNull(id)` would read more like Kotlin and pair better with the JVM `getJob` it wraps. Your call — I am not renaming a public API in review.

`JobDetails<String>` from the untyped path has `payload == payloadJson`, two identical strings. Redundant but documented and it keeps one uniform shape. Fine.

**2. The naming decision holds.** **READ.** `dev.honker.JobSnapshot` exists; two star-imports supplying the same simple name is an ambiguity error at every use site in Kotlin. Stronger than the PR body says: `Jobs.kt` itself needs `dev.honker.JobSnapshot` as the receiver of `toDetails`, so a same-named class would need an alias in the wrapper's own source. And `scripts/test-jvm-consumer.sh:154-157` shows real consumers importing from both packages. `JobDetails` also pairs with the `details()` / `jobDetails()` method names. `JobRow` would match Go, but I would not churn the API for that. Keep it.

**3. The CI gate should be lifted out — reporting, not doing.** **READ.** Agreed, and more strongly than "nice to have": the Kotlin binding is ungated until #124 → #141 → #144 all land, and the two defects above show what accumulates in the meantime — a ten-minute hang and a silent skip, both in code that predates every PR in this stack. The `Run honker-kotlin tests` step depends on nothing in this PR beyond the `dev.honker:honker` install the `jvm` job already does. It would apply to `main` on its own. Not doing it, per instructions.

**4. Completeness and nullability: correct.** **READ.** All twelve fields map 1:1 in `Jobs.kt:45-59` with no crossings; field order matches the Java record. `workerId: String?`, `claimExpiresAt: Long?`, `expiresAt: Long?`, no `?: 0` or `?: ""` anywhere. Claimed and snapshot paths share `toDetails`, so they cannot drift. `Queue.jobDetails` scoped, `Database.jobDetails` global, README:77-80 says so.

**5. CI wiring is honest.** **READ.** `mvn -B test`, no exclusions, all 14 gate; `HONKER_EXTENSION_PATH` points at the release `.so` the job already built; runs after the `dev.honker:honker` install. The dependabot comment flip is accurate.

## Left alone, on purpose

- **`fun <T> Queue.typed(codec)` (HonkerKotlin.kt:166) is dead code.** `Queue.java:160` has a `typed(JsonCodec<T>)` member, and Kotlin members always beat extensions — at the call site *and* inside the extension body. So it is unreachable, not infinite recursion. Pre-existing and unrelated to job parity; flagging, not fixing.
- **`Job.claimExpiresAt()` / `TypedJob.claimExpiresAt()` return primitive `long` while unboxing a nullable `Long`** (`Job.java:70`). Not reachable today — a `Job` only ever comes from a claim, and a claimed row always has the column. Belongs to #141's owner, not this branch. Kotlin is unaffected: `details()` reads the boxed `JobSnapshot.claimExpiresAt()`.
- **`scripts/test-jvm-consumer.sh` is not run by any CI job** (grep of `.github/workflows/` finds nothing), and its Kotlin consumer test does not touch `jobDetails`. Pre-existing gap, separate concern.
- No payload-encoding change (#146), no `claimed_at` (#135), no scoping beyond #141 (#134). No change to `honker-core`, `packages/honker-jvm`, or any other binding.

## Verdict

**Merge**, after #124 and #141. The `JobDetails` design and the twelve-field mapping were already right; what this round added is coverage for the four overloads nobody had exercised, the retry nullability round-trip, and a fix for a hang that would otherwise have been this binding's first CI incident.

### Environment

macOS, local JDK 26 (CI is temurin 17; `maven.compiler.release=17`, so bytecode target matched, running JVM did not). Shared `~/.m2` in offline mode with the `dev.honker:honker:0.5.0` jar the previous reviewer left installed. Extension: I could not use the main checkout's `target/release` build — it is from another branch and lacks the widened claim — so I copied the release `.dylib` from #141's worktree, whose `honker-core` and `honker-extension` are byte-identical to this branch's (`git diff --stat HEAD origin/feat/jvm-kotlin-job-parity -- honker-core honker-extension` is empty). No Rust build. Mutations were restored from scratchpad copies, never with `git checkout --`; `git diff` clean after each.

PR #145, opened 2026-09-01
feat(bun): typed job payloads and full job details (#136)

russellromney:
Part of #136 — the Bun half. The .NET half is #142.

**Stacked on #124 (`feat/node-typed-jobs`), which is the base of this PR. Merge after #124.**
It depends on the widened `honker_claim_batch` RETURNING clause that only exists there. No honker-core or SQL ABI change here; core already returns everything needed.

## What changed

- Claimed jobs and read-only snapshots carry all twelve fields: `id`, `queue`, `payload`, `state`, `priority`, `runAt`, `workerId`, `claimExpiresAt`, `attempts`, `maxAttempts`, `createdAt`, `expiresAt`. camelCase, matching Node; the core JSON stays snake_case.
- `Queue`, `Job`, `JobSnapshot`, `ClaimWaker`, and `Outbox` take a payload generic defaulting to `JsonValue` — the same shape `packages/honker-node/wrapper.d.ts` established, deliberately not a new invention.
- **Behavior change:** `getJob()` returns the camelCase `JobSnapshot` instead of the raw snake_case `JobRow`. Field *names* changed; the payload *encoding* did not.

### Snapshot payload encoding (flagged for a repo-wide decision)

Bindings disagree and this PR does not try to settle it:

| binding | claimed `Job` payload | snapshot payload |
| --- | --- | --- |
| Node (#124) | decoded | decoded |
| **Bun (this PR)** | decoded (unchanged) | **raw JSON text (unchanged)** |
| **.NET (#142)** | `JsonElement` / `GetPayload<T>()` (unchanged) | `JobRow.Payload` raw JSON text (unchanged); the new opt-in `JobSnapshot<T>` carries both `Payload` (decoded) and `PayloadRaw` |
| Go, Python | raw JSON text | raw JSON text |

An earlier revision of this PR decoded Bun's snapshot payload to match Node. That was reverted — it is a breaking change #136 did not ask for. `JobSnapshot` is therefore not generic over the payload; `Queue<TPayload>` still types enqueue and claims.
- A claimed `Job` keeps `ack`/`retry`/`fail`/`heartbeat`; `JobSnapshot` is data only.
- A claimed row arriving without `worker_id`/`claim_expires_at` throws instead of being silently accepted.
- Typing is **compile-time only**. No runtime payload validation was added, and the README says so explicitly.

### One prerequisite fix, in its own commit

`fix(bun): make the binding compile under strict tsc` — the package ships its `.ts` source as `types`, and that source did not type-check: `ReturnType<typeof dlopen>` erased the FFI symbol map (every symbol `never`) and `Transaction.query` bound params as `unknown[]` where `bun:sqlite` wants `SQLQueryBindings[]`. Type-only changes, runtime identical. Without them there is no way to gate the new compile-time typing, and strict TS consumers of this package could not compile it at all. Flagging it because it is adjacent to, not part of, #136.

## Tests

New `packages/honker-bun/test/typed_jobs.test.ts`, 6 tests, asserting values rather than presence (times bounded by `unixepoch()` reads taken around each call):

- `a claimed job carries every field the core returns`
- `a pending snapshot carries every field the core returns`
- `a reader sees the processing snapshot, then nothing after ack` (two handles on one db file)
- `a delayed job reports its runAt and is not claimable yet`
- `a retried job reports the new attempt count and run_at`
- `the claim waker returns a job carrying the same detail`

Compile-time gate: `packages/honker-bun/test/types/typed-jobs.ts` + `tsconfig.json`, run by `bun run test:types` (new script) and wired into the CI Bun step. It includes a `@ts-expect-error` on an enqueue that violates the payload contract, so the gate fails if the generics go loose.

Local: `bun test` → 64 pass, 0 fail (whole suite; kernel/shm backends skip on a locally built extension without those features). `bun run test:types` → exit 0.

### Falsification

Broke `this.maxAttempts = row.max_attempts` to `row.attempts` in the `Job` constructor:

```
95 |       expect(job!.maxAttempts).toBe(5);
                                    ^
error: expect(received).toBe(expected)
Expected: 5
Received: 1
      at test/typed_jobs.test.ts:95:32
(fail) honker-bun job details > a claimed job carries every field the core returns
```

(`a retried job reports the new attempt count and run_at` failed too: expected 3, received 1.) Restored, suite green again.

## Not in scope

- `claimed_at` (#135) — deliberately not added.
- Queue-scoping of `getJob` (#134) — Bun's `getJob` is still a global id lookup, unlike Node's now-scoped one. Worth noting since the payload generic is only sound for ids the queue owns.
- The known `ClaimWaker` bug in `src/index.ts` (it parks on `idlePollS` after a passed `runAt` deadline, the Bun twin of the Node fix in #122) is untouched. The new waker test claims an immediately-ready job and does not exercise that path.
- No changes to honker-core, the SQL ABI, honker-node, or any other binding.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -576,6 +576,7 @@ jobs:
           HONKER_INTEROP_PYTHON: ${{ github.workspace }}/.venv/bin/python
         run: |
           bun install --frozen-lockfile
+          bun run test:types
           bun test
       - name: Ruby
         working-directory: packages/honker-ruby
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,10 +31,13 @@
   `expiresAt` join the id, queue, payload, worker, and attempts fields
   they already had.
 - **Behavior change:** `Queue.getJob()` returns the camelCase
-  `JobSnapshot` shape with a parsed payload, instead of the SQL ABI's
-  raw snake_case row with a JSON string payload. The old exported
-  `JobRow` interface is replaced by `JobSnapshot<TPayload>`. Same change
-  Node made in the typed-jobs work.
+  `JobSnapshot` shape instead of the SQL ABI's raw snake_case row. The
+  exported `JobRow` interface is replaced by `JobSnapshot`. The
+  snapshot's `payload` stays raw JSON text, exactly as this binding has
+  always returned it; only the field names changed. A claimed
+  `Job.payload` stays decoded, as it always was. Bindings disagree on
+  snapshot payload encoding (Node decodes; Bun, Go, and Python return
+  raw text) and one convention has not been chosen yet.
 - A claimed row without `worker_id` or `claim_expires_at` now throws
   instead of being silently accepted.
 - Payload typing is compile-time only; honker adds no runtime payload
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,31 @@
 - Regression test claims a job at a deadline with one claim forced to
   return empty: it fails on the unpatched `api.js` and passes on the fix.
 
+## Unreleased — typed Bun jobs
+
+- Bun `Queue`, `Job`, `JobSnapshot`, `ClaimWaker`, and `Outbox` now take
+  a payload generic defaulting to `JsonValue`, matching the Node
+  binding's shape: `db.queue<EmailPayload>("emails")` types enqueue,
+  claims, and snapshots.
+- Claimed jobs carry every field the core returns. `state`, `priority`,
+  `runAt`, `claimExpiresAt`, `maxAttempts`, `createdAt`, and
+  `expiresAt` join the id, queue, payload, worker, and attempts fields
+  they already had.
+- **Behavior change:** `Queue.getJob()` returns the camelCase
+  `JobSnapshot` shape with a parsed payload, instead of the SQL ABI's
+  raw snake_case row with a JSON string payload. The old exported
+  `JobRow` interface is replaced by `JobSnapshot<TPayload>`. Same change
+  Node made in the typed-jobs work.
+- A claimed row without `worker_id` or `claim_expires_at` now throws
+  instead of being silently accepted.
+- Payload typing is compile-time only; honker adds no runtime payload
+  validation. `bun run test:types` type-checks the binding and a typed
+  usage fixture, and CI runs it.
+- The binding's own sources now type-check under strict `tsc`. The
+  `bun:ffi` watcher handle and one `bun:sqlite` query binding were
+  previously untypeable (`never` symbols), which broke any strict
+  TypeScript consumer of this package.
+
 ## Unreleased — typed Node jobs
 
 - Node `Queue`, `Job`, `JobSnapshot`, `ClaimWaker`, and `Outbox` APIs now carry
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,11 @@
 - A claimed row without `worker_id` or `claim_expires_at` now throws
   instead of being silently accepted, and a test builds both a complete
   and a truncated row to prove the guard fires only on the bad one.
+- `Queue.getJob()` is documented as NOT queue-scoped: job ids are
+  globally unique, so it returns a row from any queue. The Node binding
+  scopes its lookup; #134 tracks making the bindings agree. Behavior is
+  unchanged here, only the README and the JSDoc, which previously said
+  nothing about it.
 - A job enqueued without `expires` now has its `expiresAt` asserted to be
   `null` on both the snapshot and the claimed job. Nothing pinned it
   before, so an `expires_at ?? 0` fallback would have handed callers 0 —
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@
   raw text) and one convention has not been chosen yet.
 - A claimed row without `worker_id` or `claim_expires_at` now throws
   instead of being silently accepted.
+- A job enqueued without `expires` now has its `expiresAt` asserted to be
+  `null` on both the snapshot and the claimed job. Nothing pinned it
+  before, so an `expires_at ?? 0` fallback would have handed callers 0 —
+  a valid unix timestamp meaning 1970 — with the suite still green.
 - Payload typing is compile-time only; honker adds no runtime payload
   validation. `bun run test:types` type-checks the binding and a typed
   usage fixture, and CI runs it.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,14 +39,18 @@
   snapshot payload encoding (Node decodes; Bun, Go, and Python return
   raw text) and one convention has not been chosen yet.
 - A claimed row without `worker_id` or `claim_expires_at` now throws
-  instead of being silently accepted.
+  instead of being silently accepted, and a test builds both a complete
+  and a truncated row to prove the guard fires only on the bad one.
 - A job enqueued without `expires` now has its `expiresAt` asserted to be
   `null` on both the snapshot and the claimed job. Nothing pinned it
   before, so an `expires_at ?? 0` fallback would have handed callers 0 —
   a valid unix timestamp meaning 1970 — with the suite still green.
 - Payload typing is compile-time only; honker adds no runtime payload
   validation. `bun run test:types` type-checks the binding and a typed
-  usage fixture, and CI runs it.
+  usage fixture, and CI runs it. The fixture carries two negative checks
+  (`@ts-expect-error`) so the gate fails if the generics go loose in
+  either direction — a missing required field, or `payload` decaying to
+  `any`, which every positive assertion would otherwise accept.
 - The binding's own sources now type-check under strict `tsc`. The
   `bun:ffi` watcher handle and one `bun:sqlite` query binding were
   previously untypeable (`never` symbols), which broke any strict
--- a/packages/honker-bun/README.md
+++ b/packages/honker-bun/README.md
@@ -57,9 +57,11 @@ interface EmailPayload {
 const emails = db.queue<EmailPayload>("emails");
 const id = emails.enqueue({ to: "alice@example.com", template: "welcome" });
 
-// Read-only snapshot: data, no claim methods.
+// Read-only snapshot: data, no claim methods. Its payload is raw JSON
+// text — parse it yourself.
 const pending = emails.getJob(id);
 console.log(pending?.state, pending?.priority, pending?.runAt);
+const payload = pending ? (JSON.parse(pending.payload) as EmailPayload) : null;
 
 const job = emails.claimOne("worker-1");
 if (job) {
@@ -77,6 +79,13 @@ Claimed jobs and snapshots carry every field the core returns: `id`,
 `claimExpiresAt`, `attempts`, `maxAttempts`, `createdAt`, `expiresAt`.
 A claimed `Job` also has `ack`, `retry`, `fail`, and `heartbeat`; a
 `JobSnapshot` is data only, because reading a row does not claim it.
+
+Payload encoding differs between the two, and that is deliberate: a
+claimed `job.payload` is decoded (it always was), while a snapshot's
+`payload` stays raw JSON text (it always was). Bindings currently
+disagree here — Node decodes snapshot payloads, Bun, Go, and Python do
+not — and settling on one convention is a separate decision, not part of
+the job-detail work.
 
 Recurring schedules use `schedule`:
 
--- a/packages/honker-bun/README.md
+++ b/packages/honker-bun/README.md
@@ -45,6 +45,39 @@ Delayed jobs use `runAt`:
 q.enqueue({ to: "later@example.com" }, { runAt: Math.floor(Date.now() / 1000) + 10 });
 ```
 
+Give a queue a payload contract and claimed jobs plus `getJob()`
+snapshots keep it:
+
+```ts
+interface EmailPayload {
+  to: string;
+  template: "welcome" | "receipt";
+}
+
+const emails = db.queue<EmailPayload>("emails");
+const id = emails.enqueue({ to: "alice@example.com", template: "welcome" });
+
+// Read-only snapshot: data, no claim methods.
+const pending = emails.getJob(id);
+console.log(pending?.state, pending?.priority, pending?.runAt);
+
+const job = emails.claimOne("worker-1");
+if (job) {
+  console.log(job.payload.template, `${job.attempts}/${job.maxAttempts}`);
+  job.ack();
+}
+```
+
+Payload generics describe the expected JSON shape at compile time.
+Honker never validates payload shape — not in this binding, not in the
+database — so every process writing to a queue has to agree on the type.
+
+Claimed jobs and snapshots carry every field the core returns: `id`,
+`queue`, `payload`, `state`, `priority`, `runAt`, `workerId`,
+`claimExpiresAt`, `attempts`, `maxAttempts`, `createdAt`, `expiresAt`.
+A claimed `Job` also has `ack`, `retry`, `fail`, and `heartbeat`; a
+`JobSnapshot` is data only, because reading a row does not claim it.
+
 Recurring schedules use `schedule`:
 
 ```ts
--- a/packages/honker-bun/README.md
+++ b/packages/honker-bun/README.md
@@ -45,8 +45,9 @@ Delayed jobs use `runAt`:
 q.enqueue({ to: "later@example.com" }, { runAt: Math.floor(Date.now() / 1000) + 10 });
 ```
 
-Give a queue a payload contract and claimed jobs plus `getJob()`
-snapshots keep it:
+Give a queue a payload contract and claimed jobs keep it. A `getJob()`
+snapshot carries every field but hands the payload back as raw JSON
+text, so you parse it yourself:
 
 ```ts
 interface EmailPayload {
@@ -79,6 +80,11 @@ Claimed jobs and snapshots carry every field the core returns: `id`,
 `claimExpiresAt`, `attempts`, `maxAttempts`, `createdAt`, `expiresAt`.
 A claimed `Job` also has `ack`, `retry`, `fail`, and `heartbeat`; a
 `JobSnapshot` is data only, because reading a row does not claim it.
+
+`getJob()` looks a job up by id across every queue, not just the one you
+called it on — job ids are globally unique. Check `snapshot.queue` if
+that matters to you. (The Node binding scopes its `getJob`; #134 tracks
+making the rest of the bindings agree.)
 
 Payload encoding differs between the two, and that is deliberate: a
 claimed `job.payload` is decoded (it always was), while a snapshot's
--- a/packages/honker-bun/bun.lock
+++ b/packages/honker-bun/bun.lock
@@ -4,6 +4,10 @@
   "workspaces": {
     "": {
       "name": "@russellthehippo/honker-bun",
+      "devDependencies": {
+        "@types/bun": "^1.4.0",
+        "typescript": "^5.9.3",
+      },
       "optionalDependencies": {
         "@russellthehippo/honker-ext-darwin-arm64": "0.4.6",
         "@russellthehippo/honker-ext-darwin-x64": "0.4.6",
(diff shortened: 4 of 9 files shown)


russellromney:
Adversarial review (I did not write this code). Everything below was RUN unless marked INFERRED, in a detached worktree at `a7f5e13` with `HONKER_EXT_PATH` pointed at a prebuilt extension.

Baseline: `bun test test/typed_jobs.test.ts` → `6 pass / 0 fail / 62 expect() calls, [91ms]`. Full suite `64 pass / 0 fail / 200 expect() calls` across 6 files. The suite does `describe.skip` when the extension is missing, but `test/typed_jobs.test.ts:31-33` throws when `process.env.CI` is set and `ci.yml:575` sets `HONKER_EXT_PATH`, so a silent green in CI is not possible.

**This is the best-mutated test code in the #136 series.** I ran 12 recorded mutations against `src/index.ts`; 11 were caught. The trap that bit Go and .NET — `run_at == created_at` on an immediately-enqueued job — does not apply here, because the fixtures use deliberately distinct values (`runAt = now-5` vs `createdAt ≈ now`, `attempts=1` vs `maxAttempts=5`, `claimExpiresAt ≈ now+120` vs `expiresAt ≈ now+600`, `priority=7`).

| Mutation | Result |
|---|---|
| snapshot: swap `runAt` ↔ `createdAt` | caught (3 pass / 3 fail) |
| snapshot: swap `attempts` ↔ `maxAttempts` | caught (3/3) |
| snapshot: swap `claimExpiresAt` ↔ `expiresAt` | caught (4/2) |
| Job: swap `runAt` ↔ `createdAt` | caught (5/1) |
| Job: swap `attempts` ↔ `maxAttempts` | caught (4/2) |
| Job: swap `claimExpiresAt` ↔ `expiresAt` | caught (5/1) |
| Job / snapshot: `priority` ← `row.id` | caught (4/2, 5/1) |
| snapshot: `state` passthrough → hardcoded | caught (5/1) |
| snapshot: `workerId ?? ""` instead of `?? null` | caught (4/2) |
| snapshot: `claimExpiresAt ?? 0` instead of `?? null` | caught (5/1) |
| **snapshot: `expiresAt ?? 0` instead of `?? null`** | **SURVIVED (6 pass / 0 fail)** |

### 1. MEDIUM — Bun still carries the ClaimWaker bug #122 fixed in Node. Pre-existing; confirmed untouched and still present. RAN (diff) / INFERRED (timing).

`src/index.ts:951-958`:
```ts
const nextClaimAt = this.queue.nextClaimAt();
let waitMs = this.idlePollMs;
if (nextClaimAt && nextClaimAt > 0) {
  const untilNext = Math.max(0, nextClaimAt * 1000 - Date.now());
  waitMs = waitMs == null ? untilNext : Math.min(waitMs, untilNext);
}
await waitForUpdateOrTimeout(this.updates, opts.signal, waitMs);
```
Node's fixed version (`packages/honker-node/api.js:317-331`) carries a `deadlinePassed` flag and an `else if (deadlinePassed) waitMs = min(waitMs, DEADLINE_RETRY_MS)` branch. Bun has neither the flag nor the constant. The core-side premise holds: `queue_next_claim_at` (`honker_ops.rs:763-784`) filters on `run_at > unixepoch()`, so it returns 0 the instant a deadline arrives.

Failure: the waker parks on a delayed job's `run_at`, wakes at the deadline, `claimOne()` comes back empty (writer lock held past `busy_timeout`, or a racing worker took it while other work is ready), `nextClaimAt()` now returns 0, and `waitMs` falls back to the full `idlePollS` — default 5 s of a due job sitting unclaimed.

**This PR did not touch that path.** `git diff 71007a4..HEAD -- src/index.ts` shows the only edits inside `ClaimWaker` are the generic parameter and return types; the loop body is byte-identical. Not lost, but not fixed — worth a follow-up issue mirroring #122.

### 2. LOW/MEDIUM — `expiresAt === null` on a job with no expiry is never asserted. RAN, proven by a surviving mutation.

```
-    expiresAt: row.expires_at ?? null,   // toSnapshot, line 239
+    expiresAt: row.expires_at ?? 0,
```
```
 6 pass, 0 fail, 62 expect() calls — [204ms]
```
Every test that reads `expiresAt` (tests 1 and 2) enqueues **with** `expires:`; the three that enqueue without one never assert `expiresAt` at all. `workerId === null` and `claimExpiresAt === null` on a pending snapshot *are* properly guarded — this is the one hole. A caller reading `snapshot.expiresAt` on a never-expiring job would get `0`, a valid unix timestamp meaning 1970, and no test notices. The same gap almost certainly exists on the claimed-`Job` side (`this.expiresAt = row.expires_at ?? null`, line 867) — tests 5 and 6 claim jobs with no `expires` and never assert it. INFERRED for the `Job` side; the mutation was queued when the machine ran out of disk.

One line in `test/typed_jobs.test.ts`: assert `expiresAt` is null in the delayed-job or retried-job test.

### 3. LOW — `Job.state` is a hardcoded constant, so its assertion is tautological.

`src/index.ts:859`: `this.state = "processing";` — never read from `row.state`. `expect(job!.state).toBe("processing")` cannot fail regardless of what the core returns. Correct-by-construction and the JSDoc says so, but the assertion proves nothing. The snapshot path does pass `row.state` through, and that mutation *was* caught.

### 4. LOW — `getJob` JSDoc becomes incomplete under #134.

`src/index.ts:713-718` says a miss means "ack'd, dead'd, or never existed"; the reference `wrapper.d.ts:217-225` lists a fourth case, "or belongs to a different queue".

Otherwise attack E is clean: nothing conflicts with adding scoping (`getJob` is a self-contained three-line method), **no test asserts the global behavior** (no test enqueues into queue A and reads it back through queue B; the only queue assertion is `expect(snapshot!.queue).toBe("emails")` on a job that genuinely belongs to `emails`, still true after scoping), and the README example uses an id from the same queue.

Worth recording: the type-unsoundness argument the Node wrapper gives **does not apply to Bun today**. Node returns `JobSnapshot<TPayload>`; Bun returns a non-generic `JobSnapshot` whose `payload` is a raw JSON string. A global lookup returning a foreign row cannot violate a type Bun never asserts. That changes the moment someone makes the Bun snapshot generic.

### The prerequisite type-check commit (71007a4) — all three claims verified. RAN.

**The before-state claim is true.** Built a detached worktree at `origin/feat/node-typed-jobs` (841c257), dropped in the PR's exact `compilerOptions`, ran the same strict `tsc`:
```
src/index.ts(239,7): error TS2345: Argument of type '[Buffer, Buffer, bigint, Buffer, bigint]' is not assignable to parameter of type 'never'.
src/index.ts(255,7): error TS2345: Argument of type '[unknown, bigint]' is not assignable to parameter of type 'never'.
src/index.ts(266,43): error TS2345: Argument of type '[unknown]' is not assignable to parameter of type 'never'.
src/index.ts(511,36): error TS2344: Type 'unknown[]' does not satisfy the constraint 'SQLQueryBindings | SQLQueryBindings[]'.
BASE_EXIT=2
```
Exactly the `never` FFI symbols and `unknown[]` bindings the commit message describes.

**It now type-checks clean.** `bun run test:types` → `tsc -p test/types/tsconfig.json`, exit 0.

**The fix is genuinely type-only.** I went looking specifically for runtime semantics smuggled into a "types only" commit:
- **FFI marshalling is byte-identical.** The `dlopen` symbol map moved verbatim into `openWatcherLib()`. `honker_watcher_open_v2` is still `[ptr, ptr, u64, ptr, u64] -> ptr`, `honker_watcher_wait` still `[ptr, u64] -> i32`, `honker_watcher_close` still `[ptr] -> void`. No i64/i32/ptr or cstring/ptr substitution. This was the highest-risk hiding place and it is clean.
- `Transaction.query`: `this.raw.query<T, unknown[]>` → `<T, SQLQueryBindings[]>` (type argument only) and `stmt.get(...params)` → `stmt.get(...(params as SQLQueryBindings[]))` (cast erased at runtime; spread and argument order unchanged).
- No changed defaults, no reordered arguments, no added `?.`. `dlopen` moved from the constructor body to a module-level function still called once from the constructor.

### Also checked

- `bun install --frozen-lockfile` → exit 0, no lockfile drift. `typescript@5.9.3` and `@types/bun@1.4.0` are correctly under `devDependencies`, and `bun.lock` records them there.
- `ci.yml:579` adds `bun run test:types` between `bun install --frozen-lockfile` and `bun test` — the new check is genuinely wired into CI.
- `claim_batch`'s emitted keys (`honker_ops.rs:720-733`) match `RawJobRow`'s twelve declared fields exactly, payload left as a JSON string.

Would merge after the one-line `expiresAt` null assertion. Findings 1, 3 and 4 are follow-ups, not blockers — but finding 1 should get its own issue now so it does not get lost.

russellromney:
Finding 2 fixed. CI green on `6167400` (run 33585891491).

`test(bun): assert expiresAt is null on a job with no expiry` — commit `6167400`.

Three assertions added, covering both surfaces you named — `toSnapshot` (proven) and the claimed `Job` constructor (which you had marked INFERRED):

- "a delayed job reports its runAt…" → `expect(snapshot!.expiresAt).toBeNull()`
- "a retried job reports the new attempt count and run_at" → `expect(first!.expiresAt).toBeNull()` on the claimed `Job`, and `expect(snapshot!.expiresAt).toBeNull()` on the snapshot after retry

Proof it is load-bearing. Your mutation, applied to both sites (`src/index.ts:239` and `:867`):

```
-    expiresAt: row.expires_at ?? null,
+    expiresAt: row.expires_at ?? 0,
-    this.expiresAt = row.expires_at ?? null;
+    this.expiresAt = row.expires_at ?? 0;
```

```
195 |       expect(snapshot!.expiresAt).toBeNull();
                                        ^
error: expect(received).toBeNull()
Received: 0
(fail) honker-bun job details > a delayed job reports its runAt and is not claimable yet

210 |       expect(first!.expiresAt).toBeNull();
                                     ^
error: expect(received).toBeNull()
Received: 0
(fail) honker-bun job details > a retried job reports the new attempt count and run_at

 4 pass
 2 fail
```

The `Job`-side gap you inferred was real — that mutation now fails too.

### Runs

Clean tree: `bun test test/typed_jobs.test.ts` → `6 pass / 0 fail / 65 expect() calls` (was 62). Full suite `64 pass / 0 fail / 203 expect() calls`. `bun run test:types` exit 0.

Extension verified fresh, not a stale dylib: `strings libhonker_ext.dylib | grep -c "RETURNING id, queue, payload, state"` → 1.

Findings 1 (ClaimWaker / #122), 3, and 4 left untouched as you recommended — follow-ups, not this PR. Finding 1 still wants its own issue.

russellromney:
Second adversarial round (again, I did not write this code). Everything below is marked **RAN** or **READ**. Runs used the prebuilt `libhonker_ext.dylib` via `HONKER_EXT_PATH`; freshness re-confirmed (`strings ... | grep -c "RETURNING id, queue, payload, state"` -> 1). No Rust was built.

Baseline before my changes: `bun run test:types` exit 0, `bun test` 64 pass / 0 fail / 203 expect() calls. After: **65 pass / 0 fail / 206 expect() calls**, types exit 0. Pushed as `3cfe778` and `f893dd9`.

## Lens 2 — the prerequisite type-check commit (`71007a4`). Clean. RAN + READ.

I re-derived the "type-only" claim independently rather than taking round 1's word for it.

**READ**: the whole runtime delta is hoisting the `dlopen` call into a module-level `openWatcherLib()` that the constructor calls once. The symbol map is character-identical — `honker_watcher_open_v2` still `[ptr, ptr, u64, ptr, u64] -> ptr`, `wait` `[ptr, u64] -> i32`, `close` `[ptr] -> void`. No memoization was introduced: one `dlopen` per `CoreWatcher` before and after. `Transaction.query`'s change is a type argument plus a cast that erases. Nothing else in the commit touches an expression.

**RAN**: the gate is load-bearing on exactly this. Reverting `ReturnType<typeof openWatcherLib>` back to `ReturnType<typeof dlopen>`:

```
src/index.ts(290,5): error TS2322: Type 'unknown' is not assignable to type 'bigint | Pointer | null'.
exit=2
```

So the defect that shipped in `honker-bun@0.4.2` cannot come back silently.

## Type-gate mutation battery — 13 mutations, 12 caught, 1 hole (now closed). RAN.

I mutated `src/index.ts` and re-ran `tsc -p test/types/tsconfig.json` each time.

| Mutation | Result |
|---|---|
| `JobSnapshot.workerId: string \| null` -> `string` | caught (TS2322 in `toSnapshot`) |
| `JobSnapshot.expiresAt: number \| null` -> `number` | caught (TS2322) |
| `JobSnapshot.payload: string` -> `unknown` | caught (TS2345) |
| `JobSnapshot.state: JobState` -> `string` | caught (TS1360) |
| `Job.state: "processing"` -> `string` | caught (TS1360) |
| `Job.claimExpiresAt: number` -> `number \| null` | caught (TS18047) |
| `Queue.enqueue(payload: TPayload)` -> `unknown` | caught (TS2578, unused `@ts-expect-error`) |
| `Queue.claimOne` returns `Job<JsonValue>` | caught (TS2322) |
| `Database.queue` drops the generic | caught (TS2304) |
| `ClaimWaker.next` returns `Job<JsonValue>` | caught (TS2345) |
| `Outbox` delivery payload -> `unknown` | caught (TS2345) |
| revert the FFI `ReturnType` fix | caught (TS2322) |
| **`Job<TPayload>.payload: TPayload` -> `any`** | **SURVIVED** |

Worth stressing how good the first twelve are: the nullability types are gated *structurally*, because `toSnapshot` is inside the checked source and its return type has to line up. Round 1's `expiresAt ?? 0` hole was a runtime hole, not a type hole, and the runtime assertion added in `6167400` closes it — I re-confirmed that mutation still fails.

The survivor is the classic `any` hole: every positive assertion in the fixture is `satisfies` or a property read, and `any` satisfies everything. Widening `payload` to `any` erases the entire point of the generic and the gate stayed green.

## What I changed

### 1. `test(bun)` — the claimed-row guard had no test. FIXED, proven by mutation. RAN.

The PR body and CHANGELOG both claim "a claimed row arriving without `worker_id`/`claim_expires_at` throws instead of being silently accepted". Nothing exercised it. **RAN**: deleting the guard outright left the suite green at 64 pass.

New test `a claimed row missing its claim columns throws instead of being accepted` builds one complete row and two truncated ones, so it proves the guard fires on the bad rows *and not* on the good one. It reaches the internal wire type via `ConstructorParameters<typeof Job>[1]` rather than exporting `RawJobRow`.

Proof it is load-bearing — guard deleted:

```
(fail) honker-bun job details > a claimed row missing its claim columns throws instead of being accepted
 6 pass, 1 fail
```

Note the guard is unreachable in practice today (`claim_batch`'s UPDATE always sets both columns), so this is a fail-loudly assertion about a core contract, not a live path. That is why the test constructs `Job` directly instead of trying to fake a core response.

### 2. `test(bun)` — closed the type gate's `any` hole. FIXED, proven by mutation. RAN.

One `@ts-expect-error` on an out-of-contract field access in the fixture:

```ts
// @ts-expect-error the payload contract has no `subject`. Without a
// negative check like this one, widening Job<TPayload>.payload to `any`
// erases the whole generic and every assertion above still compiles.
claimed.payload.subject;
```

Re-ran the same mutation: `tsc` exit 2, `TS2578: Unused '@ts-expect-error' directive`. 13/13 now.

### 3. `docs(bun)` — `getJob` is global here, Node's is scoped, nothing said so. FIXED (docs only). RAN.

This is the one real Bun/Node divergence nobody had written down. Node's `getJob` on the base branch does `if (row.queue !== this.name) return null;`. Bun's does not. **RAN**, two queue handles on one db:

```
a.getJob:                 {"id":1,"queue":"alpha", ...}
b.getJob (cross-queue):   {"id":1,"queue":"alpha", ...}
```

I did **not** change the behavior — scoping is #134, and dragging it in here would bury it in a parity PR. But a user reading `queue<EmailPayload>("emails").getJob(id)` has no way to know it can hand back a row from another queue. Added a JSDoc paragraph and a README paragraph naming it and pointing at #134.

Reinforcing round 1's point, because it decides how urgent #134 is for Bun: **the type-unsoundness argument does not apply to Bun today.** Node scopes to keep `JobSnapshot<TPayload>` honest. Bun's `JobSnapshot` is non-generic with a raw-string payload, so a foreign row cannot violate a type Bun never asserts. It becomes urgent the moment #146 makes Bun's snapshot generic.

### 4. `docs(bun)` — the README overclaimed. FIXED. READ.

> Give a queue a payload contract and claimed jobs plus `getJob()` snapshots keep it

Snapshots do not keep it — `payload` is `string` — which the same README says correctly two paragraphs later. Reworded the lead so it does not contradict itself.

## Lens 1 — Bun vs Node public surface, every difference accounted for. READ.

| Member | Node | Bun | Verdict |
|---|---|---|---|
| `JobSnapshot<T>` payload | generic, decoded | non-generic, raw JSON text | **deliberate, defensible** — see below |
| `Job.claimExpiresAt` | `number \| null` | `number` | Bun is *better*: the constructor guard makes non-null real, and Node's own type is looser than its runtime |
| `Job.state` | `row.state` passthrough | hardcoded `"processing"` | same declared type, correct-by-construction either way |
| `Queue.getJob` scoping | scoped | global | #134; now documented (item 3) |
| `Job.retry(delayS=60, error='')` / `fail(error='')` | defaults | **both args required** | real divergence, left alone — see below |
| `Queue.claim()` async iterator | present | absent | pre-existing gap, not #136 |
| `Queue.enqueueTx` | present | absent (`opts.tx`) | pre-existing |
| `ClaimWaker.tryNext`, `claimWaker({pollMs})` | absent | present | pre-existing Bun extras |
| `ClaimWaker` deadline retry (#122) | fixed | **not fixed** | see item below |
| `JsonValue`, `JsonPrimitive`, `JobState`, generics on `Queue`/`Job`/`ClaimWaker`/`Outbox`/`Database.queue`/`Database.outbox` | — | — | identical |

**Is the raw-snapshot divergence defensible?** Yes, on the strict reading: the type does not lie. `JobSnapshot.payload: string` is exactly what you get, the JSDoc says so, the README says so, and the type gate proves the fixture has to `JSON.parse` it. It is *awkward* — `db.queue<EmailPayload>` types your claims but not your snapshots, so the generic is half-applied — but that awkwardness is honest and reversible, whereas silently decoding would be a breaking change #136 did not ask for. Right call for this PR; #146 should settle it, and when it does, Bun's snapshot goes generic and #134 stops being cosmetic for Bun.

**`retry`/`fail` defaults**: `job.retry()` compiles on Node and does not on Bun. Adding defaults is additive and non-breaking, but "what is the default retry delay" is a semantics call, not a review patch — leaving it to you rather than inventing 60s in a review round.

## Left alone, deliberately

- **`ClaimWaker` deadline parking (#122's bug) is still present in Bun. CONFIRMED UNTOUCHED. READ.** `git diff 71007a4..HEAD -- src/index.ts` shows the only edits inside `ClaimWaker` are the generic parameter and the return types; `next()`'s loop body is byte-identical and still has no `deadlinePassed` flag and no `DEADLINE_RETRY_MS`. It ships in `honker-bun@0.4.2`. **This needs its own issue** — fixing it here would bury a real availability bug (a due job can sit unclaimed for a full `idlePollS`, default 5s) inside a parity PR. Flagging again so it does not get lost a second time.
- **Queue scoping (#134), snapshot payload encoding (#146), `claimed_at` (#135)** — out of scope by instruction, and each would change semantics.
- **`retry`/`fail` argument defaults** — new semantics, yours to decide.
- **Packaging limits of shipping `.ts` as `types`.** The gate's tsconfig uses `strict` + `moduleResolution: "bundler"`. A consumer on `node16`/`nodenext` still cannot resolve `./extension` (extensionless relative import into a `.ts` file), and a consumer with `exactOptionalPropertyTypes` or `noUncheckedIndexedAccess` is not covered. `71007a4` genuinely fixes what it claims — the binding compiles under strict `tsc` — it just does not make every strict configuration work. Not this PR's job; worth knowing before the next release note says "type-checks clean".
- **No test asserts the current global `getJob` behavior.** Round 1 flagged that absence as a feature and I agree: pinning it would make #134 harder to land. The new docs say what happens; no test cements it.
- **Version bump.** `packages/honker-bun/package.json` is untouched, matching this repo's habit of bumping in a separate release commit. Just note that `getJob`'s shape change and the removal of the exported `JobRow` make the next Bun release a minor, not a patch.

## Verdict

**Merge** (after #124, as the PR says). The two defects I found are fixed and both fixes are proven by mutation, not by assertion. The type gate is now the strongest one in this series — 13/13 — and it genuinely guards a defect that is live in the published package. The remaining items are follow-ups with issues attached, except the `ClaimWaker` bug, which still has no issue.

russellromney:
CI green on `f893dd9` — run [33612364538](https://github.com/russellromney/honker/actions/runs/33612364538), all checks pass including `binding smoke · ubuntu aggregate` (which runs `bun run test:types` then `bun test`).

PR #148, opened 2026-09-01
feat(ruby): carry full job details on claims and snapshots

russellromney:
Ruby half of #136. Stacked on #124 (`feat/node-typed-jobs`) and targets that branch: the widened `honker_claim_batch` RETURNING clause that exposes the full job snapshot only exists there. **Merge after #124.**

## What changed

- `Honker::Job` gains readers for `state`, `priority`, `run_at`, `claim_expires_at`, `max_attempts`, `created_at`, and `expires_at`, joining the existing `id`, `queue_name`, `payload`, `worker_id`, and `attempts` — the twelve fields the claim ABI already returned and the binding threw away.
- `Queue#get_job` returns a `Honker::JobSnapshot` (a `Struct`, matching `Honker::Notification` in the same file) instead of a bare `Hash`. Same twelve fields, data only: no ack/retry/fail/heartbeat, because the reader holds no claim.
- README documents the fields and the payload-shape contract. No runtime payload validation was added — Ruby has no static payload typing and Honker never checks payload shape.
- No change to honker-core or the SQL ABI.

### Behavior change

`get_job` used to return a `Hash` of the raw ABI row. `snapshot["state"]` still works (`Struct#[]` accepts member names, so the existing `phase_mantle_spec` assertions pass unchanged), but `Hash` methods no longer do.

### Snapshot payload encoding — needs one deliberate decision later

Read-only snapshots disagree across bindings today:

| binding | `get_job`/`getJob` payload |
| --- | --- |
| Ruby (this PR) | raw JSON **text** — unchanged from the old Hash |
| Elixir (sibling PR) | raw JSON **text** — unchanged |
| Python | raw JSON text |
| Go (`JobRow.Payload`) | raw JSON text |
| Node (#124) | **decoded** value |

This PR deliberately preserves Ruby's existing encoding rather than silently picking a side; changing it is a breaking API change #136 did not ask for. Claimed `Job#payload` is decoded in every binding, so Ruby is internally split between `Job#payload` (decoded) and `JobSnapshot#payload` (text) — documented in both the code and the README. Worth one repo-wide decision.

## Tests

New `packages/honker-ruby/spec/job_fields_spec.rb`, wired into the CI Ruby step. Every field is asserted against the value it was enqueued or claimed with, not for presence. Constants are deliberately distinct (priority 7, max_attempts 5, visibility 11s, delay 300s, expiry 900s) so no assertion can pass by two values coinciding, and timestamps are checked as differences (`created_at + 900 == expires_at`) or against a `unixepoch()` read taken around the call, never against a same-second sibling.

- `test_pending_snapshot_carries_every_field`
- `test_snapshot_without_expiry_has_nil_expires_at`
- `test_delayed_job_reports_its_run_at`
- `test_claimed_job_carries_every_field`
- `test_processing_snapshot_matches_the_claim_then_misses_after_ack`
- `test_retry_bumps_attempts_and_pushes_run_at_out`

Local: 6 runs, 52 assertions, 0 failures. Full Ruby suite green (honker 26, smoke 9, parity 25, phase_mantle 8, extension_resolution 8, setup_helpers 10, railtie 3).

### The tests were proven load-bearing

Mutating `Job#initialize` to read `@priority = row["max_attempts"]`:

```
1) Failure:
HonkerJobFieldsTest#test_claimed_job_carries_every_field [spec/job_fields_spec.rb:123]:
Expected: 7
  Actual: 5
```

Mutating `JobSnapshot.from_row` to pass `row["created_at"]` for `expires_at`:

```
2) Failure:
HonkerJobFieldsTest#test_pending_snapshot_carries_every_field [spec/job_fields_spec.rb:89]:
Expected |1788256899 - 1788255999| (900) to be <= 1.

3) Failure:
HonkerJobFieldsTest#test_snapshot_without_expiry_has_nil_expires_at [spec/job_fields_spec.rb:94]:
expires_at must be nil, not 0.
Expected 1788255999 to be nil.
```

Both mutations restored; suite green again.

## Not in this PR

- `claimed_at` — tracked separately in #135 / PR #140.
- Queue-scoped `get_job` — #134. Node scoped it in #124; Ruby's lookup stays global and the docstring now says so.
- `spec/phase_mantle_spec.rb` is not run by CI at all (pre-existing gap, unrelated to this change).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -587,6 +587,7 @@ jobs:
           bundle exec ruby -Ilib spec/honker_spec.rb
           bundle exec ruby -Ilib spec/smoke_spec.rb
           bundle exec ruby -Ilib spec/job_fields_spec.rb
+          bundle exec ruby -Ilib spec/phase_mantle_spec.rb
           bundle exec ruby -Ilib spec/parity_spec.rb
           bundle exec ruby -Ilib spec/extension_resolution_spec.rb
           bundle exec ruby -Ilib spec/setup_helpers_spec.rb
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -586,6 +586,7 @@ jobs:
           bundle exec ruby -e 'require "sqlite3"; db = SQLite3::Database.new(":memory:"); abort "sqlite3 gem lacks loadable-extension support" unless db.respond_to?(:enable_load_extension) && db.respond_to?(:load_extension)'
           bundle exec ruby -Ilib spec/honker_spec.rb
           bundle exec ruby -Ilib spec/smoke_spec.rb
+          bundle exec ruby -Ilib spec/job_fields_spec.rb
           bundle exec ruby -Ilib spec/parity_spec.rb
           bundle exec ruby -Ilib spec/extension_resolution_spec.rb
           bundle exec ruby -Ilib spec/setup_helpers_spec.rb
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,23 @@
   unscoped lookup could hand back a payload that did not match the queue's
   declared `TPayload`. `cancel()` is unchanged and remains not queue-scoped.
 
+## Unreleased — Ruby job details
+
+- `Honker::Job` now exposes every field the claim ABI returns: `state`,
+  `priority`, `run_at`, `claim_expires_at`, `max_attempts`, `created_at`,
+  and `expires_at` join the existing `id`, `queue_name`, `payload`,
+  `worker_id`, and `attempts`.
+- `Queue#get_job` returns a `Honker::JobSnapshot` — the same twelve fields,
+  read-only, with no ack/retry/fail/heartbeat. **Behavior change:** it used
+  to return a plain `Hash` of the raw ABI row. `snapshot["state"]` still
+  works (`Struct#[]` takes member names), but `Hash` methods no longer do.
+- Snapshot `payload` stays the raw JSON text the row stores, matching the
+  Python and Go snapshots. `Job#payload` is still decoded. The bindings do
+  not yet agree on one snapshot payload encoding; Node decodes it.
+- `spec/job_fields_spec.rb` asserts every field against the value it was
+  enqueued or claimed with, across pending, delayed, processing, retried,
+  and acked jobs.
+
 ## Unreleased — Node checkpoint interoperability (Phase Robinson)
 
 - Node stream checkpoint calls now use the shared SQL ABI's canonical
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,8 +45,11 @@
   `worker_id`, and `attempts`.
 - `Queue#get_job` returns a `Honker::JobSnapshot` — the same twelve fields,
   read-only, with no ack/retry/fail/heartbeat. **Behavior change:** it used
-  to return a plain `Hash` of the raw ABI row. `snapshot["state"]` still
-  works (`Struct#[]` takes member names), but `Hash` methods no longer do.
+  to return a plain `Hash` of the raw ABI row. Reader access still works
+  (`snapshot["state"]`, `snapshot.dig("state")`), and `JSON.dump(snapshot)`
+  still emits the same JSON object, but Hash-only methods do not: `fetch`,
+  `key?` and `keys` raise `NoMethodError`, an unknown field name raises
+  `NameError` instead of returning `nil`, and `to_h` has Symbol keys.
 - `Job#queue` is a new alias for `Job#queue_name`, so the accessor name
   `JobSnapshot` uses works on a claimed job too.
 - Snapshot `payload` stays the raw JSON text the row stores, matching the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@
   read-only, with no ack/retry/fail/heartbeat. **Behavior change:** it used
   to return a plain `Hash` of the raw ABI row. `snapshot["state"]` still
   works (`Struct#[]` takes member names), but `Hash` methods no longer do.
+- `Job#queue` is a new alias for `Job#queue_name`, so the accessor name
+  `JobSnapshot` uses works on a claimed job too.
 - Snapshot `payload` stays the raw JSON text the row stores, matching the
   Python and Go snapshots. `Job#payload` is still decoded. The bindings do
   not yet agree on one snapshot payload encoding; Node decodes it.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,10 @@
   (`snapshot["state"]`, `snapshot.dig("state")`), and `JSON.dump(snapshot)`
   still emits the same JSON object, but Hash-only methods do not: `fetch`,
   `key?` and `keys` raise `NoMethodError`, an unknown field name raises
-  `NameError` instead of returning `nil`, and `to_h` has Symbol keys.
+  `NameError` instead of returning `nil`, and `to_h` has Symbol keys. One
+  break is silent: `each` and `map` yield the twelve values, not
+  `[key, value]` pairs, so a `|name, value|` block gets a value and `nil`
+  with no error. Iterate over `to_h` instead.
 - `Job#queue` is a new alias for `Job#queue_name`, so the accessor name
   `JobSnapshot` uses works on a claimed job too.
 - Snapshot `payload` stays the raw JSON text the row stores, matching the
--- a/packages/honker-ruby/README.md
+++ b/packages/honker-ruby/README.md
@@ -303,7 +303,10 @@ still emits the same JSON object, but Hash-only methods do not: `fetch`,
 `NameError` instead of returning `nil`, and `to_h` gives you Symbol keys,
 not String ones.
 
-One break is silent rather than loud: iteration. A `Struct`'s `each` and `map` yield the twelve values, not `[key, value]` pairs, so a block written `|name, value|` against the old `Hash` now gets a value and `nil` with no error. Iterate over `snapshot.to_h` instead.
+One break is silent rather than loud: iteration. A `Struct`'s `each` and
+`map` yield the twelve values, not `[key, value]` pairs, so a block
+written `|name, value|` against the old `Hash` now gets a value and `nil`
+with no error. Iterate over `snapshot.to_h` instead.
 
 Honker never inspects a payload. The shape is a contract between the app
 that enqueues and the app that claims, and both sides have to agree on it
--- a/packages/honker-ruby/README.md
+++ b/packages/honker-ruby/README.md
@@ -292,7 +292,8 @@ ack/retry/fail/heartbeat, because the reader holds no claim. `state` is
 `nil` until some worker claims the job. The snapshot's `payload` is the
 raw JSON *text* from the row, not a decoded value — call `JSON.parse` on
 it. `nil` means the job was ack'd, dead-lettered, cancelled, or never
-existed.
+existed. The lookup is by id alone, so an id belonging to another queue
+returns that queue's row rather than `nil`.
 
 `JobSnapshot` is a `Struct`, like `Honker::Notification`. It replaces the
 plain `Hash` older versions returned. Reader access still works
@@ -305,7 +306,14 @@ not String ones.
 Honker never inspects a payload. The shape is a contract between the app
 that enqueues and the app that claims, and both sides have to agree on it
 — including across languages, since another binding may write to the same
-queue. Version the payload if it will change:
+queue.
+
+The payload goes through `JSON.dump` on the way in and `JSON.parse` on
+the way out, so it must be JSON-serializable, and **Symbol keys come back
+as Strings**: `q.enqueue({to: "alice"})` claims as `{"to" => "alice"}`.
+Write `job.payload["to"]`, not `job.payload[:to]`.
+
+Version the payload if it will change:
 
 ```ruby
 q.enqueue({"v" => 2, "to" => "alice@example.com"})
--- a/packages/honker-ruby/README.md
+++ b/packages/honker-ruby/README.md
@@ -294,6 +294,14 @@ raw JSON *text* from the row, not a decoded value — call `JSON.parse` on
 it. `nil` means the job was ack'd, dead-lettered, cancelled, or never
 existed.
 
+`JobSnapshot` is a `Struct`, like `Honker::Notification`. It replaces the
+plain `Hash` older versions returned. Reader access still works
+(`snapshot["state"]`, `snapshot.dig("state")`), and `JSON.dump(snapshot)`
+still emits the same JSON object, but Hash-only methods do not: `fetch`,
+`key?` and `keys` raise `NoMethodError`, an unknown field name raises
+`NameError` instead of returning `nil`, and `to_h` gives you Symbol keys,
+not String ones.
+
 Honker never inspects a payload. The shape is a contract between the app
 that enqueues and the app that claims, and both sides have to agree on it
 — including across languages, since another binding may write to the same
--- a/packages/honker-ruby/README.md
+++ b/packages/honker-ruby/README.md
@@ -263,3 +263,47 @@ Supported schedule forms:
 - `@every 1s`
 
 `schedule:` is the canonical recurring name. `cron:` still works as a compatibility alias.
+
+## Job details
+
+A claimed `Honker::Job` carries the whole row as it stood at claim time:
+
+```ruby
+job = q.claim_one("worker-1")
+
+job.id                # row id
+job.queue_name        # queue this job came from
+job.payload           # decoded JSON value
+job.state             # "processing"
+job.priority          # higher runs first within the queue
+job.run_at            # unix seconds; when it became claimable
+job.worker_id         # "worker-1"
+job.claim_expires_at  # unix seconds; heartbeat before this
+job.attempts          # already incremented by this claim
+job.max_attempts      # dead-letters once attempts reaches this
+job.created_at        # unix seconds
+job.expires_at        # unix seconds, or nil when enqueued without expires:
+```
+
+`q.get_job(id)` returns a `Honker::JobSnapshot` — the same twelve fields,
+read-only, for a job you did not claim. It is data alone: no
+ack/retry/fail/heartbeat, because the reader holds no claim. `state` is
+`"pending"` or `"processing"`, and `worker_id` / `claim_expires_at` are
+`nil` until some worker claims the job. The snapshot's `payload` is the
+raw JSON *text* from the row, not a decoded value — call `JSON.parse` on
+it. `nil` means the job was ack'd, dead-lettered, cancelled, or never
+existed.
+
+Honker never inspects a payload. The shape is a contract between the app
+that enqueues and the app that claims, and both sides have to agree on it
+— including across languages, since another binding may write to the same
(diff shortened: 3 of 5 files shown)


russellromney:
## Adversarial review — #148

Labelled **RAN** (executed) or **INFERRED** (read, not executed). Run locally against an extension built from this branch (the claim ABI changed, so a prebuilt one from `main` does not work). Baseline: **6 runs, 52 assertions, 0 failures, 0 skips**.

### 1. The spec cannot tell `run_at` from `created_at` on a claimed job — **RAN**

Both mutations to `Job#initialize` leave the spec fully green:

```ruby
@run_at = row["created_at"]     # 6 runs, 52 assertions, 0 failures
@created_at = row["run_at"]     # 6 runs, 52 assertions, 0 failures
```

Control that fails, confirming the rest of the mapping is covered:

```ruby
@priority = row["max_attempts"] # 6 runs, 44 assertions, 1 failure
```

Cause: `test_claimed_job_carries_every_field` asserts `assert_equal pending.run_at, job.run_at` on a job enqueued with no delay, so both sides are equal under the swap; and `test_delayed_job_reports_its_run_at` never claims (it asserts `claim_one` is nil). So no claimed job in this spec ever has `run_at != created_at`.

**Fix — one line.** Ruby's `enqueue` already takes an absolute `run_at:`, so backdate rather than delay:

```ruby
id = @q.enqueue(payload, run_at: db_now - 100)
job = @q.claim_one("worker-a")
refute_equal job.created_at, job.run_at
```

RAN: gives `run_at=1788316878`, `created_at=1788316978`, and `claim_one` still returns the job. Same hole is mutation-proven in #124, #143 and #149.

### 2. `job_fields_spec.rb` skips silently and exits 0 — **RAN**

With no `target/` directory:

```
6 runs, 0 assertions, 0 failures, 0 errors, 6 skips
EXIT CODE: 0
```

— and that is *with* `HONKER_REQUIRE_RUBY_EXTENSION_LOADING=1` set. The whole acceptance spec passes while asserting nothing.

Two separate things here:

- The new spec never calls `require_load_extension_support!`, which `honker_spec.rb`, `parity_spec.rb` and `setup_helpers_spec.rb` all do. That helper is what honours `HONKER_REQUIRE_RUBY_EXTENSION_LOADING=1` (it `flunk`s instead of skipping). `job_fields_spec.rb` and `smoke_spec.rb` are the only two that omit it, so this spec does not honour the flag at all.
- The `skip "honker extension not built" unless ext` guard is unguarded in *every* Ruby spec, so that part is pre-existing, not new here.

**In CI the skip cannot silently fire** (INFERRED, from `ci.yml`): the `Build honker-extension` step runs first and a failure there fails the job, and the crate's lib name is `honker_ext`, so `target/release/libhonker_ext.so` will exist for `find_ext`. So this is a local-developer trap and a consistency gap, not a live CI hole. Adding the `require_load_extension_support!` call would close half of it cheaply.

### 3. `get_job` returns a `Struct` instead of a `Hash` — break on a published gem, in practice benign — **RAN**

The one in-repo consumer, `phase_mantle_spec.rb`, uses `row["queue"]` / `row["state"]` / `row["id"]`, and `Struct#[]` accepts a String member name, so it still passes (8 runs, 44 assertions, 0 failures). Worth noting anyway: `fetch`, `key?`, `dig`, `keys` and `JSON.dump` (which now produces an array, not an object) all change for external callers. Also note `phase_mantle_spec.rb` is not in the CI Ruby step's spec list, so that caller is not actually exercised by CI (INFERRED, from `ci.yml`).

### 4. `Job` has `queue_name` but no `queue`; `JobSnapshot` has both — **RAN**

```
snapshot.queue      = "e"
snapshot.queue_name = "e"   (alias)
job.queue_name      = "e"
job.queue           = NoMethodError - Job has no #queue
```

So code written against a snapshot breaks when handed a claimed job. Elixir uses `queue` on both. A matching `alias_method :queue, :queue_name` on `Job` would make the two interchangeable.

### What holds up — **RAN**

Distinct constants for priority / max_attempts / visibility / delay / expiry mean no assertion passes by coincidence. `assert_nil ... "must be nil, not 0"` on `expires_at`. `claim_expires_at` is bounded to `before+VISIBILITY_S .. after+VISIBILITY_S` rather than merely `> 0` — the strongest version of that assertion in the stack. Full Ruby suite green: `honker_spec` 26/117, `parity_spec` 25/70, `smoke_spec` 9/26, `phase_mantle_spec` 8/44, all 0 failures. The `JobSnapshot` doc names the raw-text-vs-decoded payload divergence honestly.

**Recommendation:** merge once the `run_at` assertion is tightened; optionally add the `require_load_extension_support!` call for consistency with the other specs.

russellromney:
## Findings 1 and 2 fixed

Two atomic commits:

- `81149ee` — `test(ruby): fail loudly when the extension is missing in job_fields_spec`
- `2e007c3` — `test(ruby): prove a claimed job reports run_at, not created_at`

### Finding 1 — the `run_at` / `created_at` gap is now load-bearing

Added `test_claimed_job_reports_its_own_run_at_not_created_at`. It uses `@q.enqueue(..., run_at: db_now - 100)` — a **back-dated absolute run_at**, not a delay — so the job stays claimable while `run_at != created_at`, and it pins `run_at` to the exact value enqueued rather than comparing it to another field.

**Mutation — claimed `Job#initialize` (`lib/honker.rb:946/951`)**

```ruby
@run_at = row["created_at"]
@created_at = row["run_at"]
```

```
  1) Failure:
HonkerJobFieldsTest#test_claimed_job_reports_its_own_run_at_not_created_at [spec/job_fields_spec.rb:178]:
Expected: 1788318894
  Actual: 1788318994

7 runs, 57 assertions, 1 failures, 0 errors, 0 skips
```

Line 178 is `assert_equal run_at, job.run_at`. **Only the new test fails** — the other six stay green, which is the blind spot you reported.

I also mutated `JobSnapshot.from_row` (`lib/honker.rb:910/915`) for completeness: 3 failures, the new test among them (`assert_equal run_at, snap.run_at`, line 172). That site was already covered by the delayed test; the claimed path was the hole.

Both sites restored; `git diff` on `lib/` is empty. Baseline is now **7 runs, 59 assertions, 0 failures, 0 skips** (was 6/52).

### Finding 2 — my call: I fixed it rather than leaving it

You framed this as optional. I closed it, because a spec that reports "0 assertions, 0 failures" and exits 0 is exactly the failure mode that let finding 1 through in the first place, and the fix is a few lines.

`job_fields_spec.rb` now calls both guards in `setup`:

- `require_load_extension_support!` — the same helper `honker_spec`, `parity_spec` and `setup_helpers_spec` use, so the sqlite3 gem capability check honours `HONKER_REQUIRE_RUBY_EXTENSION_LOADING=1`.
- `require_built_extension!(ext)` — new, and the one that actually closes the hole you measured: a missing build artifact now `flunk`s under that flag instead of skipping.

Both are defined behind `unless defined?(...)`, matching `setup_helpers_spec.rb`, since CI runs each spec as its own process.

**Proof the guard is load-bearing** — with `target/release/libhonker_ext.dylib` removed:

```
$ HONKER_REQUIRE_RUBY_EXTENSION_LOADING=1 bundle exec ruby -Ilib spec/job_fields_spec.rb
HonkerJobFieldsTest#test_pending_snapshot_carries_every_field [spec/job_fields_spec.rb:49]:
honker extension not built
...
7 runs, 7 assertions, 7 failures, 0 errors, 0 skips
EXIT CODE: 1
```

and without the flag it still skips for local developers:

```
7 runs, 0 assertions, 0 failures, 0 errors, 7 skips
EXIT CODE: 0
```

With the extension present it is back to exit 0. As you noted, CI could not have hit the silent pass today — the `Build honker-extension` step runs first — so this closes a local-developer trap, not a live hole.

I left the unguarded `skip ... unless ext` in the *other* specs alone: that is pre-existing and out of scope for this PR.

Findings 3 and 4 (`Struct` vs `Hash`, `Job#queue` alias) are untouched.

russellromney:
CI note: the first run failed on `jvm · honker + documented ORM recipes` — `HonkerJvmTest.concurrentWorkersClaimEachJobOnce:670 expected: <true> but was: <false>`. That is a JVM concurrency test, untouched by this PR (the diff is two additions to `packages/honker-ruby/spec/job_fields_spec.rb` and nothing else), the same job passed on #124 off the same base, and this file has a history of timing-assertion fixes (#114, #121). Re-ran the failed job only; it passed. **All 19 checks green.**

russellromney:
## Adversarial review — round two verification

Second-round review of `feat/ruby-job-parity`. I did not write this code. The round-two fixes (`40446c6`, `1252fcd`, `322d981`, `5b72707`) were pushed but never verified in writing, so this is that write-up, plus two new findings I fixed on the branch.

Every claim below is marked **RUN** (I executed it and pasted the result) or **INFERRED** (I read it and did not execute it).

**Environment (RUN):** macOS, Ruby 4.0.5, sqlite3 gem 2.9.6, debug `libhonker_ext.dylib` from `cargo build -p honker-extension`, `HONKER_REQUIRE_RUBY_EXTENSION_LOADING=1`. Baseline at `5b72707`: `spec/job_fields_spec.rb` is 12 runs, 99 assertions, 0 failures, 0 errors, **0 skips**.

### Method

I did not take commit messages as findings. For each new or changed spec I broke the behavior it guards in `lib/honker.rb`, ran the spec, watched it fail, restored it, watched it pass. A spec that survived its own mutation would itself be a finding. None did.

### Round-two findings, re-derived and confirmed

**1. `1252fcd` — `get_job`'s JSON shape was silently broken. Real, and the most dangerous of the four.**

This is where a Struct that has to keep quacking like a Hash hides a break, so I probed the whole surface instead of trusting the doc comment.

RUN — mutation: delete `JobSnapshot#to_json`. `test_snapshot_keeps_the_reader_and_json_shape_the_hash_had` fails:

```
Expected "#<struct Honker::JobSnapshot id=1, queue=\"details\", payload=\"{\\\"x\\\":1}\", ...>"
to be an instance of Hash, not String.
```

So before `1252fcd`, `JSON.dump(snapshot)` produced the Struct's `#inspect` string. A caller logging or shipping a snapshot got garbage that still parsed as valid JSON — a String, not an object. Genuine defect, genuinely fixed, genuinely pinned.

RUN — full shape probe of what a pre-PR caller gets now (a script, not the spec):

| call | result |
| --- | --- |
| `snap["payload"]`, `snap[:state]`, `snap[3]`, `snap.dig("state")` | work, same values as the Hash |
| `JSON.dump(snap)`, `snap.to_json`, `{"j"=>snap}.to_json`, `[snap].to_json`, `JSON.generate`, `JSON.pretty_generate` | all emit the row's JSON object |
| `snap.fetch("state")`, `snap.key?(...)`, `snap.keys`, `snap.merge(...)` | raise `NoMethodError` |
| `snap["nope"]` | raises `NameError: no member 'nope' in struct` |
| `snap.to_h["state"]` | `nil` (Symbol keys) |
| `snap.each { \|k, v\| }` | **yields `[1, nil]`, `["details", nil]` …** |

The last row is a new finding — see B below. Everything else matches what the code comment, README and CHANGELOG already claim, so those docs are accurate as far as they go.

**2. `2e007c3` does the right thing on the round-one transposition trap.**

RUN — mutation: `@run_at = row["created_at"]` in `Job#initialize`, then the three `/claimed_job/` tests:

```
.F.
1) Failure: test_claimed_job_reports_its_own_run_at_not_created_at
   Expected: 1788553143
     Actual: 1788553243
3 runs, 22 assertions, 1 failures
```

`test_claimed_job_carries_every_field` **passed** under that mutation. It compares `job.run_at` against a snapshot of an undelayed enqueue, where the two fields hold the same second — exactly the round-one wrong-reason pass. `2e007c3` back-dates an absolute `run_at` (`db_now - 100`), which keeps the job claimable and makes the fields genuinely differ, and it is the only test that catches the transposition. Correct fix.

RUN — the same mutation on the snapshot side (`from_row` run_at ← created_at) fails 3 tests, including `test_delayed_job_reports_its_run_at`.

**3. `322d981` — nil expiry is real coverage, not a filler test.**

RUN — mutation: `@expires_at = row["expires_at"] || 0` in `Job`:

```
1) Failure: test_claimed_job_without_expiry_has_nil_expires_at
   expires_at must be nil, not 0. Expected 0 to be nil.
```

The three other tests in that filter passed, including `test_claimed_job_carries_every_field` — it always enqueues with `expires:`, so it cannot see a nil-to-0 coercion. The new test is the only guard. Same result on the snapshot side.

RUN — the second-reader test is load-bearing too: mutating `from_row`'s `worker_id` to `nil` fails both `test_a_second_connection_sees_the_claim_and_then_the_ack` and `test_processing_snapshot_matches_the_claim_then_misses_after_ack`.

**4. `40446c6` — the alias is additive, and the primary name is documented.**

RUN — mutation: delete `alias_method :queue, :queue_name` from `Job` → `NoMethodError: undefined method 'queue' for an instance of Honker::Job`. So `Job#queue` did **not** exist before this PR. The alias adds a name; it does not rename or shadow one. The `@queue` ivar (the `Queue` object) stays private and is untouched.

RUN — mutation: delete `alias_method :queue_name, :queue` from `JobSnapshot` → `NoMethodError: undefined method 'queue_name'`. Both directions are guarded by `test_job_and_snapshot_agree_on_the_queue_accessor`.

RUN (grep) — no other caller in `packages/honker-ruby` or `site/` uses either name, so no collision. The README documents it: `job.queue_name  # queue this job came from (also job.queue)`. `Job` keeps `queue_name` as primary (pre-existing), `JobSnapshot` keeps `queue` (matches the ABI column and Go/Node/Elixir). Two primaries across two classes is a little awkward, but each answers to both and it is written down. Accepted, not a defect.

**5. `81149ee` — the fail-loudly helper actually fails loudly.**

RUN — moved the built extension aside:

- with `HONKER_REQUIRE_RUBY_EXTENSION_LOADING=1` → `12 runs, 12 assertions, 12 failures, 0 skips` ("honker extension not built")
- without the flag → `12 runs, 0 assertions, 0 failures, 12 skips`

Correct in both modes.

### New findings (fixed on this branch)

**A. `spec/phase_mantle_spec.rb` has never run in CI, and it is the only pre-existing spec that reads a `get_job` result. Fixed in `f86d5fd`.**

RUN — `git log -S"phase_mantle_spec" -- .github/workflows/ci.yml` returns nothing. The ruby step runs seven spec files; this is the eighth and was never on the list. It asserts `row["queue"]`, `row["state"]`, `row["id"]` on a `get_job` result — precisely the reader compatibility this PR's headline behavior change rests on.

RUN — it is a real gate, not decoration. Mutating `from_row` to map `queue` from `row["state"]`:

```
1) Failure: PhaseMantleTest#test_queue_cancel_and_get_job
   Expected: "emails"
     Actual: "pending"
```

RUN — unmutated against the new Struct it is clean: `8 runs, 44 assertions, 0 failures, 0 errors, 0 skips`. So the Hash-to-Struct change does not break the one piece of existing caller code in the repo — but nothing in CI was checking that. A pre-existing repo gap that this PR made load-bearing; the fix is one line in the block the PR already edits.

**B. The list of Hash-to-Struct breaks was missing the only one that does not raise. Fixed in `4bc4789`.**

The docs list `fetch`, `key?`, `keys`, unknown names, and `to_h`'s Symbol keys — every one of which raises or is visible at the call site. Iteration is not in the list, and it changes quietly: a Struct's `each`/`map` yield the twelve values, so a block written `|name, value|` against the old Hash now gets a value and `nil` with no error (`[[1, nil], ["details", nil], ...]` in the probe above). Named it in the `JobSnapshot` docs, the README and the CHANGELOG, and pinned it with `test_snapshot_iteration_yields_values_not_key_value_pairs`.

RUN — the pin is load-bearing: defining `JobSnapshot#each` to delegate to `to_h.each` fails it (`[1, "details", ...]` vs `[[:id, 1], [:queue, "details"], ...]`). With no mutation the file is `13 runs, 105 assertions, 0 failures, 0 errors, 0 skips`.

I documented the behavior rather than changing it. Making `each` yield pairs is new semantics on a Struct, not a defect inside the accepted design.

(`3e024a9` is whitespace only — it wraps the new README paragraph like its neighbors.)

### Checked and clean

- **JSON decode fails loudly (RUN).** Inserted a row whose payload is the literal text `not json at all` and claimed it: `JSON::ParserError: unexpected token 'not' at line 1 column 1`. No rescue, no fallback to a bare string, no nil. `grep -n rescue lib/honker.rb` finds six rescues; none is in the `Job` / `JobSnapshot` / `get_job` path.
- **Field mapping against the twelve-column contract (RUN + read).** Both `honker_ops.rs::claim_batch` and `honker_ops.rs::get_job` emit a JSON **object** keyed by column name, and Ruby maps by name in both `Job#initialize` and `JobSnapshot.from_row`, so column order cannot transpose fields here. All twelve names match exactly. Mutating `from_row`'s `max_attempts` to `row["attempts"]` fails 2 tests; mutating `payload` to skip `JSON.parse` fails 2 tests.
- **nil vs 0 vs absent (RUN).** `worker_id`, `claim_expires_at` and `expires_at` come through as `nil`, and `JSON.dump(snapshot)` emits them as JSON `null` with the key present — the spec asserts `round_trip.key?("worker_id")`. `attempts` 0 vs nil is covered on the pending snapshot.
- **The cross-binding claims in the CHANGELOG are true (read).** Go's `JobRow.Payload` is a `string`, and Python's `get_job` returns the row with `payload` as raw text, so "matching the Python and Go snapshots" holds. Node's `getJob` does call `parseJson`, so "Node decodes it" holds too.
- `spec/job_fields_spec.rb` uses distinct constants (priority 7, max_attempts 5, visibility 11, delay 300, expires 900) so no assertion can pass by two numbers coinciding. That is the right instinct, and it is why each mutation above lands on exactly one test.

### Observations, not fixed

- **`get_job` is still unscoped, and Node is now scoped.** `5a25c84`, in this branch's base, made Node's `getJob` return `null` for an id from another queue. Ruby, Go, Python and Elixir all still look up by id alone. This PR documents the Ruby behavior honestly and points at #134, which is the right call for this PR — but once this merges the bindings disagree on `getJob` scoping in a way a user can hit, so #134 is more urgent than it was.
- **`Listener#decode_payload` (`lib/honker.rb:442`) rescues `JSON::ParserError` and returns the raw text.** Pre-existing, on the notify path, not touched by this PR, and documented at the `Notification` struct as intended, since raw SQL `notify()` payloads need not be JSON. Naming it because it is the one rescue-and-continue left in the file. The job path is clean.
- **`phase_mantle_spec.rb` skips silently when the extension is missing**, unlike `job_fields_spec.rb`'s flunk-under-the-CI-flag helper from `81149ee`. `honker_spec.rb` and `smoke_spec.rb` have the same plain `skip` and are already in CI, so this is a repo-wide pattern rather than something to change inside this PR. Worth a follow-up moving all of them onto the `81149ee` helper.

### CI after my three commits (RUN)

Head is `3e024a9`. `gh pr checks 148` is 19 pass, 2 skipping (the two bench regression checks), 0 failures — `ci · all green` passed. The ruby step in `binding smoke · ubuntu aggregate` now shows the new file running:

```
bundle exec ruby -Ilib spec/phase_mantle_spec.rb
...
8 runs, 44 assertions, 0 failures, 0 errors, 0 skips
```

and `job_fields_spec.rb` at `13 runs, 0 failures, 0 errors, 0 skips` on the runner.

### Verdict

The four round-two commits are real fixes to real defects, and every spec they added or changed fails when the behavior it guards is broken. I found and fixed two more things: a dark spec that gates the exact method this PR changes, and an undocumented silent break in the Hash-to-Struct migration. No unresolved defects from my side.

Issue #146, opened 2026-09-01
Snapshot payload encoding differs across bindings


russellromney:
Read-only job snapshots disagree on whether `payload` is decoded or raw JSON text:

- Node (`getJob()`, after #124) — decoded value
- Go (`JobRow.Payload`) — raw JSON text
- Python (`get_job()["payload"]`) — raw JSON text

Claimed `Job.payload` is decoded everywhere; only the snapshot path diverges.

Surfaced while doing #136. The binding PRs deliberately preserve each binding's
current encoding, because changing it is a breaking API change #136 did not ask
for. That leaves the inconsistency in place and growing as the remaining
bindings land.

Needs one decision applied everywhere, then a coordinated change:

- decode in every snapshot (matches claimed jobs; breaks Go and Python callers), or
- raw text in every snapshot (breaks Node callers on an API added in #124, which
  is not released yet — cheapest window to change it is before #124 ships)

russellromney:
Updated tally as the #136 binding PRs land. Three conventions now, not two:

- **decoded** — Node `getJob()` (#124)
- **raw JSON text** — Go, Python, JVM, Kotlin, Bun
- **both** — .NET's new `JobSnapshot<T>` exposes `Payload` (decoded) and `PayloadRaw`

Every binding PR preserved its existing encoding rather than picking a side, so
nothing new is broken — but "both" is worth considering as the resolution. It
costs one extra field, breaks no caller, and lets the decoded form be the
documented default without forcing a decode on bindings where raw text is the
natural shape.

Still cheapest to decide before #124 ships, since Node's decoded `getJob()` is
the only unreleased side.

PR #149, opened 2026-09-01
feat(elixir): carry full job details on claims and snapshots

russellromney:
Elixir half of #136. Stacked on #124 (`feat/node-typed-jobs`) and targets that branch: the widened `honker_claim_batch` RETURNING clause that exposes the full job snapshot only exists there. **Merge after #124.** Sibling PR #148 does the same for Ruby.

## What changed

- `%Honker.Job{}` gains `:state`, `:priority`, `:run_at`, `:claim_expires_at`, `:max_attempts`, `:created_at`, and `:expires_at`, joining the existing `:id`, `:queue`, `:payload`, `:worker_id`, and `:attempts` — the twelve fields the claim ABI already returned and the binding threw away. `@type t` added.
- New `Honker.JobSnapshot` struct, returned by `Honker.Queue.get_job/2`. Same twelve fields, data only: no ack/retry/fail/heartbeat, because the reader holds no claim.
- README documents the fields and the payload-shape contract. No runtime payload validation was added — Elixir has no static payload typing and Honker never checks payload shape.
- No change to honker-core or the SQL ABI.

### Behavior change

`get_job/2` used to return `{:ok, map}` with the raw ABI row's string keys. It now returns `{:ok, %Honker.JobSnapshot{}}`, so `row["state"]` becomes `row.state`. The one existing caller (`test/phase_mantle_test.exs`) is updated in this PR.

### Snapshot payload encoding — needs one deliberate decision later

Read-only snapshots disagree across bindings today:

| binding | `get_job`/`getJob` payload |
| --- | --- |
| Elixir (this PR) | raw JSON **text** — unchanged from the old map |
| Ruby (#148) | raw JSON **text** — unchanged |
| Python | raw JSON text |
| Go (`JobRow.Payload`) | raw JSON text |
| Node (#124) | **decoded** value |

This PR deliberately preserves Elixir's existing encoding rather than silently picking a side; changing it is a breaking API change #136 did not ask for. Claimed job payloads are decoded in every binding, so Elixir is internally split between `Job`'s decoded `:payload` and `JobSnapshot`'s text `:payload` — documented in both the module doc and the README. Worth one repo-wide decision.

## Tests

New `packages/honker-ex/test/job_fields_test.exs` — picked up by `mix test`, no CI change needed. Every field is asserted against the value it was enqueued or claimed with, not for presence. Constants are deliberately distinct (priority 7, max_attempts 5, visibility 11s, delay 300s, expiry 900s) so no assertion can pass by two values coinciding, and timestamps are checked as differences (`created_at + 900 == expires_at`) or against a `unixepoch()` read taken around the call, never against a same-second sibling.

- `a pending snapshot carries every field`
- `a job enqueued without :expires has a nil expires_at`
- `a delayed job reports its run_at`
- `a claimed job carries every field`
- `a processing snapshot matches the claim, then misses after ack`
- `retry keeps attempts and pushes run_at out`

Local: `mix test test/job_fields_test.exs` — 6 tests, 0 failures. Full suite `mix test` — 74 tests, 0 failures.

### The tests were proven load-bearing

Mutating `Queue.row_to_job/1` to read `priority: row["max_attempts"]`:

```
1) test a claimed job carries every field (HonkerJobFieldsTest)
   test/job_fields_test.exs:116
   Assertion with == failed
   code:  assert job.priority == @priority
   left:  5
   right: 7
```

Mutating `JobSnapshot.from_row/1` to pass `row["created_at"]` for `:expires_at`:

```
2) test a pending snapshot carries every field (HonkerJobFieldsTest)
   code:  assert abs(snap.expires_at - (snap.created_at + @expires_s)) <= 1
   left:  900
   right: 1

3) test a claimed job carries every field (HonkerJobFieldsTest)
   code:  assert job.expires_at == pending.expires_at
   left:  1788257207
   right: 1788256307
```

Both mutations restored; suite green again.

## Not in this PR

- `claimed_at` — tracked separately in #135 / PR #140.
- Queue-scoped `get_job` — #134. Node scoped it in #124; Elixir's lookup stays global and the docstring now says so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # CHANGELOG
 
+## Unreleased — Elixir job details
+
+- `%Honker.Job{}` now carries every field the claim ABI returns: `:state`,
+  `:priority`, `:run_at`, `:claim_expires_at`, `:max_attempts`, `:created_at`,
+  and `:expires_at` join the existing `:id`, `:queue`, `:payload`,
+  `:worker_id`, and `:attempts`.
+- New `Honker.JobSnapshot` struct: the same twelve fields, read-only, with no
+  ack/retry/fail/heartbeat. **Behavior change:** `Honker.Queue.get_job/2`
+  returns `{:ok, %Honker.JobSnapshot{}}` instead of a map of the raw ABI row,
+  so `row["state"]` becomes `row.state`.
+- Snapshot `:payload` stays the raw JSON text the row stores, matching the
+  Python and Go snapshots. `Honker.Job`'s `:payload` is still decoded. The
+  bindings do not yet agree on one snapshot payload encoding; Node decodes it.
+- `test/job_fields_test.exs` asserts every field against the value it was
+  enqueued or claimed with, across pending, delayed, processing, retried, and
+  acked jobs.
+
 ## 2026-08-27 — Node 0.5.1
 
 - Node `@russellthehippo/honker-node`: 0.5.1, with the four
--- a/packages/honker-ex/README.md
+++ b/packages/honker-ex/README.md
@@ -80,3 +80,47 @@ Supported schedule forms:
 - `@every 1s`
 
 `schedule:` is the canonical recurring name. `cron:` is compatibility-only.
+
+## Job details
+
+A claimed `%Honker.Job{}` carries the whole row as it stood at claim time:
+
+```elixir
+{:ok, job} = Honker.Queue.claim_one(db, "emails", "worker-1")
+
+job.id                # row id
+job.queue             # queue this job came from
+job.payload           # decoded JSON value
+job.state             # "processing"
+job.priority          # higher runs first within the queue
+job.run_at            # unix seconds; when it became claimable
+job.worker_id         # "worker-1"
+job.claim_expires_at  # unix seconds; heartbeat before this
+job.attempts          # already incremented by this claim
+job.max_attempts      # dead-letters once attempts reaches this
+job.created_at        # unix seconds
+job.expires_at        # unix seconds, or nil when enqueued without :expires
+```
+
+`Honker.Queue.get_job/2` returns `{:ok, %Honker.JobSnapshot{}}` — the same
+twelve fields, read-only, for a job you did not claim. It is data alone: no
+ack/retry/fail/heartbeat, because the reader holds no claim. `state` is
+`"pending"` or `"processing"`, and `worker_id` / `claim_expires_at` are `nil`
+until some worker claims the job. The snapshot's `payload` is the raw JSON
+*text* from the row, not a decoded value — call `Jason.decode!/1` on it.
+`{:ok, nil}` means the job was ack'd, dead-lettered, cancelled, or never
+existed.
+
+Honker never inspects a payload. The shape is a contract between the app that
+enqueues and the app that claims, and both sides have to agree on it —
+including across languages, since another binding may write to the same queue.
+Version the payload if it will change:
+
+```elixir
+{:ok, _id} = Honker.Queue.enqueue(db, "emails", %{"v" => 2, "to" => "alice@example.com"})
+
+case job.payload do
+  %{"v" => 2, "to" => to} -> send_email(to)
+  other -> raise "unknown payload version: #{inspect(other)}"
+end
+```
--- a/packages/honker-ex/README.md
+++ b/packages/honker-ex/README.md
@@ -111,6 +111,17 @@ until some worker claims the job. The snapshot's `payload` is the raw JSON
 `{:ok, nil}` means the job was ack'd, dead-lettered, cancelled, or never
 existed.
 
+`get_job/2` looks a job up by id and takes no queue name, so an id from
+another queue still resolves today; the snapshot's `queue` names the real
+one. Do not build on that — queue-scoped lookup is planned and would
+narrow it. Check `snapshot.queue` yourself if it matters.
+
+**Breaking change.** `get_job/2` used to return `{:ok, map}` — the raw ABI
+row keyed by strings. It now returns a struct, so `row["state"]` becomes
+`row.state`. A struct has no `Access` behaviour, so the old bracket form
+raises `UndefinedFunctionError` instead of quietly returning nil; the
+compiler will not catch it for you, so grep your callers.
+
 Honker never inspects a payload. The shape is a contract between the app that
 enqueues and the app that claims, and both sides have to agree on it —
 including across languages, since another binding may write to the same queue.
--- a/packages/honker-ex/lib/honker/job.ex
+++ b/packages/honker-ex/lib/honker/job.ex
@@ -4,7 +4,9 @@ defmodule Honker.Job do
   (`ack/2`, `retry/4`, `fail/3`, `heartbeat/3`) to close it out.
 
   The struct itself is a plain data carrier — all state lives in the
-  database; Elixir just holds a snapshot of the row.
+  database; Elixir just holds a copy of the row as it stood at claim
+  time. To read a job you did *not* claim, use `Honker.Queue.get_job/2`,
+  which returns the read-only `Honker.JobSnapshot`.
 
   Fields, as the row stood at claim time:
 
--- a/packages/honker-ex/lib/honker/job.ex
+++ b/packages/honker-ex/lib/honker/job.ex
@@ -5,9 +5,57 @@ defmodule Honker.Job do
 
   The struct itself is a plain data carrier — all state lives in the
   database; Elixir just holds a snapshot of the row.
+
+  Fields, as the row stood at claim time:
+
+    * `:id`               — row id
+    * `:queue`            — queue this job came from
+    * `:payload`          — the decoded JSON value
+    * `:state`            — `"processing"`
+    * `:priority`         — higher runs first within the queue
+    * `:run_at`           — unix seconds; when it became claimable
+    * `:worker_id`        — the claiming worker
+    * `:claim_expires_at` — unix seconds; heartbeat before this
+    * `:attempts`         — already incremented by this claim
+    * `:max_attempts`     — dead-letters once `:attempts` reaches this
+    * `:created_at`       — unix seconds
+    * `:expires_at`       — unix seconds, or nil when enqueued without
+      `:expires`
+
+  Honker never inspects a payload. Its shape is a contract between the
+  app that enqueues and the app that claims — including across
+  languages, since another binding may write to the same queue.
   """
 
-  defstruct [:id, :queue, :payload, :worker_id, :attempts]
+  defstruct [
+    :id,
+    :queue,
+    :payload,
+    :state,
+    :priority,
+    :run_at,
+    :worker_id,
+    :claim_expires_at,
+    :attempts,
+    :max_attempts,
+    :created_at,
+    :expires_at
+  ]
+
+  @type t :: %__MODULE__{
+          id: integer(),
+          queue: String.t(),
+          payload: term(),
+          state: String.t(),
+          priority: integer(),
+          run_at: integer(),
+          worker_id: String.t(),
+          claim_expires_at: integer(),
+          attempts: integer(),
+          max_attempts: integer(),
+          created_at: integer(),
+          expires_at: integer() | nil
+        }
 
   alias Honker.Database
 
--- a/packages/honker-ex/lib/honker/job_snapshot.ex
+++ b/packages/honker-ex/lib/honker/job_snapshot.ex
@@ -1,0 +1,79 @@
+defmodule Honker.JobSnapshot do
+  @moduledoc """
+  A read-only view of a live job row, returned by
+  `Honker.Queue.get_job/2`.
+
+  Data only — no `ack`/`retry`/`fail`/`heartbeat`, because the reader
+  does not hold the claim. The fields are the same twelve a claimed
+  `Honker.Job` carries:
+
+    * `:id`               — row id
+    * `:queue`            — queue that owns the row
+    * `:payload`          — the raw JSON **text** stored in the row
+    * `:state`            — `"pending"` or `"processing"`
+    * `:priority`         — higher runs first within the queue
+    * `:run_at`           — unix seconds; when it becomes claimable
+    * `:worker_id`        — nil until a worker claims the job
+    * `:claim_expires_at` — nil until a worker claims the job
+    * `:attempts`         — claims made so far
+    * `:max_attempts`     — dead-letters once `:attempts` reaches this
+    * `:created_at`       — unix seconds
+    * `:expires_at`       — unix seconds, or nil when enqueued without
+      `:expires`
+
+  NOTE: `:payload` is the raw JSON text, not a decoded value — unlike
+  `Honker.Job.payload`, which `Honker.Queue.claim_batch/4` decodes.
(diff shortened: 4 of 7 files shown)


russellromney:
Accepted: keep the struct, take the break.

`Queue.get_job/2` returning `%Honker.JobSnapshot{}` instead of a map means
`row["state"]` becomes `row.state` for anyone on the published Hex package. That
is the right Elixir even though it costs a break — Ruby staying source-compatible
via `Struct` is a property of Ruby, not a bar Elixir has to clear.

Goes in the combined migration guide with the other breaking changes in this
release (#124's `getJob` reshape, #152's JSON contract, #134's queue scoping),
rather than shipping on its own.

russellromney:
## Adversarial review — #149

Labelled **RAN** (executed) or **INFERRED** (read, not executed). Run against an extension built from this branch. Baseline: **6 tests, 0 failures**; full suite **74 tests, 0 failures**.

### 1. The test cannot tell `run_at` from `created_at` on a claimed job — **RAN**

Both mutations to the `Job` builder in `queue.ex` leave the test green:

```elixir
run_at: row["created_at"],      # 6 tests, 0 failures
created_at: row["run_at"],      # 6 tests, 0 failures
```

Control that fails:

```elixir
priority: row["max_attempts"],  # 6 tests, 1 failure
```

Cause, same as the other three bindings: `assert job.run_at == pending.run_at` runs on a job enqueued with no delay, and the delayed test never claims (`assert {:ok, nil} = Queue.claim_one(...)`).

**Fix — one line.** `Queue.enqueue` already accepts `:run_at`:

```elixir
{:ok, id} = Queue.enqueue(db, "details", payload, run_at: now - 100)
{:ok, job} = Queue.claim_one(db, "details", "worker-a")
refute job.run_at == job.created_at
```

RAN: gives `run_at=1788316988`, `created_at=1788317088`, and the job still claims. This hole is now mutation-proven in all four bindings (#124, #143, #148, #149) — it is a property of the shared test design, not of any one PR.

### 2. The `setup` skip is not a valid ExUnit return — it hard-fails — **RAN**

`setup` returns `{:skip, "honker extension not built"}` when the extension is missing. ExUnit does not accept that:

```
** (RuntimeError) expected ExUnit setup callback in HonkerJobFieldsTest to return
   the atom :ok, a keyword, or a map, got {:skip, "honker extension not built"} instead

6 tests, 6 failures
```

So the intended graceful skip is dead code; without a built extension the suite fails with a misleading error rather than the intended message. Failing loudly is arguably the better behaviour, but this is not the behaviour the code is written to produce. In CI it never triggers (the extension is built first), so this is a local-developer trap. If a skip is genuinely wanted, use `@moduletag :skip` or an `ExUnit.Case` tag; if a hard failure is wanted, `raise` with the real message.

Note this is the opposite of #148, where the Ruby skip *does* work and the whole spec passes with exit code 0.

### 3. `Queue.get_job/2` takes no queue, so #134 will force an arity change — **RAN + INFERRED**

RAN — the lookup is global today, matching the docstring:

```
get_job(sms_id).queue = "sms"   # read from a db handle, no queue arg exists
cancel(sms_id)        = {:ok, true}
```

INFERRED — every other function on `Honker.Queue` takes a queue name (`enqueue/4`, `sweep_expired/2`, `claim_batch/4`, `claim_one/3`), while `get_job/2` and `cancel/2` do not. Node has already made `getJob` queue-scoped in #124. When #134 lands, Elixir cannot scope `get_job` without becoming `get_job/3` — a breaking arity change on a module function, which is harsher than in Ruby or Python where the method already hangs off a queue object.

That also makes the new docstring — "Ids are globally unique but not scoped to a queue, so a foreign id returns that queue's row (#134)" — accurate today but scheduled to become wrong. Worth deciding the shape now rather than documenting behaviour that #134 will invert. At least `get_job` and `cancel` are consistently global here, which is better than Node's split (scoped read, global delete).

### What holds up — **RAN**

All in-repo callers are updated: `phase_mantle_test.exs` was correctly migrated from `row["queue"]` to `row.queue`, and a repo-wide grep finds no other consumer (the ORM proof scripts call `honker_get_job()` in raw SQL and are unaffected). Full suite is green at 74 tests. Typespecs on both structs correctly mark `worker_id`/`claim_expires_at` as nullable on the snapshot but not on `Job`, and `expires_at` as nullable on both. The `JobSnapshot` doc names the raw-text-vs-decoded payload divergence explicitly.

**Recommendation:** merge once the `run_at` assertion is tightened and the `{:skip, ...}` return is fixed. Field mapping and caller migration are both correct.

russellromney:
## Findings 1 and 2 fixed

Two atomic commits:

- `560bbca` — `test(elixir): make the missing-extension skip a real ExUnit skip`
- `dd3e51c` — `test(elixir): prove a claimed job reports run_at, not created_at`

### Finding 1 — the `run_at` / `created_at` gap is now load-bearing

Added `a claimed job reports its own run_at, not its created_at`. It uses `Queue.enqueue(db, "details", ..., run_at: db_now(db) - 100)` — a **back-dated absolute run_at**, not a delay — so the job stays claimable while `run_at != created_at`, and it pins `run_at` to the exact value enqueued rather than comparing it to another field.

**Mutation — claimed `Job` builder (`lib/honker/queue.ex:157/162`)**

```elixir
run_at: row["created_at"],
created_at: row["run_at"],
```

```
  1) test a claimed job reports its own run_at, not its created_at (HonkerJobFieldsTest)
     test/job_fields_test.exs:148
     Assertion with == failed
     code:  assert job.run_at == run_at
     left:  1788319146
     right: 1788319046
     stacktrace:
       test/job_fields_test.exs:159: (test)

7 tests, 1 failure
```

**Only the new test fails** — the other six stay green, which is the blind spot you reported.

I also mutated `JobSnapshot.from_row` (`lib/honker/job_snapshot.ex:70/75`) for completeness: 3 failures, the new test among them (`assert snap.run_at == run_at`, line 153). That site was already covered by the delayed and retry tests; the claimed path was the hole.

Both sites restored; `git diff` on `lib/` is empty.

### Finding 2 — the `{:skip, ...}` return

Made it a real skip, per your suggestion. The path is now resolved at compile time into `@extension_path`, and the guard is the tag ExUnit actually honours:

```elixir
@moduletag skip: @extension_path == nil && "honker extension not built"
```

`skip: false` is not a skip (ExUnit only skips on `true` or a binary), so this fires only when the extension is genuinely absent. `setup` no longer branches.

**Before** — with the extension missing:

```
** (RuntimeError) expected ExUnit setup callback in HonkerJobFieldsTest to return
   the atom :ok, a keyword, or a map, got {:skip, "honker extension not built"} instead

6 tests, 6 failures
```

**After** — same condition, verified by removing the dylib:

```
Running ExUnit with seed: 267616, max_cases: 16

*******
Finished in 0.1 seconds (0.00s async, 0.1s sync)
7 tests, 0 failures, 7 skipped
```

With the extension present: `7 tests, 0 failures`. As you noted, CI never triggers this — the build step runs first — so this was a local-developer trap.

One note for the record: unlike Ruby, I did **not** make this fail loudly under a required-extension env flag. Elixir has no equivalent to `HONKER_REQUIRE_RUBY_EXTENSION_LOADING` in this repo, and inventing one felt like scope creep for a test-coverage PR. The CI step does set `HONKER_EXTENSION_PATH`, which `find_extension` ignores in favour of repo-relative paths — worth reconciling separately if you want the same belt-and-braces guarantee Ruby now has.

### Local run

Full suite: **75 tests, 0 failures** (was 74). The `honker-update-poll` panic in the output is the deliberate database-file-replacement test, unchanged.

Finding 3 (`get_job/2` arity under #134) is untouched — that is a design decision, not a review defect.

russellromney:
## Second adversarial review — #149

Labelled **RAN** (executed here) or **READ** (from source/diff, not executed). Extension: the debug `libhonker_ext.dylib` already built from a branch carrying #124's widened claim — confirmed by `strings | grep 'RETURNING id, queue, payload, state, priority, run_at, worker_id'`. Baseline before my changes: **7 tests, 0 failures**; full suite **75 tests, 0 failures**.

Five commits pushed to this branch. Verdict at the bottom.

---

### 1. A claimed job's `id` was indistinguishable from its `attempts` — **RAN** — fixed in `c1dbcd3`

Same shape as the `run_at`/`created_at` hole from last round, and it was still open. Every test opens a fresh database, so the first job has `id == 1`, and once claimed `attempts == 1` too. `assert job.id == id` could not tell the columns apart.

```elixir
id: row["attempts"],   # in Queue.row_to_job/1
```

→ **7 tests, 0 failures.** Survived everything, including `Job.ack/2` in "a processing snapshot matches the claim, then misses after ack" — the ack still targeted the right row because the wrong column held the right number.

Fix: spend 20 row ids in an unrelated queue during setup, so the first id a test sees is 21 — clear of `attempts` (0/1), `priority` (7), `max_attempts` (5), and the visibility timeout (11). Ids are unique across the whole database; nothing ever claims from that queue. With it applied the same mutation fails 4 tests:

```
a claimed job carries every field
a claimed job reports its own run_at, not its created_at
a processing snapshot matches the claim, then misses after ack
retry keeps attempts and pushes run_at out
```

**Full pairwise sweep, after the fix — RAN.** Every field swapped for every type-compatible sibling in both builders (`Queue.row_to_job/1` and `JobSnapshot.from_row/1`): **136 mutations, 0 survivors.** The field mapping is now genuinely pinned in both directions. Harness kept out of the repo.

This is the third binding where the blind spot is "two fields that coincide in a fresh database". Worth promoting to the repo's CLAUDE.md as a named trap: *a fresh-database test cannot distinguish `id`, `attempts`, or any small integer from any other small integer — burn ids and use distinct constants.*

### 2. `expires_at` nil-vs-0 was only proven on the snapshot path — **RAN** — fixed in `69541d0`

`JobSnapshot.from_row/1` and `Queue.row_to_job/1` are separate builders on separate ABI calls, and every claim test enqueued with `:expires`, so the claimed path was unchecked.

```elixir
expires_at: row["expires_at"] || 0,   # in Queue.row_to_job/1
```

→ passed before, fails exactly one test after. The test now also asserts `claim_expires_at != nil` so it cannot pass by every timestamp being nil.

### 3. The break is not documented anywhere a Hex user looks — **RAN** — fixed in `4b84f25`

`mix hex.build` on this branch:

```
lib/…, mix.exs, README.md, LICENSE, LICENSE-APACHE, LICENSE-MIT
```

No CHANGELOG — it lives at the repo root, and `package()` sets no `:files`, so Hex's default list (which is relative to `packages/honker-ex/`) never picks it up. mix.exs also declares no `docs:` config, so ex_doc renders module pages only: no README extras page either. The `{:ok, map}` → `{:ok, %JobSnapshot{}}` break was documented in exactly one place a Hex user cannot reach.

Fixed by putting it in `get_job/2`'s own `@doc` (which *is* hexdocs) and in the README, and by naming the failure mode: a struct has no `Access` behaviour, so `row["state"]` **raises `UndefinedFunctionError`** rather than returning nil. That is the good case — it fails loudly — but the compiler will not catch it, so callers have to be grepped.

**On the Ruby comparison — RAN.** Elixir does have the escape hatch. A struct that implements the `Access` behaviour with a string-key `fetch/2` keeps `row["state"]` working, exactly as Ruby's `Struct` did:

```
plain raised: function Plain.fetch/2 is undefined (Plain does not implement the Access behaviour…)
Access struct string-key: "pending"
Access struct atom-key:   "pending"
Access struct dot:        "pending"
```

So "Ruby's source-compatibility is a property of Ruby, not a bar Elixir has to clear" is not quite right — Elixir could have cleared it. Not implementing it: the decision to take the break is recorded above and I agree with it. `Access` on a struct is discouraged in modern Elixir precisely because it hides typos behind nil, and a raise beats a silent nil for a migration. Recording the option so the combined migration guide can say the break was chosen, not forced.

### 4. Two moduledoc statements that are wrong or scheduled to become wrong — **READ** — fixed in `4b84f25`

- `Honker.Queue`'s moduledoc: *"All methods take a `Honker.Database` and the queue name as the first two args."* Flatly false for `get_job/2` and `cancel/2`. This predates the PR but it is the first thing a user reads on the module the PR reshapes. Now names the two exceptions.
- `get_job/2`: *"Ids are globally unique but not scoped to a queue, so a foreign id returns that queue's row (#134)."* Accurate today, and stated as a property a caller may rely on. Softened to describe today's behaviour, tell the reader not to build on it, and point at `snapshot.queue`. Also dropped the bare `#134` — issue numbers mean nothing on hexdocs. (I did **not** implement scoping; `feat/cancel-queue-scoped` is already adding the core side, so the wording would have inverted.)
- `Honker.Job`'s moduledoc said the struct "holds a snapshot of the row", which now collides with the type literally named `JobSnapshot`. Reworded, and it now points at `get_job/2` for the read-only case.

### 5. The new `@type t`s had nothing pointing at them — **READ** — fixed in `4a1387f`

`Honker.Job.t/0` and `Honker.JobSnapshot.t/0` were added, but no function spec referenced either, so dialyzer had no contract to check and ex_doc had no signature to render — they were comments in type syntax. Added `@spec` to the three functions that hand a caller one of these structs: `claim_batch/4`, `claim_one/3`, `get_job/2`. The `nil` arm on the two singular reads is now in the signature, which is the thing a first-time reader most needs.

### 6. Ecto users had a public `from_row/1` with no path to it — **READ** — fixed in `d4591d0`

`Honker.Extension`'s moduledoc steers Ecto users to load Honker onto a connection they own and call the SQL directly (the only way to keep an enqueue inside their transaction — and what `scripts/proof/orm/elixir/ecto_proof.exs` does). Those users never call `Queue.get_job/2`, so `JobSnapshot.from_row/1` being public was useful but undiscoverable. Added the two-line example and the empty-string-on-miss caveat.

---

### What holds up — **RAN + READ**

- **All twelve fields, no thirteenth.** `honker_ops.rs` `get_job` and the claim `RETURNING` both emit exactly `id, queue, payload, state, priority, run_at, worker_id, claim_expires_at, attempts, max_attempts, created_at, expires_at`. Nothing is dropped and nothing is invented.
- **The `@moduletag` skip guard is correct — RAN.** Verified both arms standalone: `skip: false` runs the test, `skip: "reason"` skips it. Last round's fix is right, and the predecessor really is rejected (`expected ExUnit setup callback … to return the atom :ok, a keyword, or a map`).
- Nil handling: `worker_id` / `claim_expires_at` / `expires_at` are genuinely `nil`, on both builders, now mutation-proven on both.
- Caller migration and the full suite: **75 tests, 0 failures** on the final tree.

### Left deliberately

- **The same broken `{:skip, reason}` is still in `phase_mantle_test.exs`, `smoke_test.exs`, and `parity_test.exs`** — pre-existing, unrelated to #136, so out of scope for this PR. Consequence: a local dev without the extension now sees `job_fields_test` skip gracefully while three other files hard-fail with the misleading merge error. One-line fix each, same shape as `560bbca`; say the word and I will do it as a separate commit or PR.
- **mix.exs ships no CHANGELOG and no ex_doc `extras`.** Root cause of finding 3. Fixing it means touching release packaging on a stacked PR, which is a release-time decision, not this one's.
- **Snapshot payload stays raw JSON text** while the claimed job's is decoded (#146), **no queue scoping** (#134), **no `claimed_at`** (#135) — per the brief. The split is documented in the moduledoc and the README and is now pinned by assertions on both sides, so whichever way #146 lands, a test will notice.

### Verdict

**Merge**, after #124. Field mapping is correct and now exhaustively mutation-proven; the break is documented where it will actually be read; nothing in the diff touches honker-core or another binding.

PR #150, opened 2026-09-01
feat(rust): full job details and typed queue payloads (#136)

russellromney:
Part of #136 (Rust binding checklist item).

## Stacked on #124

**This PR targets `feat/node-typed-jobs` (#124), not `main`, and must merge after #124.** It depends on the widened `honker_claim_batch` RETURNING clause that exists only on that branch. `honker-core` is unchanged here.

## What changed

`packages/honker-rs/src/lib.rs`:

- **All twelve fields on claimed jobs.** `Job` gained `state`, `priority`, `run_at`, `claim_expires_at`, `max_attempts`, `created_at`, `expires_at` alongside the existing `id`, `queue`, `payload`, `worker_id`, `attempts`. `JobRow` (the read-only snapshot from `get_job`) already had all twelve and keeps them.
- `Job` keeps `ack` / `retry` / `fail` / `heartbeat`. `JobRow` is data only. On `Job`, `worker_id` and `claim_expires_at` are non-optional (holding a claim is what makes it a `Job`); on `JobRow` they are `Option` because a pending row has neither.
- **One raw wire struct.** `RawJob` now decodes the full twelve-key snake_case shape that `honker_claim_batch()` and `honker_get_job()` both return, with `into_job()` / `into_row()` conversions. A claimed row that comes back without a `worker_id` or `claim_expires_at` is now an error, not a silent default.
- **Typed payloads.** `Queue`, `Job`, `JobRow`, `ClaimWaker`, and `Outbox` carry a payload type parameter defaulting to `serde_json::Value`.
  - `db.queue(..)` and `db.outbox(..)` still return the `serde_json::Value` handles, so existing call sites are untouched (every test and example in the repo compiled unmodified).
  - `db.typed_queue::<T>(..)`, `db.typed_outbox::<T>(..)`, and `queue.typed::<T>()` give typed handles.
  - `payload_typed()` decodes into the handle's `T`; the existing `payload_as::<U>()` still decodes into anything you name.
  - `Database::queue` could not itself be made generic with a default: Rust does not allow default type parameters on functions, and type-parameter defaults do not participate in function-return inference, so a generic `queue()` would break every `let q = db.queue(..)`. Hence the separate `typed_queue` constructor.

## Payload encoding (please read: cross-binding divergence)

Per the coordination note, snapshot payload encoding is **unchanged** in this PR:

| binding | claimed job payload | snapshot (`getJob`/`get_job`) payload |
| --- | --- | --- |
| Rust (this PR) | raw JSON bytes (`Vec<u8>`) | raw JSON text (`String`) |
| Node (#124) | decoded value | **decoded value** |
| Go | raw JSON | raw JSON text |
| Python | raw JSON | raw JSON text |

Three bindings, two conventions. Rust hands back raw text on both paths and adds `payload_typed()` / `payload_as::<U>()` as the decode step; it does **not** silently start decoding to make the generic tidier. Worth one deliberate decision across all bindings later. Flagging, not resolving.

## No runtime payload validation

Documented in the crate docs, the README, and asserted by a test: the type parameter is a compile-time convenience for the caller's code. honker stores the payload as opaque JSON and never checks its shape. A mismatch surfaces only as a `serde` error at `payload_typed()` time.

## Source-breaking change

`Queue::enqueue` / `enqueue_tx` and `Outbox::enqueue` / `enqueue_tx` now take `&T` rather than any `&impl Serialize`, and `Outbox::run_once`'s delivery closure takes `T` rather than `serde_json::Value`. On the default handles `T` is `serde_json::Value`, so `q.enqueue(&json!(..))` is unaffected: every existing test, example, and doc example compiled without edits. A caller who passed a custom struct to an untyped queue now takes a typed handle. Noted in `CHANGELOG.md`.

## Tests

New `packages/honker-rs/tests/job_details.rs`, 7 tests, all passing:

- `claimed_job_carries_every_core_field`
- `claimed_job_attempts_counts_up_across_retries`
- `snapshot_shows_pending_then_processing_then_misses`
- `delayed_job_snapshot_reports_the_future_run_at`
- `typed_queue_round_trips_the_payload_type`
- `typed_queue_reads_a_payload_another_writer_produced`
- `payload_shape_is_never_validated_by_honker`

Every field is asserted against a value that could not be produced by another field. The fixture deliberately avoids defaults and same-second coincidences: visibility timeout 137s (not 300), `max_attempts` 9 (not 3), priorities 42 and 7 on two rows so `priority` cannot be a constant, `run_at` back-dated 300s so it is never equal to `created_at`, `expires_at` set on one row and null on the other, and `claim_expires_at` bracketed against the claim instant plus 137s.

The full existing suite (`cargo test`, 57 tests across `basic`, `surface`, `phase_mantle`, `python_interop`, doc tests) passes unmodified.

### Proof the tests are load-bearing

Two mutations, each reverted after:

**1. `run_at` sourced from `created_at`** in `RawJob::into_job`. `claimed_job_carries_every_core_field` fails:

```
thread 'claimed_job_carries_every_core_field' panicked at tests/job_details.rs:112:5:
assertion `left == right` failed: run_at
  left: 1788256131
 right: 1788255831
```

The 300-second gap is the deliberate `run_at` back-date.

**2. `created_at` sourced from `run_at`** in `RawJob::into_row`. `delayed_job_snapshot_reports_the_future_run_at` fails:

```
thread 'delayed_job_snapshot_reports_the_future_run_at' panicked at tests/job_details.rs:290:5:
assertion `left == right` failed: run_at must sit exactly 600s after created_at
  left: 0
 right: 600
```

This one is the interesting case: `snapshot_shows_pending_then_processing_then_misses` **passed** under that mutation, because an undelayed job legitimately has `run_at == created_at`. That is exactly the coincidence-equality false pass to watch for, and the delayed-job test is what discriminates it.

## Not in scope

- `claimed_at`: that is #135 / PR #140 on another branch, deliberately not added here.
- `honker-core` and the SQL ABI are untouched.
- Queue-scoped `get_job`: #134.
- No other binding touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,14 @@
   `db.typed_outbox::<T>()`, and `Queue::typed::<T>()` produce typed
   handles; `payload_typed()` decodes into the handle's `T` and the
   existing `payload_as::<U>()` still decodes into anything you name.
+  A default type parameter cannot go on a function, so `Database::queue`
+  stays non-generic (otherwise `let q = db.queue(..)` would stop
+  inferring) and `typed_queue` is a separate constructor. The cost is
+  that `db.queue::<Email>(..)` is `error[E0107]: method takes 0 generic
+  arguments` — with a "remove the unnecessary generics" hint pointing
+  the wrong way — and `let q: Queue<Email> = db.queue(..)` is an E0308.
+  Neither message mentions `typed_queue`, so `Database::queue`'s and
+  `Queue`'s rustdoc now say where to go.
 - Payload encoding is unchanged: `Job::payload` stays raw JSON bytes and
   `JobRow::payload` stays a raw JSON string. The type parameter is
   compile-time only — honker never validates payload shape in the
@@ -50,12 +58,21 @@
   - Two different payload types through one handle — you now need one
     typed handle per type, or keep the handle at `serde_json::Value` and
     `serde_json::to_value(..)` at the call site.
-  - **A generic helper has no drop-in migration.**
-    `fn helper<P: Serialize>(q: &Queue, p: &P)` no longer compiles; the
-    signature must become `fn helper<P: Serialize>(q: &Queue<P>, p: &P)`,
-    which changes every call site of the helper too, not just the body.
-    If you cannot change the call sites, keep `q: &Queue` and pass
-    `&serde_json::to_value(p)?` instead.
+  - A generic helper, `fn helper<P: Serialize>(q: &Queue, p: &P)`, no
+    longer compiles as written, but the migration is body-only —
+    `Queue::typed` reinterprets the handle in place, so the signature
+    and every call site stay as they are:
+    ```rust
+    fn helper<P: Serialize>(q: &Queue, p: &P) -> honker::Result<i64> {
+        q.typed::<P>().enqueue(p, EnqueueOpts::default())
+    }
+    ```
+    `typed()` clones the handle (an `Arc` and the queue name); it does
+    not touch the database. Widening the signature to
+    `fn helper<P: Serialize>(q: &Queue<P>, p: &P)` also works but
+    changes the helper's call sites too, and
+    `q.enqueue(&serde_json::to_value(p)?, ..)` works at the cost of an
+    intermediate `Value`. Prefer `typed::<P>()`.
 - **Breaking (source), 2 of 2 — `JobRow` is no longer constructible with
   struct literal syntax.** `JobRow<T>` needs a private `PhantomData`
   field to carry `T`, and a private field disables struct literals
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,32 @@
 - Regression test claims a job at a deadline with one claim forced to
   return empty: it fails on the unpatched `api.js` and passes on the fix.
 
+## Unreleased — typed Rust jobs
+
+- `honker` (Rust): claimed `Job` and the `JobRow` snapshot now carry all
+  twelve fields core returns — `id`, `queue`, `payload`, `state`,
+  `priority`, `run_at`, `worker_id`, `claim_expires_at`, `attempts`,
+  `max_attempts`, `created_at`, `expires_at`. `Job` keeps `ack`, `retry`,
+  `fail`, `heartbeat`; `JobRow` stays data only.
+- `Queue`, `Job`, `JobRow`, `ClaimWaker`, and `Outbox` gained a payload
+  type parameter defaulting to `serde_json::Value`, so `db.queue(..)`
+  and `db.outbox(..)` behave exactly as before. `db.typed_queue::<T>()`,
+  `db.typed_outbox::<T>()`, and `Queue::typed::<T>()` produce typed
+  handles; `payload_typed()` decodes into the handle's `T` and the
+  existing `payload_as::<U>()` still decodes into anything you name.
+- Payload encoding is unchanged: `Job::payload` stays raw JSON bytes and
+  `JobRow::payload` stays a raw JSON string. The type parameter is
+  compile-time only — honker never validates payload shape in the
+  database, and a mismatch surfaces only as a `serde` error at decode.
+- **Breaking (source):** `Queue::enqueue` / `enqueue_tx` and
+  `Outbox::enqueue` / `enqueue_tx` now take `&T` instead of any
+  `&impl Serialize`, and `Outbox::run_once`'s delivery closure takes `T`
+  instead of `serde_json::Value`. On the default handles `T` is
+  `serde_json::Value`, so `q.enqueue(&json!(..))` is untouched; callers
+  who passed a custom struct to an untyped queue now take a typed handle.
+- A claimed row that comes back without a `worker_id` or
+  `claim_expires_at` is now an error instead of a silent default.
+
 ## Unreleased — typed Node jobs
 
 - Node `Queue`, `Job`, `JobSnapshot`, `ClaimWaker`, and `Outbox` APIs now carry
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,12 +37,52 @@
   `JobRow::payload` stays a raw JSON string. The type parameter is
   compile-time only — honker never validates payload shape in the
   database, and a mismatch surfaces only as a `serde` error at decode.
-- **Breaking (source):** `Queue::enqueue` / `enqueue_tx` and
-  `Outbox::enqueue` / `enqueue_tx` now take `&T` instead of any
-  `&impl Serialize`, and `Outbox::run_once`'s delivery closure takes `T`
-  instead of `serde_json::Value`. On the default handles `T` is
-  `serde_json::Value`, so `q.enqueue(&json!(..))` is untouched; callers
-  who passed a custom struct to an untyped queue now take a typed handle.
+- **Breaking (source), 1 of 2 — `enqueue` takes `&T`.**
+  `Queue::enqueue` / `enqueue_tx` and `Outbox::enqueue` / `enqueue_tx`
+  now take `&T` instead of any `&impl Serialize`, and
+  `Outbox::run_once`'s delivery closure takes `T` instead of
+  `serde_json::Value`. On the default handles `T` is `serde_json::Value`,
+  so `q.enqueue(&json!(..))` is untouched. Everything else is
+  `error[E0308]: mismatched types / expected &Value, found &X`:
+  - `q.enqueue(&MyStruct { .. })`, `q.enqueue(&"hello")`,
+    `q.enqueue(&vec![1u8, 2, 3])` — migrate the handle:
+    `db.typed_queue::<MyStruct>(..)` or `q.typed::<MyStruct>()`.
+  - Two different payload types through one handle — you now need one
+    typed handle per type, or keep the handle at `serde_json::Value` and
+    `serde_json::to_value(..)` at the call site.
+  - **A generic helper has no drop-in migration.**
+    `fn helper<P: Serialize>(q: &Queue, p: &P)` no longer compiles; the
+    signature must become `fn helper<P: Serialize>(q: &Queue<P>, p: &P)`,
+    which changes every call site of the helper too, not just the body.
+    If you cannot change the call sites, keep `q: &Queue` and pass
+    `&serde_json::to_value(p)?` instead.
+- **Breaking (source), 2 of 2 — `JobRow` is no longer constructible with
+  struct literal syntax.** `JobRow<T>` needs a private `PhantomData`
+  field to carry `T`, and a private field disables struct literals
+  outside the crate:
+  ```
+  error: cannot construct `JobRow<_>` with struct literal syntax due to private fields
+    = note: ...and other private field `_payload` that was not provided
+  ```
+  There is no form of the type parameter that keeps struct literals
+  working, so this one is kept, not fixed. Build the row through
+  `Deserialize` instead — `serde_json::from_value(json!({ .. }))` or
+  `serde_json::from_str(..)` with the twelve snake_case keys — which is
+  what a fake in your own tests should do anyway.
+- **Not breaking after all:** `JobRow` keeps `Debug`, `Clone`, and
+  `Deserialize`. The generic briefly dropped all three; `Debug` and
+  `Clone` are hand-written (so they no longer require `T: Debug` /
+  `T: Clone`), and `Deserialize` is re-derived with `#[serde(bound = "")]`
+  plus `#[serde(skip)]` on the phantom field. Deserializing a raw
+  `honker_get_job()` blob straight into `JobRow` still compiles and is
+  covered by a test.
+- **Not breaking after all:** `Outbox::new(&db, name, opts)` still infers
+  its payload type. Making `Outbox` generic had moved `new` onto
+  `impl<T>`, which broke a bare `let ob = Outbox::new(..)` with
+  `error[E0282]: type annotations needed for Outbox<_>`. `new` is back on
+  an inherent `impl Outbox<serde_json::Value>`; the generic constructor
+  is the new `Outbox::new_typed`, which is what `Database::typed_outbox`
+  calls.
 - A claimed row that comes back without a `worker_id` or
   `claim_expires_at` is now an error instead of a silent default.
 
--- a/packages/honker-rs/README.md
+++ b/packages/honker-rs/README.md
@@ -27,6 +27,52 @@ if let Some(job) = q.claim_one("worker-1")? {
 
 Delayed jobs use `run_at` / `RunAt`-style options in the binding API.
 
+## Job details
+
+`Job` (a claimed unit of work) and `JobRow` (the read-only snapshot from
+`Queue::get_job`) both carry every field the core claim/lookup returns:
+`id`, `queue`, `payload`, `state`, `priority`, `run_at`, `worker_id`,
+`claim_expires_at`, `attempts`, `max_attempts`, `created_at`, and
+`expires_at`.
+
+`Job` also holds a claim, so it has `ack`, `retry`, `fail`, and
+`heartbeat`, and its `worker_id` / `claim_expires_at` are non-optional.
+`JobRow` is data only: no claim methods, and `worker_id` /
+`claim_expires_at` are `Option` because a pending row has neither.
+
+Payload encoding is unchanged: `Job::payload` is `Vec<u8>` of raw JSON
+bytes and `JobRow::payload` is the raw JSON `String`.
+
+## Typed payloads
+
+`Queue<T>` carries the payload type. `db.queue(..)` still hands back a
+`Queue<serde_json::Value>`; `db.typed_queue::<T>(..)` (or
+`queue.typed::<T>()`) gives you your own type:
+
+```rust
+#[derive(serde::Serialize, serde::Deserialize)]
+struct Email { to: String }
+
+let q = db.typed_queue::<Email>("emails", QueueOpts::default());
+q.enqueue(&Email { to: "alice@example.com".into() }, EnqueueOpts::default())?;
+
+if let Some(job) = q.claim_one("worker-1")? {
+    let email: Email = job.payload_typed()?;
+    job.ack()?;
+}
+```
+
+`payload_typed()` decodes into the queue's `T`; `payload_as::<U>()`
+decodes into anything you name. Both are plain `serde` deserialization.
+
+**honker never checks payload shape.** The type parameter is a
+compile-time convenience for your code only. The database stores the
+payload as opaque JSON text and nothing on the write path validates it,
+so two processes writing the same queue with different types will
+produce rows the other cannot decode. The only error you get is a
+`serde` failure at decode time. Keeping producers and consumers in
+agreement is your job.
+
 Recurring schedules use schedule expressions:
 
 ```rust
--- a/packages/honker-rs/README.md
+++ b/packages/honker-rs/README.md
@@ -65,6 +65,23 @@ if let Some(job) = q.claim_one("worker-1")? {
 `payload_typed()` decodes into the queue's `T`; `payload_as::<U>()`
 decodes into anything you name. Both are plain `serde` deserialization.
 
+There is no `db.queue::<Email>(..)`. Rust cannot put a default on a
+function's type parameter, so making `queue` generic would break every
+bare `let q = db.queue(..)`; `typed_queue` is a separate constructor
+instead. `db.queue::<Email>(..)` therefore fails with `error[E0107]:
+method takes 0 generic arguments` and the compiler's "remove the
+unnecessary generics" hint points the wrong way — reach for
+`db.typed_queue::<Email>(..)` or `q.typed::<Email>()`.
+
+A generic helper that used to take a plain `&Queue` keeps its
+signature — and so its call sites — by reinterpreting inside the body:
+
+```rust
+fn helper<P: serde::Serialize>(q: &honker::Queue, p: &P) -> honker::Result<i64> {
+    q.typed::<P>().enqueue(p, EnqueueOpts::default())
+}
+```
+
 **honker never checks payload shape.** The type parameter is a
 compile-time convenience for your code only. The database stores the
 payload as opaque JSON text and nothing on the write path validates it,
--- a/packages/honker-rs/README.md
+++ b/packages/honker-rs/README.md
@@ -9,21 +9,31 @@ Full docs:
 
 ## Install
 
-Add the crate, and make sure the Honker SQLite extension is available at runtime.
+Add the crate. `honker-core` is linked in and bootstraps the schema, so
+there is no `.dylib` to load at runtime. Use the loadable extension
+(`honker-extension`) only when other SQLite clients share the file.
 
 ## Quick start
 
 ```rust
-let db = honker::Database::open("app.db", "./libhonker_ext.dylib")?;
-let q = db.queue("emails");
+use honker::{Database, EnqueueOpts, QueueOpts};
+use serde_json::json;
 
-q.enqueue(serde_json::json!({ "to": "alice@example.com" }))?;
+let db = Database::open("app.db")?;
+let q = db.queue("emails", QueueOpts::default());
+
+q.enqueue(&json!({ "to": "alice@example.com" }), EnqueueOpts::default())?;
 
 if let Some(job) = q.claim_one("worker-1")? {
-    send_email(&job.payload);
+    let body: serde_json::Value = job.payload_as()?;
+    send_email(&body);
     job.ack()?;
 }
 ```
+
+Runnable versions of this live in
+[`examples/basic.rs`](examples/basic.rs) and
+[`examples/atomic.rs`](examples/atomic.rs), which CI compiles.
 
 Delayed jobs use `run_at` / `RunAt`-style options in the binding API.
 
@@ -93,8 +103,17 @@ agreement is your job.
 Recurring schedules use schedule expressions:
 
 ```rust
+use honker::ScheduledTask;
+
 let sched = db.scheduler();
-sched.add("fast", "emails", "@every 1s", serde_json::json!({ "kind": "tick" }))?;
+sched.add(ScheduledTask {
+    name: "fast".into(),
+    queue: "emails".into(),
+    schedule: "@every 1s".into(),
+    payload: json!({ "kind": "tick" }),
+    priority: 0,
+    expires_s: None,
+})?;
 ```
 
 Supported schedule forms:
--- a/packages/honker-rs/src/lib.rs
+++ b/packages/honker-rs/src/lib.rs
@@ -864,6 +864,8 @@ impl RawJob {
 ///
 /// `payload` stays the raw JSON text exactly as stored. Use
 /// [`JobRow::payload_typed`] to decode it into the queue's `T`.
+#[derive(Deserialize)]
+#[serde(bound = "")]
 pub struct JobRow<T = serde_json::Value> {
     pub id: i64,
     pub queue: String,
@@ -879,6 +881,7 @@ pub struct JobRow<T = serde_json::Value> {
     pub max_attempts: i64,
     pub created_at: i64,
     pub expires_at: Option<i64>,
+    #[serde(skip)]
     _payload: PhantomData<fn(T) -> T>,
 }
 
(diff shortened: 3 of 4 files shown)


russellromney:
Adversarial review (I did not write this code). Everything below was RUN unless marked INFERRED, in a detached worktree at `38246ba`.

Baseline: the exact CI command (`cargo test --locked --manifest-path packages/honker-rs/Cargo.toml --features kernel-watcher,shm-fast-path`) → **55 tests, 0 failed**, `job_details` included (7 tests). No skip path exists — the crate links `honker-core` directly and never loads the `.dylib` — so a green run is real. `cargo test --no-run` builds all 5 test binaries unmodified, `cargo build --examples` is clean, `cargo test --doc` → 2 passed.

### 1. HIGH — two undisclosed source breaks on the public `JobRow`. RAN, compiled both sides.

`JobRow` lost `#[derive(Debug, Deserialize, Clone)]` and gained a private `_payload: PhantomData<fn(T) -> T>` while all its other fields stay `pub`. `Debug` and `Clone` are hand-written back (lib.rs:901, 921); `Deserialize` is not, and the private field makes the struct non-constructible. The CHANGELOG discloses only the `enqueue(&T)` / `run_once` change.

Same probe, `cargo check --tests`, base vs head. Base: `Finished dev profile`. Head:

```
error[E0277]: the trait bound `JobRow: serde::Deserialize<'de>` is not satisfied
 --> tests/zz_break_probe.rs:5:32
  = note: for local types consider adding `#[derive(serde::Deserialize)]` to your `JobRow` type

error: cannot construct `JobRow<_>` with struct literal syntax due to private fields
 --> tests/zz_break_probe.rs:9:16
  = note: ...and other private field `_payload` that was not provided
```

Who hits it: anyone who deserialized a raw `honker_get_job()` blob into `JobRow` themselves, or built one in their own tests or fakes. `Job` is unaffected — it already had a private `inner: Arc<Inner>` and no derives, so it was never externally constructible. Either re-derive `Deserialize` (via a `#[serde(skip)]` on `_payload`) or add these two to the breaking-change note.

### 2. MEDIUM — `Outbox::new` inference regression, also undisclosed. RAN.

`Outbox::new` moved from an inherent impl on the concrete type to `impl<T> Outbox<T>`, so a direct call with nothing to pin `T` stops compiling:

```
error[E0282]: type annotations needed for `Outbox<_>`
 7 |     let ob = Outbox::new(&db, "direct", OutboxOpts::default());
   |         ^^
help: consider giving `ob` an explicit type, where the type for type parameter `T` is specified
```
`Database::outbox(..)` is unaffected. This is the only *inference* regression found — `db.queue("x")` followed by `enqueue(&json!(..))` still infers cleanly.

### 3. MEDIUM — typed `Queue<T>` plus a global `get_job` is unsound, and the silent case is the dangerous one. RAN.

`Queue<T>::get_job` deserializes any queue's row into `T`:
```
SILENT WRONG-SHAPE DECODE: Email { to: "a@x", subject: "inv" } from queue billing
HARD ERROR CASE:           json error: missing field `to` at line 1 column 10
```
`db.typed_queue::<Email>("emails").get_job(<a billing id>)` returns `Some(JobRow<Email>)` with `row.queue == "billing"`, and because the other queue's payload is a structural superset, `payload_typed()` **succeeds with the wrong shape**. That is precisely the reason `wrapper.d.ts` gives for scoping `Queue.getJob`. #150 did not create the global lookup, but it added the `T` that makes it unsound — the two changes want to land together.

### 4. MEDIUM — four field wirings the new tests cannot catch: constants, not swaps. RAN.

The fixture header claims "Every value asserted here is chosen to be distinguishable from the others." That holds against *swaps* — and it genuinely does, see below — but not against *hardcoded constants* on the snapshot path, because every `get_job` assertion in the file uses the same `parity_queue` fixture (`priority == 42`, `max_attempts == 9`, `queue == "details"`).

| Mutation | Result |
|---|---|
| `into_job` `state:` → `"processing".to_string()` | **STILL PASSES** |
| `into_row` `priority:` → `42` | **STILL PASSES** |
| `into_row` `max_attempts:` → `9` | **STILL PASSES** |
| `into_row` `queue:` → `"details".to_string()` | **STILL PASSES** |

Confirmed independently by reading: every snapshot `priority` assertion is `HIGH_PRIORITY` (lines 200, 250), every snapshot `max_attempts` assertion is `MAX_ATTEMPTS` (212, 252), the only `.queue` assertions are `"details"` (94, 194), and every claimed-job `state` assertion is `"processing"`. The two typed-queue tests do call `get_job` on a `QueueOpts::default()` queue but assert only the payload. Line 348 (`job.max_attempts == 3` on a default queue) is what saves the *claimed-job* max_attempts wiring — the snapshot has no equivalent.

Fix is one line each: in `typed_queue_round_trips_the_payload_type`, assert `snapshot.max_attempts == 3` and `snapshot.queue == "typed"`; and read back the `LOW_PRIORITY` row's snapshot in the first test.

### 5. LOW — `Queue::get_job`'s new rustdoc documents a contract #134 will invert.

`lib.rs:731-732`: *"NOT queue-scoped: job ids are globally unique and this lookup hits any queue's row. Scoping is tracked in #134."* Leaving it global is expected; committing the global contract to rustdoc means #134 has to remember to delete it. No **test** asserts the global behavior — every `get_job` call in `tests/` and `examples/` uses an id from the queue that enqueued it — and the signature does not change under scoping (the filter belongs in core's `honker_get_job`), so there is no real conflict, only the doc.

### The source-breaking change, fully characterized (attack A)

All breaks are E0308 "expected `&Value`, found `&X`", on a **default** `db.queue(..)` handle:

1. `q.enqueue(&MyStruct{..})` — found `&Email`
2. one handle, two `Serialize` types — found `&Email`, then `&Sms`
3. `q.enqueue(&"hello")` — found `&&'static str`
4. `q.enqueue(&vec![1u8,2,3])` — found `&Vec<u8>`
5. `fn helper<P: Serialize>(q: &Queue, p: &P) { q.enqueue(p, ..) }` — found `&P`

Migration is `db.typed_queue::<T>(..)` or `q.typed::<T>()` for 1-4. **Case 5 has no drop-in migration**: the helper must become `fn helper<P: Serialize>(q: &Queue<P>, p: &P)`, which changes its call sites too. That is the one worth naming explicitly in the CHANGELOG — "callers who passed a custom struct to an untyped queue now take a typed handle" understates it.

Survives unchanged: `enqueue(&json!(..))`; `Queue` / `Outbox` / `ClaimWaker` / `Job` / `JobRow` named without a type argument; `Vec<Job>`; `payload_as::<U>()`; `run_once` closures with or without a `serde_json::Value` annotation; `db.typed_queue("e", ..)` with `T` inferred from the enqueue argument. Every in-repo call site uses `&json!(..)`, which is why the examples, tests and doctests all still compile.

- **The swap traps are genuinely covered.** 14 wrong-wiring mutations, all caught. The 300 s `run_at` back-date and the `row.run_at - row.created_at == 600` assertion in the delayed test are the load-bearing ones — the "both are about now" pair alone would not catch a run_at/created_at swap, and this fixture knows it. `claim_expires_at` (137 s) vs `expires_at` (9000 s), `attempts` (0/1/2) vs `max_attempts` (9) are genuinely distinct.
- **Field wiring is correct.** `RawJob` deserializes by name, no `#[serde(rename)]`, no positional reads; all twelve names match core's `json!` keys in `honker_ops.rs`. (Core emits `attempts` and `claim_expires_at` in a different order between `claim_batch` and `get_job` — irrelevant, serde matches by name.)
- **Nulls verified by running:** a pending snapshot gives `worker_id == None`, `claim_expires_at == None`, `expires_at == None` — not `Some("")`, not `Some(0)`.
- **`Send + Sync` preserved.** `PhantomData<fn(T) -> T>` is the right choice; compiled `ss::<Queue<Rc<u8>>>()` and equivalents for all four handle types. Existing multithreaded users are unaffected.
- `ClaimWaker::next` and `try_next` both route through `RawJob::into_job`, so the new missing-`worker_id` error is consistent across all three claim paths — though that error path itself has no test, the same gap #151 has.

### Notes, not defects

- `typed_queue_reads_a_payload_another_writer_produced` is two Rust types in one process, not the cross-binding read #136's test list asks for ("one binding writes a versioned payload and another binding reads it"). Fine as a unit test; it does not discharge that checklist item.
- `packages/honker-rs/README.md` "Quick start" (lines 16-24, **pre-existing, untouched here**) does not compile: `Database::open("app.db", "./libhonker_ext.dylib")` takes two args where the real signature takes one, `db.queue("emails")` omits `QueueOpts`, `q.enqueue(json!(..))` passes an owned value with one arg. It is the first code a new user copies. The sections #150 adds directly below it are correct.
- Running the suite prints `honker: data_version poll failed: disk I/O error` to stderr repeatedly (tempdir torn down under the watcher). Pre-existing warn-and-continue in the watcher path.

Would not merge until findings 1 and 2 are either fixed or added to the breaking-change note, and finding 4's four assertions are added. Findings 3 and 5 belong with #134.

Environment note: the machine hit 100% disk mid-review. Separately, another concurrent agent's stdout leaked into one captured shell output and produced one false "STILL PASSES"; every mutation result above was re-verified with `git diff --numstat` proving the edit was applied, and the four in finding 4 were additionally confirmed by reading the assertions. Worktree left clean, still detached at `38246ba`.

russellromney:
Findings 1, 2 and 4 addressed. `66c874f` on `feat/rust-job-parity`. CI green (run [33585970391](https://github.com/russellromney/honker/actions/runs/33585970391), 18 success / 2 skipped bench, `binding smoke · ubuntu aggregate` carries the `Rust wrapper` step). 82 tests, 0 failed.

## The break surface, compiled — not guessed

I wrote a throwaway consumer (`tests/zz_break_probe.rs`, deleted before commit) doing the four things a user actually does: derive-`Deserialize` a type holding a `JobRow`, `serde_json::from_str` a raw `honker_get_job()` blob into one, build one by hand as a test fake, and call `Outbox::new` with no annotation. `cargo check --tests` against `origin/main` and against `38246ba`.

**`origin/main`: `Finished dev profile`, exit 0.** **`38246ba`: 6 errors.** Verbatim:

```
error[E0277]: the trait bound `JobRow: serde::Deserialize<'de>` is not satisfied
  --> tests/zz_break_probe.rs:9:10
   |
 9 |     row: JobRow,
   |          ^^^^^^ the trait `Deserialize<'_>` is not implemented for `JobRow`
   |
   = note: for local types consider adding `#[derive(serde::Deserialize)]` to your `JobRow` type
note: required by a bound in `next_element`
```
(the same E0277 four times over: `SeqAccess::next_element`, `MapAccess::next_value`, `missing_field` from the derive, and `serde_json::from_str`)

```
error: cannot construct `JobRow<_>` with struct literal syntax due to private fields
  --> tests/zz_break_probe.rs:19:5
   |
19 |     JobRow {
   |     ^^^^^^
   |
   = note: ...and other private field `_payload` that was not provided

error[E0282]: type annotations needed for `Outbox<_>`
  --> tests/zz_break_probe.rs:37:9
   |
37 |     let ob = Outbox::new(db, "direct", OutboxOpts::default());
   |         ^^
help: consider giving `ob` an explicit type, where the type for type parameter `T` is specified
   |
37 |     let ob: Outbox<T> = Outbox::new(db, "direct", OutboxOpts::default());
   |           +++++++++++
```

Confirms both of your findings exactly. Per break:

### `Deserialize` — RESTORED (`a51bde5`)

Nothing about the generic requires losing it. Re-derived with `#[serde(bound = "")]` so the phantom `T` does not demand `T: Deserialize`, and `#[serde(skip)]` on `_payload` so it stays out of the wire shape. Losing a derive on a public type in a published crate should be a decision, and there was no reason to make it.

`Debug` and `Clone` stay hand-written rather than re-derived on purpose: a derived `Clone` would require `T: Clone`, and `JobRow<T>` never holds a `T`. The hand-written impls are strictly wider than what `main` had.

### `Outbox::new` inference — RESTORED (`8ca9427`)

`new` is back on an inherent `impl Outbox<serde_json::Value>`, so `let ob = Outbox::new(..)` infers again. The generic constructor is the new `Outbox::new_typed`, which is what `Database::typed_outbox` calls.

### `JobRow` struct-literal construction — KEPT, documented

There is no form of the type parameter that preserves struct literals. A private field kills them (E0451); making `_payload` public only trades E0451 for E0063 "missing field" and buys a permanently ugly public field; `#[non_exhaustive]` is worse. So it is kept.

Restoring `Deserialize` gave it a real migration, which is the part that makes keeping it defensible. The probe's hand-built fake compiles when written as:

```rust
let row: JobRow = serde_json::from_value(json!({
    "id": 1, "queue": "q", "payload": "{}", "state": "pending",
    "priority": 0, "run_at": 0, "worker_id": null, "claim_expires_at": null,
    "attempts": 0, "max_attempts": 3, "created_at": 0, "expires_at": null
})).unwrap();
```

With that one substitution the entire probe compiles clean on this branch. The exact error and this migration are now in the CHANGELOG, and `raw_job_mapping::job_row_still_deserializes_from_a_raw_get_job_blob` guards the derive from regressing.

## Finding 2 — the `enqueue(&T)` break, no longer understated

CHANGELOG now names case 5 as having **no drop-in migration**: `fn helper<P: Serialize>(q: &Queue, p: &P)` must become `fn helper<P: Serialize>(q: &Queue<P>, p: &P)`, which changes the helper's call sites too, not just its body — with `&serde_json::to_value(p)?` as the escape hatch for callers who cannot change those. Cases 1–4 keep the `typed_queue` / `typed` migration.

## Finding 4 — the four constants now fail

Two layers, because one of the four is not reachable end to end.

**Integration.** `claimed_job_carries_every_core_field` now reads *both* rows back through `get_job` and asserts their two different priorities, run_ats and expires_ats. `typed_queue_round_trips_the_payload_type` now asserts its snapshot's `queue` (`"typed"`), `max_attempts` (`3`) and `priority` (`0`) — every one different from the `parity_queue` fixture's constants.

**Unit.** A claimed job's `state` is *always* `"processing"` end to end, so no integration test can distinguish a real read from a hardcoded `"processing"` — your table's first row has no integration fix. New `raw_job_mapping` tests in `src/lib.rs` drive `RawJob::into_job` / `into_row` directly with a synthetic wire row whose twelve values are all distinct (`id 11`, `queue "mapping"`, `priority 13`, `run_at 17`, `worker_id "w-19"`, `claim_expires_at 23`, `attempts 29`, `max_attempts 31`, `created_at 37`, `expires_at 41`, `state "sentinel-state"`), so neither a swap nor a constant survives.

Each mutation applied one at a time, `git diff --numstat` checked before and after every one (101/3 while applied, back to 100/2 after revert — your concurrent-stdout caution taken).

**`into_job` `state:` → `"processing".to_string()`** — was STILL PASSES:
```
thread 'raw_job_mapping::into_job_reads_every_field_from_the_row' panicked at src/lib.rs:2001:9:
assertion `left == right` failed: state, not a constant
  left: "processing"
 right: "sentinel-state"
```

**`into_row` `priority:` → `42`** — was STILL PASSES:
```
thread 'raw_job_mapping::into_row_reads_every_field_from_the_row' panicked at src/lib.rs:2023:9:
assertion `left == right` failed: priority, not a constant
  left: 42
 right: 13

thread 'claimed_job_carries_every_core_field' panicked at tests/job_details.rs:158:5:
assertion `left == right` failed: snapshot priority is per-row, not a constant
  left: 42
 right: 7

thread 'typed_queue_round_trips_the_payload_type' panicked at tests/job_details.rs:380:5:
assertion `left == right` failed: snapshot priority is the default 0, not the fixture's 42
  left: 42
 right: 0
```

**`into_row` `max_attempts:` → `9`** — was STILL PASSES:
```
thread 'raw_job_mapping::into_row_reads_every_field_from_the_row' panicked at src/lib.rs:2028:9:
assertion `left == right` failed: max_attempts, not a constant
  left: 9
 right: 31

thread 'typed_queue_round_trips_the_payload_type' panicked at tests/job_details.rs:376:5:
assertion `left == right` failed: snapshot max_attempts is QueueOpts::default's 3, not the fixture's 9
  left: 9
 right: 3
```

**`into_row` `queue:` → `"details".to_string()`** — was STILL PASSES:
```
thread 'raw_job_mapping::into_row_reads_every_field_from_the_row' panicked at src/lib.rs:2021:9:
assertion `left == right` failed: queue, not a constant
  left: "details"
 right: "mapping"

thread 'typed_queue_round_trips_the_payload_type' panicked at tests/job_details.rs:375:5:
assertion `left == right` failed: snapshot queue is read per-row
  left: "details"
 right: "typed"
```

All four reverted; final tree is byte-identical to the tested state and to what was pushed.

`run_at` / `created_at` coverage left exactly as it was, as you said.

## Not addressed here

- **Finding 3** (typed `Queue<T>` + global `get_job`) — belongs with #134, as you said. Not touched.
- **Finding 5** (the global-lookup rustdoc) — left in place; #134 deletes it.
- The pre-existing broken README "Quick start" (`Database::open` with two args, `db.queue` without `QueueOpts`, `q.enqueue` with one arg) is real and untouched by this PR. Worth its own issue — it is the first code a new user copies — but fixing it here would be unrelated scope.

russellromney:
## Second-round adversarial review

Three commits pushed: `456e74b`, `455444c`, `3b8483e`. Everything below is marked **[compiled]** (a compiler or a test run settled it) or **[read]** (source inspection only).

**Verdict: merge, after #124.** The design is right, the field wiring is right, the disclosed breaks are now correctly characterized, and the new surface is covered. The remaining items are deferrals to existing issues, not blockers.

---

### 1. Does it make sense to a Rust user?

**The `Queue<T = serde_json::Value>` + separate `typed_queue` split is the right call.** [read] The default type parameter on the *type* is exactly `HashMap<K, V, S = RandomState>`; it is idiomatic, and it keeps one `Queue` rather than the `TypedQueue<T>` newtype #136 suggests for Go. A newtype would have avoided both source breaks but doubled the surface (every one of `claim_batch` / `get_job` / `claim_waker` / `depth` / `sweep` re-exported or reached through `.raw()`), and it would read as a port of the Go design rather than as Rust. The `Database::queue` / `Database::typed_queue` split is forced — Rust has no default type parameter on a function, and a generic `queue()` would break every bare `let q = db.queue(..)` with E0282. The PR's reasoning holds and I could not find a better workaround.

**But the trap is real and worse than "the user silently gets `Value`."** [compiled] I probed the two things a user actually types:

```
error[E0107]: method takes 0 generic arguments but 1 generic argument was supplied
   --> tests/zz_error_probe.rs:16:16
    |
 16 |     let a = db.queue::<Email>("emails", QueueOpts::default());
    |                ^^^^^--------- help: remove the unnecessary generics
```

```
error[E0308]: mismatched types
19 |     let b: Queue<Email> = db.queue("emails", QueueOpts::default());
   |            ------------   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Queue<Email>`, found `Queue`
   = note: expected struct `Queue<Email>`
              found struct `Queue<serde_json::value::Value>`
```

Neither message mentions `typed_queue`, and the first actively steers the user *away* — "remove the unnecessary generics" is the wrong advice for someone who wants a typed queue. The compiler can't be steered here (no `#[diagnostic::on_unimplemented]` lever for E0107), so the docs have to. **Fixed in `456e74b`:** `Database::queue`'s rustdoc now opens with "for your own payload type, use `typed_queue` — not this method", quotes both errors, and links `Queue::typed`; the `Queue` struct doc explains why the split exists; the README and CHANGELOG say the same.

Naming: `typed_queue` / `typed_outbox` / `Queue::typed` reads fine. Not a TypeScript port — the TS design is `Queue<T>` with inference, which Rust genuinely cannot have here.

### 2. The disclosed breaks — migrations compiled

I wrote a throwaway consumer (`tests/zz_migration_probe.rs`, deleted before commit) that **compiles and runs** every migration the CHANGELOG documents. All six of its tests pass. [compiled]

**Finding — MEDIUM: the CHANGELOG's generic-helper claim was wrong.** It said, in bold, "**A generic helper has no drop-in migration**... the signature must become `fn helper<P: Serialize>(q: &Queue<P>, p: &P)`, which changes every call site of the helper too." That is false. This compiles and runs:

```rust
fn helper<P: Serialize>(q: &Queue, p: &P) -> honker::Result<i64> {
    q.typed::<P>().enqueue(p, EnqueueOpts::default())
}
```

Body-only. The signature is unchanged, so every call site of the helper is unchanged. It is also cheaper than the documented `&serde_json::to_value(p)?` escape hatch — `typed()` clones an `Arc` and a `String` and never touches the database, where `to_value` builds a whole intermediate `Value` tree. This was the single worst-characterized part of the break surface: a migration described as impossible is a migration a user will not look for. **Fixed in `456e74b`**: the CHANGELOG entry now leads with `typed::<P>()`, keeps the other two as alternatives, and `Queue::typed`'s rustdoc carries the same snippet as a doctest so it cannot silently stop compiling. `455444c` adds `a_generic_helper_migrates_without_touching_its_call_sites`, which calls the helper through a plain `db.queue(..)` handle and reads the row back.

Everything else the CHANGELOG documents checks out [compiled]: the `typed_queue` / `typed` migration for `&MyStruct`, `&"hello"`, `&vec![1u8,2,3]`; the `JobRow` `serde_json::from_value(json!({twelve keys}))` fake in **both** its untyped and typed (`JobRow<Email>`) form; a user type deriving `Deserialize` around a `JobRow` field (the restored derive); and a bare `let ob = Outbox::new(..)` inferring. `JobRow` keeping struct literals is genuinely impossible under any form of the parameter — public `_payload` trades E0451 for E0063 and buys a permanent wart — so keeping it with a working `Deserialize` migration is right.

### 3. Completeness

**All twelve fields, correct.** [read] `RawJob` matches core's `json!` keys in `honker_ops.rs` by name, no renames, no positional reads; `into_job` / `into_row` are the single mapping and all three claim paths (`claim_batch`, `ClaimWaker::next`, `try_next`) route through it. `worker_id`, `claim_expires_at`, `expires_at` are `Option` on `JobRow` and genuinely arrive as `None` — now asserted at the unit level too, not just end-to-end (`into_row_keeps_the_same_nulls_as_none`, `455444c`), so "never `Some("")`, never `Some(0)`" is pinned. [compiled]

**The `Job` / `JobRow` payload split (`Vec<u8>` vs `String`) is a genuine intra-binding inconsistency, not just the cross-binding one #146 tracks.** [read] Both hold the same raw JSON, but a user reads `row.payload` as a `&str` and has to `String::from_utf8(job.payload.clone())` for the identical bytes — the PR's own test does exactly that. Both types predate this PR, and #146 is the right place to settle it in one pass with the other bindings. Left alone deliberately; not touched here.

**Release-coordination risk, flagging only.** [read] `packages/honker-rs/Cargo.toml` requires `honker-core = "0.5.0"`, and core is still `0.5.0` on this branch even though #124 widened `honker_claim_batch`'s RETURNING from six columns to twelve. `RawJob` now requires all twelve keys — `state`, `priority`, `run_at`, `max_attempts`, `created_at` are non-`Option` — so a resolution to a *narrow* honker-core 0.5.0 fails to deserialize on every claim. Failing loudly is correct; the version requirement is the only thing keeping the pairing honest. Bumping it belongs to #124's release (and core is off-limits here), so this is a release checklist item, not a change: **honker-rs must not publish until honker-core publishes the twelve-key version, and the requirement must move to it.**

### 4. Test comprehensiveness — what was still uncovered

Round one closed the "hardcoded constant" hole well; I could not break the twelve-field wiring. Three gaps remained, all closed in `455444c`:

- **The typed `Outbox` had zero tests.** `db.typed_outbox`, `Outbox::new_typed`, `Outbox<T>::enqueue` / `enqueue_tx`, and `run_once`'s `FnMut(T)` closure — and `run_once`'s closure type changing from `serde_json::Value` to `T` is *one of the two disclosed source breaks*. `typed_outbox_delivers_the_payload_type` now runs a typed delivery including the rollback case.
- **`ClaimWaker<T>` had no typed test.** `typed_claim_waker_hands_back_a_typed_job` claims through `try_next` on a non-default `QueueOpts` and asserts `priority` and `max_attempts` from it, so a constant cannot pass.
- **The new "fail loudly" error path had no test.** "A claimed row without a `worker_id` or `claim_expires_at` is now an error instead of a silent default" was disclosed with nothing holding it. Two `raw_job_mapping` tests now drive both null cases and assert the error names the field and the job id.

**Mutations, each applied alone with `git diff --numstat` checked before and after** [compiled]:

| Mutation | Result |
|---|---|
| `into_job`'s `worker_id` `ok_or_else(..)?` → `unwrap_or_default()` | **FAILS** — `Job { .. worker_id: "", .. }`, exactly the silent default |
| `Database::typed_outbox` ignoring `name` | **FAILS** — `left: "_outbox:webhook"`, `right: "_outbox:mail"` |
| `Queue::typed`'s `name: self.name.clone()` → `"details"` | **FAILS** on the queue name |

That third one caught a weakness in *my own* new test and is worth recording, because it is the same failure mode round one found. `a_generic_helper_migrates_without_touching_its_call_sites` originally asserted only that the helper's row decoded back to `Email`. It **passed** under the mutation — because `get_job` is not queue-scoped (#134), it happily found the row after `typed()` had misrouted it to another queue. The payload assertion alone was not enough. Added `assert_eq!(helper_row.queue, "helper")` and it fails correctly. **Note for #134: the un-scoped `get_job` is quietly weakening tests in this file, not just the API.**

One mutation I could not turn into a clean failure: making `Outbox<T>::enqueue_tx` ignore its `tx` and write on a fresh connection **deadlocks** on the SQLite write lock rather than asserting false. The transactional path is exercised, but by hang, not by assertion — a property of the fixture shape (`surface.rs`'s untyped outbox test behaves the same way), not of the new test. Recorded rather than papered over.

### 5. The ~2000-line `lib.rs`

**Its own issue, not this PR.** [read] The file is 2048 lines against the repo's ~1000-line guideline, and yes, `RawJob` being private forced the new mapping tests to live inside it. But splitting here would bury a 490-line API change under a 2000-line file move and make exactly the diff this PR most needs reviewed unreviewable. The right split is roughly `queue.rs` / `job.rs` / `stream.rs` / `scheduler.rs` with a `pub(crate)` `RawJob`, which also un-forces the in-file unit tests. Worth filing.

### Notes, not defects

- **`Outbox` has no `typed()`** to match `Queue::typed`. Reinterpreting an outbox means going back through `Database::typed_outbox`. Asymmetric, but `typed_outbox` exists, so there is a path. Adding a method is new surface; flagging rather than expanding scope.
- **A poison payload in a typed outbox is slower to dead-letter than it looks.** `run_once` does `job.payload_typed()?` before the delivery closure, so an undecodable payload returns `Err` with the job still claimed — it re-claims after the visibility timeout, and only reaches `dead_letter_exhausted_claimable` after `max_attempts` full timeout cycles, rather than failing immediately. The structure predates this PR (it was `from_slice::<Value>` before, which essentially never failed); typed payloads make the failure realistic. Changing it to `job.fail()` on a decode error is new semantics — flagging, not fixing.
- `typed_queue_reads_a_payload_another_writer_produced` is still two Rust types in one process, not the cross-binding read #136's test list asks for. Fine as a unit test; does not discharge that checklist item.
- Round one's `Queue::get_job` rustdoc note about global lookup is left in place for #134 to delete, as agreed.
- `claim_waker_wakes_on_run_at_deadline` failed once under parallel load and passed alone and on a clean re-run of the full CI command. Timing flake on a loaded box, unrelated to this change.

### The README (#157)

**Fixed here** (`3b8483e`), not deferred — it is the first code a new user copies and it sat directly above the correct sections this PR adds. `Database::open` took two arguments where it takes one, `db.queue` omitted `QueueOpts`, `q.enqueue` passed an owned value with one argument, and the Install line still told the reader to make a `.dylib` available at runtime. The scheduler snippet a few paragraphs down was wrong the same way (`sched.add("fast", "emails", "@every 1s", json!(..))` against `add(&self, task: ScheduledTask)`) and is fixed with it — leaving one broken block two paragraphs from a corrected one is worse than fixing both. Signatures checked against `src/lib.rs` and the compiled `examples/basic.rs`, which the quick start now mirrors and links. **[read], not [compiled]** — README blocks are not doctested (no `include_str!`). Making them doctests is a real improvement and a bigger change than this PR should carry.

### Verification

`cargo test --locked --manifest-path packages/honker-rs/Cargo.toml --features kernel-watcher,shm-fast-path` — 89 tests pass, 0 failed, including 3 doctests. `cargo fmt --check` clean. `cargo clippy --lib --test job_details -- -D warnings` clean (`--all-targets` has two pre-existing `manual_find` failures in `python_interop` and `phase_mantle`, untouched by this PR and not part of the CI gate).

PR #151, opened 2026-09-01
feat(cpp): full job details and a JobSnapshot type (#136)

russellromney:
Part of #136 (C++ binding checklist item).

## Stacked on #124

**This PR targets `feat/node-typed-jobs` (#124), not `main`, and must merge after #124.** It depends on the widened `honker_claim_batch` RETURNING clause that exists only on that branch. `honker-core` is unchanged here.

## What changed

`packages/honker-cpp/include/honker.hpp` only — no Zig, no C ABI change.

- **All twelve fields on `honker::Job`.** Added `queue()`, `state()`, `priority()`, `run_at()`, `claim_expires_at()`, `max_attempts()`, `created_at()`, `expires_at()` alongside the existing `id()`, `payload()`, `worker_id()`, `attempts()`. Keeps `ack()` / `retry()` / `fail()` / `heartbeat()`.
- **New `honker::JobSnapshot`.** Same twelve fields, data only, no claim methods. Returned by the new `Queue::get_job(id) -> std::optional<JobSnapshot>`.
- On `Job`, `worker_id()` and `claim_expires_at()` are plain values because holding a claim is what makes it a `Job`. On `JobSnapshot` they are `std::optional` because a pending row has neither. `expires_at()` is `std::optional<int64_t>` on both.
- `Queue::get_job_json(id)` is unchanged and still returns the undecoded blob.
- **Style.** Followed what is already in this header: snake_case accessor methods, trailing-underscore members, `std::optional` for nullable fields, `nlohmann::json` for parsing (already a header dependency, used by `Outbox`). No templates and no codec — #136 explicitly says C++ keeps dynamic values / raw JSON unless a small codec helper is useful, and nothing here needed one.

## Payload encoding (please read: cross-binding divergence)

Per the coordination note, payload encoding is **unchanged**: `payload()` is the raw JSON text on both `Job` and `JobSnapshot`, exactly as before.

| binding | claimed job payload | snapshot (`getJob`/`get_job`) payload |
| --- | --- | --- |
| C++ (this PR) | raw JSON text | raw JSON text |
| Node (#124) | decoded value | **decoded value** |
| Rust | raw JSON bytes | raw JSON text |
| Go | raw JSON | raw JSON text |
| Python | raw JSON | raw JSON text |

Node decodes; the rest hand back raw. Worth one deliberate decision across all bindings later. Flagging, not resolving.

## Fail loudly (behavior change)

Claim parsing used to be wrapped in `catch (...) {}` and pulled fields with `j.value("id", 0)`, `j.value("attempts", 1)`, and friends. A malformed blob silently became an empty result, and a missing field silently became a plausible-looking default — exactly the failure mode that would make a field-parity test pass for the wrong reason.

Decoding now throws `honker::Error` on unparseable JSON or on a missing/null required field. Nullable fields (`worker_id` and `claim_expires_at` on a snapshot, `expires_at` on both) still decode to `std::nullopt`. An empty claim result is still an empty result — that is a real answer, not a failure.

## ABI

No exported C symbol added, removed, or renamed. `honker_cpp_cancel`, `honker_cpp_get_job`, `honker_cpp_claim_batch` and the rest are untouched, and `src/honker.zig` has no diff. `Job`'s public constructor is replaced by a `Job::from_json(sqlite3*, const nlohmann::json&)` factory; nothing in the repo constructed a `Job` directly.

## Tests

New `packages/honker-cpp/test/test_job_details.cpp`, registered in `build.zig`. 5 tests, all passing:

- `claimed_job_carries_every_core_field`
- `claimed_job_attempts_count_up_across_retries`
- `snapshot_pending_then_processing_then_miss`
- `delayed_job_snapshot_reports_future_run_at`
- `get_job_json_still_returns_the_raw_blob`

Every field is asserted against a value that could not be produced by another field. The fixture deliberately avoids defaults and same-second coincidences: visibility timeout 137s (not 300), `max_attempts` 9 (not 3), priorities 42 and 7 on two rows so `priority` cannot be a constant, `run_at` back-dated 300s so it is never equal to `created_at`, `expires_at` set on one row and null on the other, and `claim_expires_at` bracketed against the claim instant plus 137s.

The C ABI's `honker_cpp_enqueue` takes only delay/priority/max_attempts, so the fixture rows that need an absolute `run_at` or an expiry are inserted through the `honker_enqueue` SQL function on `db.raw()`. That is only how the row gets on disk; the read path under test is the ordinary binding one. The binding's own `enqueue()` is exercised by the delay and raw-blob tests.

Assertions use small `check_i64` / `check_str` / `check_range` helpers so a failure prints got-vs-want rather than just an expression.

The existing suites (`test_basic`, `test_parity` 28 tests, `test_phase_mantle` 8 tests) pass unmodified against the new header.

### Proof the tests are load-bearing

Two mutations, each reverted after:

**1. `run_at` sourced from `created_at`** in `Job::from_json`. `claimed_job_carries_every_core_field` fails:

```
FAIL: run_at: got 1788257032, want 1788256732
```

The 300-second gap is the deliberate `run_at` back-date.

**2. `created_at` sourced from `run_at`** in `JobSnapshot::from_json`. `delayed_job_snapshot_reports_future_run_at` fails:

```
FAIL: run_at must sit exactly 600s after created_at: got 0, want 600
```

This one is the interesting case: `snapshot_pending_then_processing_then_miss` **passed** under that mutation, because an undelayed job legitimately has `run_at == created_at`. That is exactly the coincidence-equality false pass to watch for, and the delayed-job test is what discriminates it.

Run locally with:

```
zig build test -Dsqlite-prefix="$(brew --prefix sqlite)" \
  -Djson-prefix="$(brew --prefix nlohmann-json)" \
  -Dhonker-ext=/path/to/libhonker_ext.dylib
```

## Not in scope

- `claimed_at`: that is #135 / PR #140 on another branch, deliberately not added here.
- `honker-core`, the SQL ABI, and `src/honker.zig` are untouched.
- Queue-scoped `get_job`: #134.
- No other binding touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,15 @@
   that saved that offset replay the whole topic; a malformed
   `scheduler_tick` row used to become a fire with an empty name and
   `job_id = 0`. Both now throw `honker::Error`.
+- `StreamSubscription`'s checkpoint write now fails loudly too. It discarded
+  the return of `honker_cpp_stream_save_offset` and cleared its pending
+  counter regardless, so a checkpoint that failed — read-only volume, full
+  disk, `SQLITE_BUSY` — looked identical to one that succeeded, and the
+  consumer replayed the topic with nothing reporting why. `save_offset()`
+  and `next()` now throw `honker::Error`, and a failed write leaves the
+  position pending so the destructor and a manual retry try again. `next()`
+  rolls its in-memory position back on that throw, so the event is
+  redelivered rather than skipped.
 
 ## Unreleased — Node checkpoint interoperability (Phase Robinson)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,10 +52,17 @@
 - Payload encoding is unchanged: `payload()` is the raw JSON text on both
   types. The binding stays dynamic — no templates, no codec — matching the
   issue's guidance for C++.
-- Claim and snapshot decoding now fails loudly. A malformed JSON blob or a
-  missing required field throws `honker::Error` instead of the previous
-  `catch (...) {}` that returned an empty result, and required fields no
-  longer fall back to `0` / `""` / `1` defaults.
+- Claim and snapshot decoding now fails loudly. A malformed JSON blob, a
+  missing required field, or a field of the wrong type throws
+  `honker::Error` instead of the previous `catch (...) {}` that returned an
+  empty result, and required fields no longer fall back to `0` / `""` / `1`
+  defaults. `nlohmann::json::type_error` is wrapped, so `honker::Error` is
+  the only exception type the decoders let out.
+- Stream and scheduler decoding fails loudly the same way. A malformed
+  `stream_read_since` row used to become `offset = 0`, which made a consumer
+  that saved that offset replay the whole topic; a malformed
+  `scheduler_tick` row used to become a fire with an empty name and
+  `job_id = 0`. Both now throw `honker::Error`.
 
 ## Unreleased — Node checkpoint interoperability (Phase Robinson)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,10 @@
 - `Queue::get_job_json()` and `Scheduler::list_json()` free the shim's
   string with `honker_cpp_free()` instead of `std::free()`, matching every
   other call that takes ownership of a `char*` from the shim.
+- `Job` and `JobSnapshot` string accessors return `const std::string&`
+  instead of a copy. They were marked `noexcept` while returning by value,
+  which allocates — a `std::bad_alloc` there would have called
+  `std::terminate` rather than propagating.
 
 ## Unreleased — Node checkpoint interoperability (Phase Robinson)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,26 @@
   unscoped lookup could hand back a payload that did not match the queue's
   declared `TPayload`. `cancel()` is unchanged and remains not queue-scoped.
 
+## Unreleased — C++ job details
+
+- C++ `honker::Job` now carries all twelve fields core returns. It gained
+  `queue()`, `state()`, `priority()`, `run_at()`, `claim_expires_at()`,
+  `max_attempts()`, `created_at()`, and `expires_at()` alongside the
+  existing `id()`, `payload()`, `worker_id()`, and `attempts()`, and keeps
+  `ack()` / `retry()` / `fail()` / `heartbeat()`.
+- New `honker::JobSnapshot`: the same twelve fields, data only, returned by
+  `Queue::get_job(id)` as `std::optional<JobSnapshot>`. `worker_id()` and
+  `claim_expires_at()` are `std::optional` there because a pending row has
+  neither. `Queue::get_job_json(id)` is unchanged and still returns the
+  undecoded blob.
+- Payload encoding is unchanged: `payload()` is the raw JSON text on both
+  types. The binding stays dynamic — no templates, no codec — matching the
+  issue's guidance for C++.
+- Claim and snapshot decoding now fails loudly. A malformed JSON blob or a
+  missing required field throws `honker::Error` instead of the previous
+  `catch (...) {}` that returned an empty result, and required fields no
+  longer fall back to `0` / `""` / `1` defaults.
+
 ## Unreleased — Node checkpoint interoperability (Phase Robinson)
 
 - Node stream checkpoint calls now use the shared SQL ABI's canonical
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,9 @@
   position pending so the destructor and a manual retry try again. `next()`
   rolls its in-memory position back on that throw, so the event is
   redelivered rather than skipped.
+- `Queue::get_job_json()` and `Scheduler::list_json()` free the shim's
+  string with `honker_cpp_free()` instead of `std::free()`, matching every
+  other call that takes ownership of a `char*` from the shim.
 
 ## Unreleased — Node checkpoint interoperability (Phase Robinson)
 
--- a/packages/honker-cpp/README.md
+++ b/packages/honker-cpp/README.md
@@ -106,9 +106,11 @@ if (auto row = q.get_job(id)) {
 `Queue::get_job_json(id)` still returns the undecoded JSON blob for
 callers that want the bytes.
 
-A claim or lookup whose JSON is malformed, or that is missing a
-required field, throws `honker::Error` rather than returning a
-default-filled job.
+A claim or lookup whose JSON is malformed, that is missing a required
+field, or that carries a field of the wrong type throws `honker::Error`
+rather than returning a default-filled job. That is the only exception
+type honker's decoders let out — nlohmann's `type_error` is wrapped, so
+`catch (const honker::Error&)` is enough.
 
 Delayed jobs use `run_at` options on enqueue. Recurring schedules use schedule expressions:
 
--- a/packages/honker-cpp/README.md
+++ b/packages/honker-cpp/README.md
@@ -96,6 +96,10 @@ is `std::optional<int64_t>` on both.
 preferred JSON library. honker never checks the payload's shape; every
 process writing a queue must agree on it.
 
+The string accessors (`queue()`, `payload()`, `state()`, `worker_id()`)
+return a `const&` into the `Job` or `JobSnapshot`, so reading a payload
+costs nothing. Copy it if you need it to outlive the job.
+
 ```cpp
 if (auto row = q.get_job(id)) {
     row->state();                       // "pending" or "processing"
--- a/packages/honker-cpp/README.md
+++ b/packages/honker-cpp/README.md
@@ -71,6 +71,45 @@ int main() {
 }
 ```
 
+## Job details
+
+`honker::Job` (a claimed unit of work) and `honker::JobSnapshot` (the
+read-only row from `Queue::get_job`) both carry every field the core
+claim/lookup returns:
+
+```cpp
+auto job = q.claim_one("worker-1");
+job->id(); job->queue(); job->payload(); job->state();
+job->priority(); job->run_at(); job->worker_id(); job->claim_expires_at();
+job->attempts(); job->max_attempts(); job->created_at(); job->expires_at();
+```
+
+`Job` also holds a claim, so it has `ack()`, `retry()`, `fail()`, and
+`heartbeat()`. `JobSnapshot` is data only.
+
+On `Job`, `worker_id()` and `claim_expires_at()` are plain values
+because holding a claim is what makes it a `Job`. On `JobSnapshot` they
+are `std::optional` because a pending row has neither. `expires_at()`
+is `std::optional<int64_t>` on both.
+
+`payload()` is the raw JSON text exactly as stored — parse it with your
+preferred JSON library. honker never checks the payload's shape; every
+process writing a queue must agree on it.
+
+```cpp
+if (auto row = q.get_job(id)) {
+    row->state();                       // "pending" or "processing"
+    nlohmann::json::parse(row->payload());
+}
+```
+
+`Queue::get_job_json(id)` still returns the undecoded JSON blob for
+callers that want the bytes.
+
+A claim or lookup whose JSON is malformed, or that is missing a
+required field, throws `honker::Error` rather than returning a
+default-filled job.
+
 Delayed jobs use `run_at` options on enqueue. Recurring schedules use schedule expressions:
 
 ```cpp
--- a/packages/honker-cpp/build.zig
+++ b/packages/honker-cpp/build.zig
@@ -34,6 +34,7 @@ pub fn build(b: *std.Build) void {
         "test/test_basic.cpp",
         "test/test_parity.cpp",
         "test/test_phase_mantle.cpp",
+        "test/test_job_details.cpp",
     };
 
     inline for (test_files) |tf| {
--- a/packages/honker-cpp/include/honker.hpp
+++ b/packages/honker-cpp/include/honker.hpp
@@ -1028,6 +1028,27 @@ private:
     int64_t     job_id_;
 };
 
+namespace detail {
+
+/// Decode one honker_scheduler_tick() blob. Every field is required: a
+/// fire with name "" or job_id 0 is not a plausible default, it is a
+/// row we failed to read.
+inline std::vector<ScheduledFire> parse_scheduler_fires(const std::string& json) {
+    auto arr = parse_row_array(json, "scheduler tick");
+    std::vector<ScheduledFire> out;
+    out.reserve(arr.size());
+    for (const auto& j : arr) {
+        out.emplace_back(
+            row_str(j, "name", "scheduler fire"),
+            row_str(j, "queue", "scheduler fire"),
+            row_i64(j, "fire_at", "scheduler fire"),
+            row_i64(j, "job_id", "scheduler fire"));
+    }
+    return out;
+}
+
+}  // namespace detail
+
 // =====================================================================
 // Scheduler
 // =====================================================================
@@ -1193,23 +1214,10 @@ public:
     Scheduler(Database* db) : db_(db) {}
 
 private:
-    /// Decode one honker_scheduler_tick() blob. Every field is required:
-    /// a fire with name "" or job_id 0 is not a plausible default, it is
-    /// a row we failed to read.
     std::vector<ScheduledFire> parse_fires(char* rows) {
         std::string json{rows};
         honker_cpp_free(rows);
-        auto arr = detail::parse_row_array(json, "scheduler tick");
-        std::vector<ScheduledFire> out;
-        out.reserve(arr.size());
-        for (const auto& j : arr) {
-            out.emplace_back(
-                detail::row_str(j, "name", "scheduler fire"),
-                detail::row_str(j, "queue", "scheduler fire"),
-                detail::row_i64(j, "fire_at", "scheduler fire"),
-                detail::row_i64(j, "job_id", "scheduler fire"));
-        }
-        return out;
+        return detail::parse_scheduler_fires(json);
     }
 
     Database* db_;
--- a/packages/honker-cpp/include/honker.hpp
+++ b/packages/honker-cpp/include/honker.hpp
@@ -410,8 +410,15 @@ inline std::string row_str(const nlohmann::json& j, const char* key, const char*
     return row_get<std::string>(row_field(j, key, what), key, what, "a string");
 }
 
+/// A genuinely nullable key: absent or null means std::nullopt, never
+/// 0 or "". The row still has to be an object — otherwise every
+/// nullable field on a garbage row would silently read as null, which
+/// is the tolerant lookup this file exists to avoid.
 inline std::optional<int64_t> row_opt_i64(
     const nlohmann::json& j, const char* key, const char* what) {
+    if (!j.is_object()) {
+        throw Error{std::string{what} + " is not a JSON object"};
+    }
     const auto it = j.find(key);
     if (it == j.end() || it->is_null()) return std::nullopt;
     return row_get<int64_t>(*it, key, what, "an integer");
@@ -419,6 +426,9 @@ inline std::optional<int64_t> row_opt_i64(
 
 inline std::optional<std::string> row_opt_str(
     const nlohmann::json& j, const char* key, const char* what) {
+    if (!j.is_object()) {
+        throw Error{std::string{what} + " is not a JSON object"};
+    }
     const auto it = j.find(key);
     if (it == j.end() || it->is_null()) return std::nullopt;
     return row_get<std::string>(*it, key, what, "a string");
--- a/packages/honker-cpp/include/honker.hpp
+++ b/packages/honker-cpp/include/honker.hpp
@@ -674,6 +674,12 @@ public:
         return std::move(jobs.front());
     }
 
+    /// Claim up to n jobs. Decoding is all-or-nothing: one unreadable
+    /// row throws and no Job comes back, even though the claim UPDATE
+    /// already ran on every row it matched. Those rows stay claimed and
+    /// carry a burnt attempt until their visibility timeout expires,
+    /// and the exception carries no ids, so treat it as a core bug and
+    /// not as a retryable condition.
     std::vector<Job> claim_batch(std::string_view worker_id, int64_t n) {
         const std::string w{worker_id};
         char* rows = honker_cpp_claim_batch(
@@ -733,8 +739,9 @@ public:
     /// Read a single job row by id as a decoded JobSnapshot. Returns
     /// nullopt on miss (ack'd, dead'd, or never existed).
     ///
-    /// NOT queue-scoped: job ids are globally unique and this lookup
-    /// hits any queue's row. Scoping is tracked in #134.
+    /// Today this lookup is not queue-scoped — job ids are globally
+    /// unique and it hits any queue's row. Whether it stays that way
+    /// is #134; do not depend on either answer.
     std::optional<JobSnapshot> get_job(int64_t job_id) {
         const std::string json = get_job_json(job_id);
         if (json.empty()) return std::nullopt;
--- a/packages/honker-cpp/include/honker.hpp
+++ b/packages/honker-cpp/include/honker.hpp
@@ -483,16 +483,20 @@ inline nlohmann::json parse_row_array(const std::string& json, const char* what)
 /// worker_id() and claim_expires_at() are optional because a pending
 /// row has neither. payload() is the raw JSON text exactly as stored;
 /// parse it with your preferred JSON library.
+///
+/// The string accessors hand back a reference into the snapshot, so
+/// they neither copy nor allocate — keep the JobSnapshot alive for as
+/// long as you hold one.
 class JobSnapshot {
 public:
     int64_t     id()         const noexcept { return id_; }
-    std::string queue()      const noexcept { return queue_; }
-    std::string payload()    const noexcept { return payload_; }
+    const std::string& queue()   const noexcept { return queue_; }
+    const std::string& payload() const noexcept { return payload_; }
     /// "pending" or "processing".
-    std::string state()      const noexcept { return state_; }
+    const std::string& state()   const noexcept { return state_; }
     int64_t     priority()   const noexcept { return priority_; }
     int64_t     run_at()     const noexcept { return run_at_; }
-    std::optional<std::string> worker_id() const noexcept { return worker_id_; }
+    const std::optional<std::string>& worker_id() const noexcept { return worker_id_; }
     std::optional<int64_t> claim_expires_at() const noexcept { return claim_expires_at_; }
     int64_t     attempts()     const noexcept { return attempts_; }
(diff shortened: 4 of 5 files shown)


russellromney:
Adversarial review (I did not write this code). Everything below was RUN unless marked INFERRED, in a detached worktree at `5911d83` against a prebuilt extension, `zig build test -Dhonker-ext=... -Dsqlite-prefix=... -Djson-prefix=...`.

Baseline: all five new tests execute and pass, plus the existing 28 parity + phase-mantle tests.

### 1. HIGH — the headline "fails loudly" claim has ZERO test coverage. RAN.

The CHANGELOG and README both promise that malformed JSON or a missing required field now throws instead of returning a default-filled job. Nothing tests it. I reverted the behavior wholesale in `include/honker.hpp` — back to `j.value(key, 0)` / `j.value(key, "")` defaults, `catch (...) { return json::object(); }` in `parse_job_json`, and `return {}` instead of throwing on a non-array claim result:

```cpp
inline int64_t job_i64(const nlohmann::json& j, const char* key) {
    return j.value(key, static_cast<int64_t>(0));
}
inline std::string job_str(const nlohmann::json& j, const char* key) {
    return j.value(key, std::string{});
}
```
(`job_field` deleted entirely — no throw path left.)

```
claimed_job_carries_every_core_field: ok
claimed_job_attempts_count_up_across_retries: ok
snapshot_pending_then_processing_then_miss: ok
delayed_job_snapshot_reports_future_run_at: ok
get_job_json_still_returns_the_raw_blob: ok
all job detail tests passed
```

Every new test still passes with the entire fix removed. The tests only ever see well-formed blobs, so they exercise the decode path, not the failure path.

Concrete failure scenario: someone refactors `job_field` back to a tolerant lookup (or adds a thirteenth field with a default), the guarantee silently regresses, and a malformed row again becomes `id=0, attempts=1`. `job->ack()` then targets id 0, returns false, and the real job silently expires and gets re-delivered.

Fix: three direct unit tests against the public `JobSnapshot::from_json` / `Job::from_json` — a blob missing `id`, a blob that is not parseable, a claim result that is not an array — asserting each throws.

### 2. MEDIUM — a wrong-*typed* field throws `nlohmann::json::type_error`, not `honker::Error`, contradicting the README. RAN.

`job_field` only guards absent/null. `.get<int64_t>()` / `.get<std::string>()` on a present-but-wrong-typed value throws out of nlohmann. Probe compiled against the PR's header:

```
missing id:                         honker::Error -> job row is missing required field 'id'
id is a string:                     nlohmann...type_error -> [json.exception.type_error.302] type must be number, but is string
payload is an object not a string:  nlohmann...type_error -> [json.exception.type_error.302] type must be string, but is object
expires_at is a string:             nlohmann...type_error -> [json.exception.type_error.302] type must be number, but is string
worker_id is a number:              nlohmann...type_error -> [json.exception.type_error.302] type must be string, but is number
```

README says: "A claim or lookup whose JSON is malformed, or that is missing a required field, throws `honker::Error`." A user who writes `catch (const honker::Error&)` around `claim_batch()` catches four of five malformed shapes and lets the fifth escape to `std::terminate` if it is the outermost frame. Either wrap the `.get<>()` calls in `job_field`'s error, or narrow the README to say honker::Error for missing fields and `std::exception` for the rest.

### 3. MEDIUM — `Queue::get_job`'s new doc comment is written to become wrong.

```cpp
/// NOT queue-scoped: job ids are globally unique and this lookup
/// hits any queue's row. Scoping is tracked in #134.
```

Leaving `get_job` global is expected and fine. Committing the *global contract* to a doc comment is not — #134 flips it, and this comment then documents behavior the code no longer has. #141 JVM is now the reference (`db.getJob(jobId).filter(job -> name.equals(job.queue()))` on `Queue`, global on `Database`). Also note C++ has **no** `Database::get_job` today, so when scoping lands there is no global escape hatch unless one is added at the same time.

Otherwise attack E is clean: no test asserts cross-queue lookup (I grepped all four test files — every `get_job`/`get_job_json` call uses an id from the same queue handle), and `get_job` is a single self-contained method body to filter.

### 4. LOW — `claim_batch` now throws away an entire successful claim on one bad element.

`parse_claim` loops `out.push_back(Job::from_json(...))`. Element 3 of 10 throwing discards all 10 `Job`s — but the core already ran the claim UPDATE: all 10 rows have `attempts` incremented and a `claim_expires_at` set. The caller has no ids, so those 10 jobs are unreachable until the visibility timeout expires, each having burned an attempt. The old `catch (...)` returned a partial list, which was lossy in a different way, so this is a tradeoff rather than a regression — but it deserves a line in the doc comment, since the exception is now the caller's only signal and it carries no ids.

### 5. LOW — new test file uses fixed `/tmp` filenames while every existing test uses unique ones.

`fs::temp_directory_path() / "honker-cpp-details-claim.db"` etc. `test_parity.cpp` uses `tmp_db()`. Two concurrent runs on one machine collide.

- **No exported C symbol changed.** `diff` of the entire `extern "C"` block between `origin/feat/node-typed-jobs` and the PR head: identical. `src/honker.zig` untouched. Confirms the ABI claim.
- **`catch (...)` count: 9 → 7.** Both removals are the job-decode ones, exactly as claimed. The seven survivors are: `Database` ctor (rethrows after `close()` — correct), `Lock` dtor and move-assign (cannot throw from a dtor — correct), `StreamSubscription` dtor `flush_offset` (same, though it silently loses the checkpoint), and three that *are* the same anti-pattern the PR fixed for jobs: `Stream::parse_events`, `StreamSubscription`'s read path, and `Scheduler::parse_fires`. Those still carry `j.value("offset", 0)`, `j.value("payload", "")`, `j.value("created_at", 0)`, `j.value("job_id", 0)` and swallow into a silently truncated vector. The `offset` one is the dangerous shape: a malformed row yields offset 0, and a consumer that saves that offset replays the whole stream. **Out of scope for this PR** — flagging so it does not get lost, since the CHANGELOG's "fails loudly" wording reads broader than what shipped.
- **Field wiring is genuinely well tested.** The fixture is careful in exactly the right way: `VISIBILITY_TIMEOUT_S=137`, `MAX_ATTEMPTS=9`, priorities 42/7, `EXPIRES_S=9000`, and — the part most bindings got wrong — `raw_enqueue` back-dates `run_at` 300 s away from `created_at` plus an explicit `check_ne_i64(high.run_at(), high.created_at())`. run_at↔created_at, attempts↔max_attempts (1 vs 9), claim_expires_at↔expires_at (claim+137 vs enqueue+9000), queue↔state, priority: every swap would fail. The second row with `expires = nullopt` proves `expires_at` is read per-row.
- **Nulls are right.** `worker_id`/`claim_expires_at` are `std::optional` on `JobSnapshot` and asserted `!has_value()` while pending; plain values on `Job`, which is correct because the claim UPDATE always writes both.
- **No memory regression.** `claim_one` now frees `rows` itself instead of delegating to the removed `parse_jobs`; both paths free exactly once, including on the throw path.
- The removed public `Job(sqlite3*, int64_t, std::string, std::string, int64_t)` ctor is a source break in principle; nothing in the repo constructs a `Job` directly, and users have no `sqlite3*` to pass. Not a practical concern.

Would not merge until finding 1 is addressed — the PR's central claim is the one thing the 327 new test lines do not check.

russellromney:
Findings 1, 2 and 3 are addressed in `1d06596` + `671ca7e`. CI is green
(`ci · all green`, run 33586093339).

The point you made is the one I had to answer: a passing run proves nothing
when the guard can be deleted and the suite stays green. So every new test
below is shown with the output it produces when the behavior it defends is
reverted. Reverts were applied to `include/honker.hpp` one at a time,
rebuilt, and run; the header was restored from a saved copy after each.

Test binary is `zig-out/bin/test_job_details`. The five new decode tests run
before the `HONKER_EXTENSION_PATH` check, so they need no extension and no
database — they call exactly the functions `Queue::get_job()` and
`Queue::claim_batch()` call.

Baseline, nothing reverted:

```
decode_rejects_unparseable_json: ok
decode_rejects_a_non_array_claim_result: ok
decode_rejects_a_missing_required_field: ok
decode_reports_a_wrong_typed_field_as_honker_error: ok
stream_and_scheduler_rows_decode_loudly: ok
claimed_job_carries_every_core_field: ok
claimed_job_attempts_count_up_across_retries: ok
snapshot_pending_then_processing_then_miss: ok
delayed_job_snapshot_reports_future_run_at: ok
get_job_json_still_returns_the_raw_blob: ok
all job detail tests passed
```

---

## Finding 1 — the fail-loudly claim now has coverage

### `decode_rejects_unparseable_json`

Revert: `detail::parse_row_blob` back to `catch (...) { return json::object(); }`.

```
FAIL: get_job blob with truncated JSON: returned normally — no error was raised
EXIT=134
```

### `decode_rejects_a_non_array_claim_result`

Revert: `detail::parse_row_array`'s `if (!arr.is_array())` back to
`return json::array()` instead of throwing.

```
decode_rejects_unparseable_json: ok
FAIL: claim result that is a single object: returned normally — no error was raised
EXIT=134
```

### `decode_rejects_a_missing_required_field`

Revert: `row_i64` / `row_str` back to `j.value(key, 0)` / `j.value(key, "")`,
the exact shape you restored.

```
decode_rejects_unparseable_json: ok
decode_rejects_a_non_array_claim_result: ok
FAIL: snapshot with no id: returned normally — no error was raised
EXIT=134
```

This one loops over every non-nullable snapshot field — `id`, `queue`,
`payload`, `state`, `priority`, `run_at`, `attempts`, `max_attempts`,
`created_at` — both absent and explicitly `null`, plus `worker_id` and
`claim_expires_at` through `Job::from_json`, plus a row that is not an
object at all. That is the loop that catches your "someone adds a
thirteenth field with a default" scenario.

### `decode_reports_a_wrong_typed_field_as_honker_error` (finding 2)

Revert: drop the try/catch in `detail::row_get`, so `.get<T>()` is called
raw, i.e. the state this PR was in when you probed it.

```
decode_rejects_unparseable_json: ok
decode_rejects_a_non_array_claim_result: ok
decode_rejects_a_missing_required_field: ok
FAIL: snapshot with a wrong-typed id: leaked nlohmann::json::exception "[json.exception.type_error.302] type must be number, but is string" instead of honker::Error
EXIT=134
```

### Wholesale revert — your experiment, repeated

All four job guards reverted at once, plus the stream/scheduler ones below:

```
FAIL: get_job blob with truncated JSON: returned normally — no error was raised
EXIT=134
```

The suite no longer survives it.

---

## Finding 2 — I wrapped, rather than narrowing the README

I took the wrap option. `detail::row_get` catches `nlohmann::json::exception`
and rethrows it as `honker::Error`:

```cpp
template <typename T>
T row_get(const nlohmann::json& v, const char* key, const char* what, const char* want) {
    try {
        return v.get<T>();
    } catch (const nlohmann::json::exception& e) {
        throw Error{std::string{what} + " field '" + key + "' is not " + want +
                    ": " + e.what()};
    }
}
```

Reason: a caller of a honker API should not have to know which JSON library
the header happens to use, and your `std::terminate` scenario is real — the
one shape that escapes is the one nobody catches. Narrowing the README would
have documented the hazard instead of removing it, and would have pinned
`nlohmann` into the public contract.

Your five probe shapes now read:

```
missing id:                         honker::Error -> job row is missing required field 'id'
id is a string:                     honker::Error -> job row field 'id' is not an integer: [json.exception.type_error.302] type must be number, but is string
payload is an object not a string:  honker::Error -> job row field 'payload' is not a string: ...
expires_at is a string:             honker::Error -> job row field 'expires_at' is not an integer: ...
worker_id is a number:              honker::Error -> job row field 'worker_id' is not a string: ...
```

The nlohmann message is kept in the text, so the diagnostic is not lost. The
README sentence is widened to say wrong-typed fields are covered too, and
that `honker::Error` is the only type the decoders let out.

---

## Finding 3 — `catch (...)` 7 → 4, and the four survivors say why

**Fixed (the three that were the same anti-pattern):**

- `Stream::parse_events` and `StreamSubscription::read_batch` were byte-identical
  bodies. Both now call one shared `detail::parse_stream_events`, which decodes
  through the strict helpers and throws. This is the `offset` case you flagged
  as the dangerous one — a swallowed row used to become `offset = 0`, and a
  consumer that saved that offset replayed the whole topic.
- `Scheduler::parse_fires` moved to `detail::parse_scheduler_fires` and throws
  the same way. A fire with `name = ""` and `job_id = 0` is not a plausible
  default; it is a row we failed to read.

`key` stays nullable on a stream event, because the core writes NULL for an
unkeyed publish. The test asserts a null key still decodes to `""`.

These are covered by `stream_and_scheduler_rows_decode_loudly`. Revert: both
decoders back to `catch (...) {}` with `j.value("offset", 0)` /
`j.value("name", "")` defaults.

```
decode_rejects_unparseable_json: ok
decode_rejects_a_non_array_claim_result: ok
decode_rejects_a_missing_required_field: ok
decode_reports_a_wrong_typed_field_as_honker_error: ok
FAIL: stream event with no offset: returned normally — no error was raised
EXIT=134
```

**Kept, each with a comment saying why** — all four are contexts that cannot
throw, none of them decode anything:

| site | why it stays |
| --- | --- |
| `Database` ctor | not a swallow: `close()` releases the handle, then the original failure is rethrown |
| `~StreamSubscription` `flush_offset` | a destructor may not throw; the lost checkpoint is survivable, the consumer replays from the last saved offset. Comment points at `save_offset()` for callers who want the failure reported |
| `~Lock` `release` | a destructor may not throw; an unreleased lock still expires on its TTL |
| `Lock::operator=(Lock&&)` | `noexcept` move-assign, same reasoning as the destructor |

No bare `catch (...)` is left without an explanation.

---

## Not touched, deliberately

- **ABI.** `diff` of the whole `extern "C"` block against `origin/feat/cpp-job-parity`
  before these commits: identical, 89 lines. `src/honker.zig` has no diff.
- **Finding 3 in your list (the `get_job` doc comment) and finding 4
  (`claim_batch` losing a whole batch on one bad element).** Both real, both
  left alone — out of the scope I was handed. Worth carrying into #134 and #146.
- **`get_job` queue scoping.** #134. Flagging what you found so it does not get
  lost: **C++ has no `Database::get_job` at all**, so when #134 scopes
  `Queue::get_job` there is no global escape hatch unless one is added in the
  same pass. The JVM shape (`Queue` filters, `Database` stays global) needs a
  new method here, not just a filter.
- **Payload encoding.** #146.
- **Finding 5 (fixed `/tmp` filenames).** The five new tests use no database at
  all, so they add no new collision surface. The existing five still use fixed
  names.

## Local verification

Extension built from this worktree, not reused from another checkout:

```
$ strings target/debug/libhonker_ext.dylib | grep -c "RETURNING id, queue, payload, state"
1
$ zig build test -Dhonker-ext=.../target/debug/libhonker_ext.dylib
ALL OK (28 tests)          # parity
all phase mantle tests passed
all job detail tests passed # 10 tests, 5 new
EXIT=0
```

russellromney:
Second-round adversarial review. I did not write this code. Everything below is
marked **RAN** or **READ**; RAN means executed in a worktree at `671ca7e` with
`zig build test -Dhonker-ext=<12-column libhonker_ext.dylib> -Dsqlite-prefix=…
-Djson-prefix=…`, against an extension whose `RETURNING id, queue, payload,
state` clause I checked with `strings` first.

Round 1 asked "is the headline claim tested". This round asked whether the PR is
complete, well designed, bug-free, comprehensively tested, and sensible to a C++
user. The interesting answer: **the decode half of "fails loudly" is now solid.
The write half was not, and the four surviving `catch (...)` comments asserted a
recovery path that did not exist.**

Fixed in `d89aef4` `4370b5a` `8921723` `287c05b` `0e09ffa`. Baseline before and
after: `ALL OK (28 tests)` parity, 8 phase-mantle, 11 job-detail (was 10).

---

## 1. HIGH — `StreamSubscription` threw away every failed checkpoint, and this PR documented that hole as closed. RAN.

```cpp
void flush_offset() {
    if (pending_ == 0) return;
    honker_cpp_stream_save_offset(db_, consumer_.c_str(), topic_.c_str(), last_offset_);
    pending_ = 0;                       // ← unconditional
}
```

The return code is discarded. `honker_cpp_stream_save_offset` returns
`HONKER_ERR_SQL` (-4) on any SQL failure — `step_scalar_int64` returns it
whenever `sqlite3_step` is not `SQLITE_ROW`, and `honker_stream_save_offset`
maps its `rusqlite::Error` through `to_sql_err`, so a read-only volume, a full
disk or `SQLITE_BUSY` all land there (READ, both sides). Every other call in the
header checks `if (n < 0) throw` — `Stream::save_offset` twenty lines away does.

Consequences, in order of how much they matter:

1. `pending_ = 0` runs anyway, so a failed checkpoint is indistinguishable from a
   successful one. The destructor will not retry it.
2. `next()` calls `flush_offset()` every `save_every_n_` events. A subscription
   can run for its whole life never persisting anything, silently.
3. **The `catch (...)` in `~StreamSubscription` is dead code for the real failure
   mode.** `flush_offset` could not throw on a SQL failure, so the swallow only
   ever caught nothing.
4. The comment this PR added on that `catch (...)` — *"Call save_offset() before
   the subscription goes out of scope if you need the failure reported"* — was
   false. `save_offset()` is `flush_offset()`. There was no way to learn the
   checkpoint failed, from anywhere.

This is the same hazard the PR names as its reason for making `parse_stream_events`
throw: *"a swallowed row used to become offset 0, and a consumer that saved that
offset replayed the whole topic."* A dropped checkpoint replays the topic exactly
the same way. The read half was closed; the write half was left open and then
documented as closed.

**Fixed** (`d89aef4`): `flush_offset` throws `honker::Error` on a negative return
and leaves `pending_` set, so the destructor and a manual retry both try again.
`next()` rolls its in-memory position back when the automatic checkpoint throws —
otherwise a caught-and-retried `next()` would skip the event permanently, which
would have been a new data-loss path introduced by my own fix.

New test `stream_subscription_reports_a_failed_checkpoint` forces the write to
fail with `PRAGMA query_only = 1` on the live connection, and asserts three
things: `save_offset()` throws, the saved offset is still 0, and a retry after
clearing the pragma commits it. Reverting `flush_offset` to the shipped version:

```
get_job_json_still_returns_the_raw_blob: ok
FAIL: save_offset swallowed a failed checkpoint write
EXIT=134
```

## 2. MEDIUM — the nullable helpers are tolerant, and only call order hides it. RAN.

`row_field` checks `is_object()`. `row_opt_i64` / `row_opt_str` do not, and
nlohmann's `find()` on a non-object returns `end()` rather than throwing. So every
nullable field on a garbage row read back as `std::nullopt` instead of an error.

Nothing hits it today only because each decoder happens to read a required field
first — `id` for jobs, `offset` for stream events, `name` for fires. That is
correctness by call order, in the one file whose entire purpose is removing
tolerant lookups. A reordered `from_json` reopens it.

**Fixed** (`4370b5a`), with two cases added to `decode_rejects_a_missing_required_field`.
Removing either guard:

```
FAIL: row_opt_i64 on a row that is not an object: returned normally — no error was raised
EXIT=134
```

## 3. MEDIUM — `std::string x() const noexcept { return x_; }` is a `std::terminate` path. READ, fixed.

Every string accessor on `Job` and `JobSnapshot` returned by value while declared
`noexcept`. Copying a `std::string` allocates; a `bad_alloc` inside a `noexcept`
function is `std::terminate`, not something `catch (const honker::Error&)` sees.
Same shape as the `type_error` leak round 1 found — an exit path the documented
catch cannot cover — and this one cannot even be caught with `catch (...)`.

It is also a heap allocation on every `job.payload()` in a claim loop, which for
a large payload is the hot path.

**Fixed** (`287c05b`): `const std::string&` / `const std::optional<std::string>&`.
The `noexcept` is now honest and the copy is gone. This matches `Database::path()`,
already `const&` in this header.

**The one behavior change worth your veto:** `const auto& p = q.get_job(id)->payload();`
used to lifetime-extend a temporary string and now dangles. `auto p = …` still
copies and is unaffected, as is every use in the four test files. I judged the
idiomatic accessor shape worth it and documented the lifetime in the class
comment and README; say the word and I will revert this one commit.

## 4. MEDIUM — `std::free` on shim-allocated memory. READ, fixed.

`Queue::get_job_json()` and `Scheduler::list_json()` used `std::free(raw)` on a
`char*` the Zig shim malloc'd; the other seven owning sites use `honker_cpp_free`.
Correct today only because `honker_cpp_free` *is* `std.c.free` and the shim links
statically — nothing in the header enforces that, and a shim on a different
Windows CRT would corrupt the heap. `Queue::get_job()` is new on this branch and
its only free goes through `get_job_json()`. **Fixed** (`8921723`).

## 5. The four `catch (...)` survivors, checked independently

| site | claim | verdict |
| --- | --- | --- |
| `Database` ctor | "close() releases the handle, then the original failure is rethrown" | **True.** READ: `db_` is set before `open_core_watcher()`, `close()` runs the full teardown, `throw;` rethrows. If `close()` itself threw you would lose the original, but you would still get an exception, not a terminate. |
| `~Lock` `release` | "an unreleased lock still expires on its TTL" | **True.** READ `honker_core::lock_acquire`: it opens with `DELETE FROM _honker_locks WHERE name = ?1 AND expires_at <= unixepoch()`. A leaked lock is reclaimed at TTL. `release()` genuinely can throw, so the swallow is load-bearing and correct. |
| `Lock::operator=(Lock&&)` | "same reasoning as the destructor" | **True.** `noexcept`, and the subsequent member moves are all `noexcept` for the default allocator. |
| `~StreamSubscription` `flush_offset` | "the lost checkpoint is survivable… call save_offset() if you need the failure reported" | **Was false.** See finding 1. True as of `d89aef4`. |

Three of four held up. The fourth is the one that mattered.

## 6. Completeness and nullability — clean. RAN.

All twelve fields present on both types, consistently. `worker_id` /
`claim_expires_at` are `std::optional` on `JobSnapshot` and asserted
`!has_value()` while pending; plain values on `Job`, which is right because the
claim UPDATE always writes both, and `Job::from_json` requires them rather than
defaulting. `expires_at` is `std::optional` on both and the fixture proves it is
per-row by leaving the second row's expiry null. No `0` or `""` stands in for a
null anywhere. The fixture's choice of 137/9/42/7/300s/9000s makes every
field-swap detectable, and the `check_ne_i64(high.run_at(), high.created_at())`
is the assertion most bindings got wrong.

## 7. Test coverage — what has none

Now covered by this round: the checkpoint write path, the nullable-helper row
shape. Still uncovered, deliberately:

- **`Queue::claim_batch` / `Queue::get_job` end to end on a malformed blob.** The
  decode tests call `detail::` and `from_json` directly because there is no way to
  make the real core hand back a bad blob. Acceptable, and it is the same
  functions the public methods call.
- **`zig build test` with no `-Dhonker-ext` is green having run almost nothing.**
  All four test binaries print `skip:` and `return 0`. CI passes the flag through
  `run.setEnvironmentVariable`, so the C++ step inside `binding smoke · ubuntu
  aggregate` really does run these (READ `build.zig` + `ci.yml`). But a green
  local `zig build test` proves nothing on its own. Pre-existing across all four
  files; not this PR's to fix, worth a `-Dhonker-ext` requirement someday.
- **Fixed `/tmp` filenames** (round 1's finding 5). Left: `test_parity.cpp`'s
  `tmp_db()` is a fixed name plus a per-process counter, so it has the identical
  collision surface. Fixing one file and not the other would be worse than either.

## 8. `honker_cpp_enqueue` and the coming JSON contract (#152/#153). READ, not fixed.

`Queue::enqueue` hands `payload_json` straight to the shim with nothing looking at
it. Two things a C++ caller should know once core enforces JSON:

- The error will not be silent, but it will be useless: `honker_enqueue` fails →
  `step_scalar_int64` returns -4 → `throw Error{"enqueue failed: code -4"}`. That
  message is byte-identical to the one you get for a disk error or a closed
  handle. Nothing says "your payload is not JSON". Worth a distinguishable code
  from the shim, or a message that names the payload, when #152/#153 land.
- `std::string_view` → `std::string{…}.c_str()` truncates at an embedded NUL,
  silently. Every `string_view` parameter in this header has that property, so it
  is not new, but a payload truncated at a NUL can still be valid JSON, which
  makes it silent corruption rather than a loud failure. Flagging only.

Not implementing validation here, as instructed.

## 9. `Database::get_job` does not exist in C++ (for #134)

Carrying forward round 1's point since it did not get fixed and should not be
fixed here. The JVM shape (#141) is `Queue.getJob` filtering by queue, `Database.getJob`
staying global. C++ has **no** `Database::get_job` at all, so #134 cannot just add
a filter to `Queue::get_job` — it has to add the global method in the same pass or
C++ users lose the cross-queue lookup entirely. `0e09ffa` softens the doc comment
that stated the global contract as a promise, so it does not become a lie the day
#134 lands.

Related, one layer below the decoders: **`get_job()` cannot tell "not found" from
"SQL error."** `step_scalar_text_duped` returns `null` for `rc != SQLITE_ROW`,
for a NULL column, and for a failed `sqlite3_prepare_v2` alike (READ
`src/honker.zig`), so `honker_cpp_get_job` collapses all of them, and
`get_job()` answers `std::nullopt` — a plausible-looking result — for a genuine
failure. Same for `claim_one`/`claim_batch` returning empty. That is the largest
remaining hole in "fails loudly", and it is in `src/honker.zig`, which is out of
bounds for this PR. Worth its own issue.

---

## Verdict on the no-templates decision: right call, wrong stated reason

**Right call for this PR.** #136 prescribes it, and adding a codec is new API
surface — that is a product decision, not something an adversarial review gets to
introduce.

**But the reasoning does not survive contact with C++.** The issue groups C++ with
Ruby and Elixir under "keep dynamic values or raw JSON". Ruby and Elixir are
dynamically typed; raw decoded JSON *is* their idiomatic value. C++ has no dynamic
values. It is the only statically typed binding in the set with no typed path:
Rust has `payload_as<T>()`, Go has generics, .NET and JVM have codecs, TS and Bun
have `Queue<T>`.

And the PR's stated reason for wrapping `nlohmann::json::exception` — *"a caller of
a honker API should not have to know which JSON library the header happens to
use"* — is already false in three public places: `Outbox`'s delivery callback is
`std::function<void(const nlohmann::json&)>`, `Job::from_json` and
`JobSnapshot::from_json` take `const nlohmann::json&`, and `nlohmann-json headers`
is a hard requirement in the README. nlohmann is not an implementation detail
here; it is the binding's dependency. The wrapping is still the right thing —
it stops an uncatchable type from escaping a documented `catch` — but the
justification for *not* adding a typed path cannot lean on avoiding nlohmann,
because nlohmann is already in the public API.

Given that, the cheapest typed path is three lines and no new dependency:

```cpp
template <typename T> T payload_as() const {
    return nlohmann::json::parse(payload_).template get<T>();
}
```

mirroring Rust's `payload_as<T>()`. **I did not add it** — it is new semantics and
belongs to you, not to review. But "C++ keeps dynamic values" is not an argument
that holds; "a codec has not earned its place yet, and we would rather decide the
shape once across bindings" is, and is what the PR body should say.

## Would I merge

Yes, now. The stream checkpoint bug was a real one hiding behind a comment that
said it was handled, and it was the kind that stays hidden for years because the
failure looks exactly like success. With that fixed and covered, the branch is
complete against #136's C++ checklist, the ABI is untouched (`diff` of the whole
89-line `extern "C"` block against `origin/feat/node-typed-jobs`: identical;
`src/honker.zig` has no diff), and every new guard fails its test when reverted.

Two things I would want on the issue list before this ships as a release, neither
blocking: the `honker_cpp_get_job` null conflation in `src/honker.zig` (finding 9),
and a decision on `payload_as<T>()` one way or the other so C++ is not the
statically typed binding with no answer.

PR #153, opened 2026-09-01
Enforce JSON payload contract in core

### LINKED ISSUE #146: Snapshot payload encoding differs across bindings

### LINKED ISSUE #152: Make JSON the payload contract: validate at the core entry points
Payload is `TEXT NOT NULL` with no validation anywhere. Bindings JSON-encode on
enqueue, but the extension path does not: an ORM user calling
`SELECT honker_enqueue('emails', ?, ...)` on their own connection can store
anything. Nothing documents a JSON requirement either.

That makes `Queue<T>` / `Job<T>` / `JobSnapshot<T>` from #124 sound by convention
only, and it is why #146 exists.

## Decision

JSON is the contract. Enforce it in core at the point of entry, not with a table
CHECK — a CHECK needs a full table rebuild (SQLite has no `ALTER TABLE ADD
CONSTRAINT`) and buys only the case where someone hand-writes
`INSERT INTO _honker_live`, which is an internal table.

Validate in these five, which is every path a payload can arrive by:

| entry point | writes |
|---|---|
| `honker_enqueue` | `_honker_live.payload` |
| `honker_stream_publish` | `_honker_stream.payload` |
| `honker_result_save` | `_honker_results.value` |
| `honker_scheduler_register` / `_update` | `_honker_scheduler_tasks.payload` |
| notify (`lib.rs:323`) | `_honker_notifications.payload` |

`_honker_dead.payload` is copied by core from an already-validated `_honker_live`
row, so it needs nothing.

Validate without materializing:

    serde_json::from_str::<serde_json::de::IgnoredAny>(payload).is_ok()

## Read side

Once writes are validated, a decode failure means legacy data or corruption —
both should be loud. Node's `parseJson` currently returns the raw string on
failure, so `Job.payload` is really `T | string` and the declared type lies.
Ruby and Python have their own rescue/except paths that need the same audit.

## Migration

No rebuild. Detection only, per table:

    SELECT id FROM _honker_live WHERE NOT json_valid(payload);

Anyone who only ever used a binding gets zero rows. The break is real but narrow:
someone deliberately storing non-JSON through the extension path starts getting
errors on enqueue.

Breaking, so it ships in the same release as #124's `getJob` change with one
migration guide. Separate PR — #124 has nine PRs stacked on it.

Supersedes #146: if payload is always valid JSON, decoded is always safe.

russellromney:
Closes #152. Also closes #146 — see below.

## Why this change is needed

Honker stores payloads as text, but its typed job APIs assume that text is JSON.

Most typed enqueue APIs serialize values before writing them. The SQLite extension did not enforce the same rule. A caller using its own SQLite or ORM connection could run `honker_enqueue` with any text, including text that could not be decoded as JSON later.

This change makes JSON a core contract instead of a binding convention.

## What changes

Honker validates payload text at every public write path, before the database write:

- queue enqueue
- stream publish
- notification publish
- result save
- scheduler register
- scheduler update when a new payload is supplied

For example, this still works:

```sql
SELECT honker_enqueue(
  'emails',
  '{"to":"alice@example.com"}',
  NULL, NULL, 0, 3, NULL
);
```

This now fails before a queue row is written:

```sql
SELECT honker_enqueue('emails', 'not json', NULL, NULL, 0, 3, NULL);
-- honker: payload must be valid JSON: expected ident at line 1 column 2
```

JSON is not limited to objects. Arrays and bare scalar values remain valid:

```text
[1, 2]
42
"ready"
true
null
```

A plain string such as `ready` is not JSON. Its serialized JSON form is `"ready"`.

## Validation is as strict as the read side

The validator parses into `serde_json::Value` — exactly what the read side does — rather than scanning the grammar structurally with `serde::de::IgnoredAny`.

A structural scan walks the grammar and discards values. It never unescapes strings, range-checks numbers, or bounds nesting, so it is strictly more permissive than any decoder. Measured on the previous revision of this branch:

| payload | structural scan | `serde_json::Value` |
|---|---|---|
| `"\ud800"` | ok | unexpected end of hex escape |
| `{"a":"\ud800"}` | ok | unexpected end of hex escape |
| `1e999` | ok | number out of range |
| `-1e999` | ok | number out of range |
| `{"a":1e999}` | ok | number out of range |
| 4096 nested arrays | ok | recursion limit exceeded |

`json.dumps` in Python and `JSON.stringify` in Node both emit lone surrogates. So a Python producer could enqueue a job, Node could read it back, and any Rust-side `serde_json` decode — `Job<T>` from #124 — would fail. That is a poison job created by one process and discovered by another.

Validating with `Value` puts the failure on the caller that caused it.

## Breaking change

Wider than first described. Three payload classes that used to enqueue now fail at the producer:

- **Lone surrogates**, such as `"\ud800"` or `{"a":"\ud800"}`. Valid by the JSON grammar, not representable as UTF-8.
- **Numbers outside f64 range**, such as `1e999`, `-1e999`, `{"a":1e999}`.
- **Nesting deeper than 128 levels**, which is `serde_json`'s own recursion limit. Previously there was no depth bound at all; 50,000 nested arrays enqueued fine.

This is the intended behavior, not a regression. A producer emitting any of these was already creating a job a consumer could not decode — the error moved, it did not appear. Callers relying on one of these shapes should encode it: escape a surrogate pair properly, send an out-of-range number as a string, or flatten the nesting.

## Error messages

`honker: payload must be valid JSON` was misleading for the new cases. A lone surrogate *is* valid JSON by the grammar; it just cannot be decoded. Telling that caller their payload is not valid JSON sends them hunting for a syntax error they do not have.

Two messages now, both carrying serde's own text:

```text
honker: payload must be valid JSON: expected ident at line 1 column 2
honker: payload is valid JSON but cannot be decoded: number out of range at line 1 column 5
honker: payload is valid JSON but cannot be decoded: unexpected end of hex escape at line 1 column 8
honker: payload is valid JSON but cannot be decoded: recursion limit exceeded at line 1 column 128
```

The second parse that picks between them runs only on the rejection path, so the accept path is a single parse.

## Performance

The `IgnoredAny` scan was chosen to avoid allocating a `Value` tree. Measured rather than assumed, on `tests/test_performance_floors.py::test_enqueue_throughput_floor_one_tx` (10,000 enqueues in one transaction, release build, M-series laptop, 5 runs each):

| | median | per enqueue |
|---|---|---|
| before (`IgnoredAny`) | 0.235 s | 23.5 µs |
| after (`Value`) | 0.241 s | 24.1 µs |

About +2.3%, or +0.6 µs per enqueue, against a 3.0 s floor. Enqueue cost is dominated by the SQLite insert and binding marshaling; a small `Value` that is dropped immediately does not show up. The full floor suite passes.

## Database and binding behavior

No table `CHECK` constraint and no schema rebuild. Validation stays at the public core entry points. Direct writes to internal tables such as `_honker_live` are still outside the public contract.

`_honker_dead.payload` is not validated again. Core only copies it from an already validated live queue row.

Note for the migration guide: `SELECT id FROM _honker_live WHERE NOT json_valid(payload)` still does not agree with honker exactly. SQLite's `json_valid()` returns 1 for `"\ud800"` and `1e999`, so it will miss legacy rows that honker now rejects. To find every row that will not decode, decode it. In the other direction the disagreement is now closed: honker's depth bound (128) is stricter than `json_valid()`'s (~1000), so the query can no longer flag a row honker itself wrote.

## Tests

All five entry points are covered for both rejection classes, including both arities of `honker_scheduler_register` and `honker_scheduler_update`, and `notify`. Bare scalars (`42`, `"str"`, `true`, `null`) plus `{}` and `[]` are asserted to still enqueue and store byte-identical text — the case a stricter validator is most likely to break by accident. Ordinary nesting (32 levels) still enqueues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,11 +40,35 @@
   an M-series laptop, about 23.5 µs to 24.1 µs per enqueue against a 3.0 s
   floor. The enqueue cost is dominated by the SQLite insert and binding
   marshaling, not by the JSON check.
+- **Check `_honker_scheduler_tasks` before upgrading.** Existing rows are not
+  validated retroactively, and the two tables behave very differently. A
+  legacy `_honker_live` row is inert: it still claims, retries, and
+  dead-letters, and the decode failure lands on whichever consumer reads it. A
+  legacy `_honker_scheduler_tasks` row is not. `scheduler_tick` fires every due
+  schedule in one call, so one unfixable payload fails the whole tick — no
+  schedule advances its `next_fire_at` and **none of your other schedules
+  fire** until that row is fixed or removed. The tick error names the schedule
+  and queue so you can find it:
+
+  ```text
+  honker: payload must be valid JSON: expected ident at line 1 column 2
+    (schedule "nightly-report", queue "reports"); fix or remove the schedule
+    row, then tick again
+  ```
+
+  Repair it with `honker_scheduler_update(name, NULL, '<json>', ...)`, or drop
+  it with `honker_scheduler_unregister(name)`.
+- **`save_result` takes raw text in most bindings** and now requires that text
+  to be JSON. Rust, Go, Node, Bun, Ruby, Elixir, C++, JVM, and Kotlin all pass
+  the value through unserialized, unlike their `enqueue` / `publish` / `notify`
+  methods. `save_result(id, "done", ttl)` used to store `done` and now fails —
+  pass `'"done"'`. Python and .NET serialize for you and are unaffected.
 - The migration detection query in the guide,
-  `SELECT id FROM _honker_live WHERE NOT json_valid(payload)`, still does not
-  agree with honker exactly. `json_valid()` returns 1 for lone surrogates and
-  for `1e999`, so it will miss legacy rows that honker now rejects. To find
-  every row that will not decode, decode it.
+  `SELECT id FROM _honker_live WHERE NOT json_valid(payload)` (and the same
+  query over `_honker_scheduler_tasks`), still does not agree with honker
+  exactly. `json_valid()` returns 1 for lone surrogates, for `1e999`, and for
+  nesting between 129 and about 1000 levels, so it will miss legacy rows that
+  honker now rejects. To find every row that will not decode, decode it.
 
 ## 2026-08-27 — Node 0.5.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # CHANGELOG
 
+## Unreleased — JSON payload contract
+
+- **Breaking:** every payload accepted by Honker core must be valid JSON. Queue
+  enqueue, stream publish, notifications, task results, and scheduler register
+  or update now reject non-JSON text with
+  `honker: payload must be valid JSON`. Objects, arrays, strings, numbers,
+  booleans, and `null` remain valid payloads. Language bindings already encode
+  payloads as JSON; callers using the SQL extension directly must pass JSON
+  text.
+
 ## 2026-08-27 — Node 0.5.1
 
 - Node `@russellthehippo/honker-node`: 0.5.1, with the four
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,49 @@
 
 ## Unreleased — JSON payload contract
 
-- **Breaking:** every payload accepted by Honker core must be valid JSON. Queue
-  enqueue, stream publish, notifications, task results, and scheduler register
-  or update now reject non-JSON text with
-  `honker: payload must be valid JSON`. Objects, arrays, strings, numbers,
-  booleans, and `null` remain valid payloads. Typed binding APIs encode values
-  where documented; raw JSON APIs and direct SQL callers must pass serialized
-  JSON text.
+- **Breaking:** every payload accepted by Honker core must be JSON that a JSON
+  decoder can read back. Queue enqueue, stream publish, notifications, task
+  results, and scheduler register or update all validate before the write.
+  Objects, arrays, strings, numbers, booleans, and `null` remain valid
+  payloads. Typed binding APIs encode values where documented; raw JSON APIs
+  and direct SQL callers must pass serialized JSON text.
+- Validation parses into `serde_json::Value` — the same thing the read side
+  does — rather than scanning the grammar structurally. A structural scan does
+  not unescape strings, range-check numbers, or bound nesting, so it accepted
+  text that no decoder can read. Three payload classes that used to enqueue
+  now fail at the producer instead:
+  - **Lone surrogates**, such as `"\ud800"` or `{"a":"\ud800"}`. Valid by the
+    JSON grammar, not representable as UTF-8. Both `json.dumps` in Python and
+    `JSON.stringify` in Node can emit these.
+  - **Numbers outside f64 range**, such as `1e999`, `-1e999`, or
+    `{"a":1e999}`.
+  - **Nesting deeper than 128 levels**, which is `serde_json`'s own recursion
+    limit. SQLite's `json_valid()` allows about 1000, so honker is now the
+    stricter of the two on depth.
+
+  This is intended, not a regression. A producer emitting any of these was
+  creating a job that a consumer could not decode — the failure moved from an
+  unrelated process at read time to the caller that caused it. If you were
+  relying on one of these shapes, encode it: escape a surrogate pair properly,
+  send an out-of-range number as a string, or flatten the nesting.
+- Rejection messages now name the actual problem and carry serde's own text.
+  Text that is not JSON gets
+  `honker: payload must be valid JSON: <serde message>`. Text that parses as
+  JSON but cannot be decoded gets
+  `honker: payload is valid JSON but cannot be decoded: <serde message>` —
+  telling that caller their payload "must be valid JSON" would send them
+  hunting for a syntax error they do not have.
+- The stricter parse is not measurable at the enqueue floor:
+  `tests/test_performance_floors.py::test_enqueue_throughput_floor_one_tx`
+  (10,000 enqueues in one transaction) went from 0.235 s to 0.241 s median on
+  an M-series laptop, about 23.5 µs to 24.1 µs per enqueue against a 3.0 s
+  floor. The enqueue cost is dominated by the SQLite insert and binding
+  marshaling, not by the JSON check.
+- The migration detection query in the guide,
+  `SELECT id FROM _honker_live WHERE NOT json_valid(payload)`, still does not
+  agree with honker exactly. `json_valid()` returns 1 for lone surrogates and
+  for `1e999`, so it will miss legacy rows that honker now rejects. To find
+  every row that will not decode, decode it.
 
 ## 2026-08-27 — Node 0.5.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,9 @@
   enqueue, stream publish, notifications, task results, and scheduler register
   or update now reject non-JSON text with
   `honker: payload must be valid JSON`. Objects, arrays, strings, numbers,
-  booleans, and `null` remain valid payloads. Language bindings already encode
-  payloads as JSON; callers using the SQL extension directly must pass JSON
-  text.
+  booleans, and `null` remain valid payloads. Typed binding APIs encode values
+  where documented; raw JSON APIs and direct SQL callers must pass serialized
+  JSON text.
 
 ## 2026-08-27 — Node 0.5.1
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,6 +177,7 @@ dependencies = [
  "notify",
  "parking_lot",
  "rusqlite",
+ "serde",
  "serde_json",
  "thiserror",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,7 +177,6 @@ dependencies = [
  "notify",
  "parking_lot",
  "rusqlite",
- "serde",
  "serde_json",
  "thiserror",
 ]
--- a/README.md
+++ b/README.md
@@ -92,8 +92,8 @@ work. Commit lands both rows. Rollback drops both rows.
 
 Payloads are JSON. Queue enqueue, stream publish, notifications, task results,
 and scheduler register/update accept any valid JSON value: objects, arrays,
-strings, numbers, booleans, or `null`. Language bindings serialize values for
-you. When calling the SQL extension directly, pass serialized JSON text (for
+strings, numbers, booleans, or `null`. Typed binding APIs serialize values where
+documented. Raw JSON APIs and direct SQL calls require serialized JSON text (for
 example, `'{"id":42}'`, `'[1,2]'`, or `'"ready"'`); non-JSON text is rejected.
 
 SQLite has no server-side push channel, so honker uses a shared watcher.
@@ -157,7 +157,7 @@ Any SQLite client that can load extensions can use honker directly:
 ```sql
 .load ./libhonker_ext
 SELECT honker_bootstrap();
-INSERT INTO _honker_live (queue, payload) VALUES ('emails', '{"to":"alice"}');
+SELECT honker_enqueue('emails', '{"to":"alice"}', NULL, NULL, 0, 3, NULL);
 SELECT honker_claim_batch('emails', 'worker-1', 32, 300);    -- JSON array
 SELECT honker_ack_batch('[1,2,3]', 'worker-1');              -- DELETEs; returns count
 SELECT honker_sweep_expired('emails');                       -- count moved to dead
@@ -181,7 +181,6 @@ SELECT honker_result_save(42, '{"ok":true}', 3600);          -- save w/ 1h TTL
 SELECT honker_result_get(42);                                -- value or NULL
 SELECT honker_result_sweep();                                -- prune expired
 SELECT notify('orders', '{"id":42}');
-SELECT honker_enqueue('emails', '{"to":"alice@example.com"}', NULL, NULL, 0, 3, NULL);
 ```
 
 The extension shares tables with the language bindings, so a Python
--- a/README.md
+++ b/README.md
@@ -91,10 +91,17 @@ All three are INSERTs inside your transaction. Put `queue.enqueue(...)`,
 work. Commit lands both rows. Rollback drops both rows.
 
 Payloads are JSON. Queue enqueue, stream publish, notifications, task results,
-and scheduler register/update accept any valid JSON value: objects, arrays,
-strings, numbers, booleans, or `null`. Typed binding APIs serialize values where
-documented. Raw JSON APIs and direct SQL calls require serialized JSON text (for
-example, `'{"id":42}'`, `'[1,2]'`, or `'"ready"'`); non-JSON text is rejected.
+and scheduler register/update accept any JSON value that decodes: objects,
+arrays, strings, numbers, booleans, or `null`. Typed binding APIs serialize
+values where documented. Raw JSON APIs and direct SQL calls require serialized
+JSON text (for example, `'{"id":42}'`, `'[1,2]'`, or `'"ready"'`); non-JSON text
+is rejected.
+
+Validation parses the payload the same way the read side does, so anything that
+enqueues will decode. That rules out a few shapes the JSON grammar allows but no
+decoder can represent: lone surrogates such as `"\ud800"`, numbers outside f64
+range such as `1e999`, and nesting deeper than 128 levels. These are rejected at
+the producer rather than surfacing as an undecodable job in another process.
 
 SQLite has no server-side push channel, so honker uses a shared watcher.
 The stable backend reads `PRAGMA data_version` every millisecond; when
--- a/README.md
+++ b/README.md
@@ -90,6 +90,12 @@ All three are INSERTs inside your transaction. Put `queue.enqueue(...)`,
 `stream.publish(...)`, or `notify(...)` beside the write that created the
 work. Commit lands both rows. Rollback drops both rows.
 
+Payloads are JSON. Queue enqueue, stream publish, notifications, task results,
+and scheduler register/update accept any valid JSON value: objects, arrays,
+strings, numbers, booleans, or `null`. Language bindings serialize values for
+you. When calling the SQL extension directly, pass serialized JSON text (for
+example, `'{"id":42}'`, `'[1,2]'`, or `'"ready"'`); non-JSON text is rejected.
+
 SQLite has no server-side push channel, so honker uses a shared watcher.
 The stable backend reads `PRAGMA data_version` every millisecond; when
 the counter changes, listeners re-read indexed SQLite state.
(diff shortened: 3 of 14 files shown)


russellromney:
Adversarial review. Throwaway worktree at 3130366, probes appended and reverted after.

**Attacked and could not break (all measured):**

- **Coverage.** All five entry points are hit, including both arities of `honker_scheduler_register` and `honker_scheduler_update` and `notify` in `lib.rs`. I enumerated every registered scalar function that takes payload text; nothing is missed. No binding writes a payload row outside these functions (the only raw INSERT left in `packages/` is `_honker_stream_consumers` in node's `api.js`).
- **Bare scalars.** `42`, `"str"`, `true`, `null`, `[]`, `{}` all enqueue, publish, save, register, update and notify fine, and store byte-identical text. Non-JSON is rejected with nothing written.
- **No `Value` allocation.** `serde_json::from_str::<serde::de::IgnoredAny>`, confirmed.
- **Stacking.** Merging this with `feat/claimed-at` (which contains #138) conflicts only in `CHANGELOG.md`; `honker_ops.rs` and `lib.rs` auto-merge, and neither of those branches enqueues a non-JSON payload in a test, so the merged suite stays green. `.optional()?` from #138 only maps `QueryReturnedNoRows`, so it does not interact with the new `UserFunctionError`.

### 1. (medium) `IgnoredAny` accepts payloads that will not decode, so "valid JSON therefore safe to decode" is not what this buys

Measured on this branch:

```
"\ud800"       IgnoredAny=ok   serde_json::Value=ERR(unexpected end of hex escape)
1e999          IgnoredAny=ok   Value=ERR(number out of range)
-1e999         IgnoredAny=ok   Value=ERR(number out of range)
{"a":1e999}    IgnoredAny=ok   Value=ERR(number out of range)
```

`IgnoredAny` scans structure; it does not unescape strings or range-check numbers.

Concrete path: `json.dumps('\ud800')` in Python produces `'"\\ud800"'` (verified), and `JSON.stringify` in Node produces the same. Python enqueues it, core accepts it, Node reads it back fine, and any Rust-side `serde_json` decode of that payload — `Job<T>` in honker-rs from #124 — fails. That is a poison job that now fails loudly on read, which is the exact case #152 says this change eliminates ("Supersedes #146: if payload is always valid JSON, decoded is always safe"). Either validate with something that walks strings and numbers, or drop that sentence from the contract and keep #146 open.

### 2. (low) The validator and `json_valid()` disagree in both directions, including with the migration detection query

Measured: 5 000 nested arrays -> honker **accepts**, `json_valid()` = 0. `"\ud800"` and `1e999` -> honker accepts, `json_valid()` = 1, `Value` rejects.

So the migration guide's `SELECT id FROM _honker_live WHERE NOT json_valid(payload)` can flag rows honker itself just wrote, and can miss rows that will not decode. Worth a sentence in the guide. Related: there is no depth limit at all — 50 000 nested arrays enqueue fine (no stack overflow; I checked for one).

### 3. (nit) `notify_accepts_every_json_payload_shape` dropped the explicit transaction

The old `notify_inserts_row` wrapped the call in `BEGIN IMMEDIATE`/`COMMIT`; the replacement runs in autocommit. The rollback test below it still covers transactional notify, so this is only the loss of the explicit-tx insert case.

**Would I merge:** yes. (1) needs a decision — tighten the validator or soften the claim — but it does not have to block this PR.

russellromney:
Addressed review finding **1 (medium)**: tightened the validator rather than softening the claim.

`validate_json_payload` now parses into `serde_json::Value` — what the read side does — instead of scanning the grammar with `IgnoredAny`. Confirmed your measurements on this branch and found a third class you flagged separately under finding 2:

| payload | `IgnoredAny` | `Value` |
|---|---|---|
| `"\ud800"` | ok | unexpected end of hex escape at line 1 column 8 |
| `{"a":"\ud800"}` | ok | unexpected end of hex escape at line 1 column 13 |
| `1e999` | ok | number out of range at line 1 column 5 |
| `-1e999` | ok | number out of range at line 1 column 6 |
| `{"a":1e999}` | ok | number out of range at line 1 column 10 |
| 128+ nested arrays | ok | recursion limit exceeded at line 1 column 128 |

So the depth hole closes too. `Value` bounds nesting at serde_json's 128-level recursion limit; `ignore_value()` is iterative, which is why 50,000 nested arrays used to go through.

**#146 can close on this.** The guarantee is now real rather than approximate: anything that enqueues will decode, because the write path runs the same parse the read path does. The "Supersedes #146" sentence in #152 holds without a caveat. I put `Closes #146` in the PR body.

**On finding 2, partially.** The disagreement with `json_valid()` is now one-directional instead of two. Honker's depth bound (128) is stricter than SQLite's (~1000, measured on 3.51.0), so the migration query can no longer flag a row honker itself just wrote. The other direction remains: `json_valid()` returns 1 for `"\ud800"` and `1e999`, so it will miss legacy rows that will not decode. Noted in the CHANGELOG and PR body — to find every undecodable row, decode it.

**Performance, measured not assumed.** `IgnoredAny` was chosen to avoid the `Value` allocation, so I ran the floor test before and after (release build, M-series, 5 runs each, `test_enqueue_throughput_floor_one_tx`, 10,000 enqueues in one transaction):

- before: 0.235 s median, 23.5 µs/enqueue
- after: 0.241 s median, 24.1 µs/enqueue

+2.3% against a 3.0 s floor. The insert and PyO3 marshaling dominate; a small `Value` that is dropped immediately is not visible.

**Error message.** `honker: payload must be valid JSON` was misleading for the new cases — a lone surrogate is valid JSON by the grammar, so that message sends the caller looking for a syntax error they do not have. Two messages now, both passing serde's own text through:

```text
honker: payload must be valid JSON: expected ident at line 1 column 2
honker: payload is valid JSON but cannot be decoded: number out of range at line 1 column 5
```

The second parse that distinguishes them runs only on the rejection path, so the accept path stays a single parse.

**Tests are load-bearing, verified by mutation.** Reverting the validator to `IgnoredAny` fails 7 tests — one per entry point plus the depth test — and leaves the bare-scalar and non-JSON tests passing, so they are not what catches it:

```
honker_ops.rs:1794: undecodable JSON payload must be rejected: "\ud800"
  orm_select_enqueue_rejects_undecodable_json_payload
  stream_publish_rejects_undecodable_json_payload
  result_save_rejects_undecodable_json_payload
  scheduler_register_rejects_undecodable_json_payload
  scheduler_update_rejects_undecodable_json_payload
  enqueue_rejects_nesting_deeper_than_the_decoder_accepts
lib.rs:1509: called `Result::unwrap_err()` on an `Ok` value: ()
  notify_rejects_undecodable_json_payload
```

Reordering the payload array to put `1e999` first fails the same six with `undecodable JSON payload must be rejected: 1e999`, so the number case is independently covered rather than shadowed by the surrogate.

Separately, dropping the serde message and the two-case split fails 11 tests, including the pre-existing non-JSON ones:

```
honker_ops.rs:1781: the error must pass serde's own message through so it names
                    the real problem, got: honker: payload must be valid JSON
honker_ops.rs:1796: expected the undecodable-JSON error for "\ud800",
                    got: honker: payload must be valid JSON
```

**Finding 3 (nit) not addressed** — the explicit-transaction notify case. Out of scope for this pass, and as you noted the rollback test below it still covers transactional notify.

Full `cargo test -p honker-core` (74 tests), `cargo fmt --all --check`, `cargo clippy -p honker-core --all-targets -D warnings`, and the Python suite (291 passed, 25 skipped) are green.

russellromney:
Second adversarial review, at `8225cf2`. Different question from round one: not "is the validator correct" but "is the contract complete, is it proven, and does it make sense to the person who hits it". Fixes pushed as `83849a4`.

Everything below is RUN unless marked INFERRED. Probes were temporary integration tests in `honker-core/tests/`, deleted before the commit.

## Attacked and could not break

**The entry points are complete.** I enumerated every path to a payload column rather than trusting the count of five.

| writer | table | validated |
|---|---|---|
| `enqueue` `honker_ops.rs:785` | `_honker_live` | yes |
| `scheduler_register` `:1238` | `_honker_scheduler_tasks` | yes |
| `scheduler_update` `:1508` | `_honker_scheduler_tasks` | yes, only when a payload is supplied |
| `result_save` `:1598` | `_honker_results.value` | yes |
| `stream_publish` `:1655` | `_honker_stream` | yes |
| `notify` `lib.rs:362` | `_honker_notifications` | yes |
| `dead_letter_exhausted_claimable` `:639`, `retry` `:878`, `fail` `:941`, `sweep_expired` `:1109` | `_honker_dead` | no, and correctly so — all four copy `_honker_live.payload` through `RETURNING`/`SELECT`, never accept caller text |
| `scheduler_tick` `:1364` | `_honker_live` | yes, redundantly — it routes the stored payload back through `enqueue` |

That is every `INSERT`/`UPDATE` naming a payload column in the crate. Six validators for six caller-facing writers; validation sits inside the Rust function, not the UDF closure, so every registered arity of `honker_scheduler_register` and `honker_scheduler_update` is covered by one call each. All 33 registered `honker_*` scalars plus `notify` were checked; the four the extension adds itself are watcher handles and write nothing. There are no triggers in the repo, and the only view (`_honker_jobs`) is a `UNION ALL`, which SQLite cannot write through without an `INSTEAD OF` trigger. `_honker_stream_consumers` holds an offset, not a payload.

No production binding writes a payload column with raw SQL — every one reaches those tables only through the validated functions. The outbox is not a bypass: all nine implementations are wrappers over a reserved `_outbox:<name>` queue going through ordinary `enqueue`. There is no requeue, revive, or dead-letter-replay helper anywhere.

**No `let _ =`.** All six call sites are `super::validate_json_payload(...)?`.

**All six sites are load-bearing.** Deleted one call at a time, full suite each time:

| deleted | tests that failed |
|---|---|
| `enqueue` | `orm_select_enqueue_rejects_non_json_payload`, `..._undecodable_...`, `enqueue_rejects_nesting_deeper_than_the_decoder_accepts` |
| `scheduler_register` | `scheduler_register_rejects_non_json_payload`, `..._undecodable_...` |
| `scheduler_update` | `scheduler_update_rejects_non_json_payload`, `..._undecodable_...` |
| `result_save` | `result_save_rejects_non_json_payload`, `..._undecodable_...` |
| `stream_publish` | `stream_publish_rejects_non_json_payload`, `..._undecodable_...` |
| `notify` | `notify_rejects_non_json_payload`, `notify_rejects_undecodable_json_payload` |

No site is silently untested, and every rejection test also asserts the row count stayed at zero (or the schedule payload stayed `{}`), so "rejected but written anyway" is covered rather than assumed.

**All three accepted break classes are tested and real.** `UNDECODABLE_JSON_PAYLOADS` covers the surrogate and out-of-range cases in five shapes across five entry points; `enqueue_rejects_nesting_deeper_than_the_decoder_accepts` covers depth, and `enqueue_accepts_ordinary_nesting` guards the other direction. Measured boundary: 127 levels accept, 128 reject. The test uses 4096 and 32 and deliberately does not pin the number, which is the right call.

**Bare scalars work end to end.** `42`, `"hi"`, `true`, `null`, plus `{}` and `[]`: all enqueue, all store byte-identical text, and all come back byte-identical through both `honker_get_job` and `honker_claim_batch`. Not just accepted at the door.

**No type-level bypass.** Every entry point with a non-TEXT payload argument — `NULL`, integer, real, and four blob shapes including `CAST('nope' AS BLOB)` — is rejected before validation runs, and nothing is written. There is no path where SQLite's type coercion smuggles unvalidated bytes into a TEXT column.

**Error text does not dump the payload.** A 5 KB bad payload produces a 77-character error naming serde's position, not the content.

## Findings

### 1. (high, fixed in `83849a4`) One legacy schedule row stops every schedule, and the error names none of them

This is the completeness gap. The migration guidance points at `_honker_live`, which is the table where a legacy bad row does not matter, and says nothing about `_honker_scheduler_tasks`, which is the table where it matters a lot.

Measured. A pre-existing `_honker_live` row with `not json` is inert: `honker_get_job`, `honker_claim_batch`, and `honker_fail` all succeed on it, and the decode failure lands on the consumer, exactly as the PR describes.

`_honker_scheduler_tasks` is different. `scheduler_tick` fires every due schedule in one call and routes each stored payload back through `enqueue`, so one row `enqueue` now rejects fails the whole tick. Two schedules, one holding `not json`, three consecutive ticks:

```text
tick1 Err("honker: payload must be valid JSON: expected ident at line 1 column 2") live_rows=0 next_fire=[("aaa-legacy-bad", 1), ("zzz-healthy", 1)]
tick2 Err(...same...)                                                             live_rows=0 next_fire=[("aaa-legacy-bad", 1), ("zzz-healthy", 1)]
tick3 Err(...same...)                                                             live_rows=0 next_fire=[("aaa-legacy-bad", 1), ("zzz-healthy", 1)]
```

`zzz-healthy` never fires. `next_fire_at` never advances on either row, so this does not self-heal — it repeats forever. Delete the bad row and the healthy one fires immediately. The lone-surrogate class does the same thing with the other message.

Failing the tick is the right half and I did not change it: skipping the row would drop its fires silently, which is the thing this project's rules forbid. The broken half is the message. A leader loop sees that string and nothing else — no schedule name, no queue, no hint that a schedule is even involved. It now reads:

```text
honker: payload must be valid JSON: expected ident at line 1 column 2
  (schedule "aaa-legacy", queue "backups"); fix or remove the schedule row,
  then tick again
```

Only `UserFunctionError` is rewritten, so a genuine `SQLITE_BUSY` keeps its code and callers can still match on it. Two tests: `a_legacy_schedule_payload_names_itself_when_it_stops_the_tick` (asserts the name, the queue, that no other schedule fired, that no `next_fire_at` advanced, and that removing the bad rows frees the scheduler) and `naming_a_scheduled_payload_error_leaves_sqlite_errors_alone`. Both proven load-bearing by mutation: removing the `map_err` fails the first with `the error must name the schedule the caller has to fix, got: honker: payload must be valid JSON: ...`; widening the match to wrap every error fails the second.

CHANGELOG now tells people to check `_honker_scheduler_tasks` before upgrading, states the blast radius, and gives the repair (`honker_scheduler_update` works on a legacy row — verified — or `honker_scheduler_unregister`).

### 2. (medium, fixed) `save_result` takes raw text in eight bindings, and two of them document the now-broken behavior

This is the break most likely to hit a real user, and it is not in the breaking-change list. `enqueue`, `publish`, and `notify` serialize for the caller in most bindings. `save_result` does not, in Rust (`packages/honker-rs/src/lib.rs:294`), Go (`honker.go:1497`), Node (`api.js:1014` — the one method that skips the file's own `jsonText()` helper), Bun, Ruby, Elixir, C++ (`honker.hpp:251`), and JVM (`Queue.java:99`). Python and .NET serialize and are unaffected.

So `save_result(id, "done", ttl)` used to store `done` and now fails. Two bindings documented exactly that:

- Ruby: "`value` is stored verbatim — JSON-encode it yourself if you want to round-trip structured data."
- Elixir: "`value` must be a string — encode JSON yourself if you want structure."

Both are now wrong in a way that sends the reader to the failure. Fixed both doc comments to say the text must already be JSON and to show the bare-string case (`'"done"'`). CHANGELOG names the affected bindings.

Making `save_result` accept a serializable value to match `enqueue`/`publish`/`notify` is the better long-term fix, but that is new API design across eight bindings and belongs in its own issue, not in this PR.

### 3. (low, fixed) The extension crate's own rustdoc still taught the bypass

`7b7e49f` fixed the README from `INSERT INTO _honker_live (queue, payload) VALUES (...)` to `SELECT honker_enqueue(...)`. The identical example in `honker-extension/src/lib.rs:12` was missed — and that is the copy that ships on docs.rs for the extension crate, where the SQL-only audience actually reads it. It is the one documented path that makes the guarantee false. Now matches the README.

### 4. (low, fixed) The `json_valid()` disagreement has a third class, not two

Measured on SQLite 3.51.0 with the planted-row query the guide recommends:

| payload | `json_valid()` | honker | detection query finds it |
|---|---|---|---|
| `not json` | 0 | reject | yes |
| `"\ud800"` | 1 | reject | **no** |
| `1e999` | 1 | reject | **no** |
| 129–~1000 nested | 1 | reject | **no** |
| 4096 nested | 0 | reject | yes |

The CHANGELOG named surrogates and `1e999` but not the nesting band. It is implied a paragraph earlier ("honker is now the stricter of the two on depth"), but the migration note is where someone actually looks. Added, and pointed the query at `_honker_scheduler_tasks` too. The claim that the disagreement is one-directional holds: 128-level nesting rejects while `json_valid()` returns 1, so the query never flags a row honker itself wrote — no false positives.

### 5. (nit, not fixed) Round one's finding 3 still stands

`notify_accepts_every_json_payload_shape` still runs in autocommit where `notify_inserts_row` used `BEGIN IMMEDIATE`/`COMMIT`. The rollback test below it still covers transactional notify, so this is only the explicit-tx insert case. Left alone, as in round one.

### 6. (nit, not fixed) `honker_enqueue('q', 42, ...)` fails with an error that mentions neither JSON nor payloads

Pre-existing rusqlite behavior, surfaced by the new docs: a SQL user reading "numbers are valid payloads" naturally writes the unquoted literal and gets `Invalid function parameter type Integer at index 1`. The README does say direct SQL calls need serialized JSON text, and nothing is written, so this is correct-but-unhelpful rather than wrong. Worth a scalar example in the SQL block if you touch it again.

Also noted, not fixed: `tests/test_litenotify.py:321,325` and `packages/honker-dotnet/tests/Honker.Tests/NotifyStreamTests.cs:247,248` are the only four places in the tracked tree that write non-JSON into a payload column. They use raw `tx.execute`, so they are outside the contract and still pass — but the PR converted the other fixtures to JSON and left these, so they read as an oversight rather than a deliberate legacy fixture.

## Would I merge

Yes, with `83849a4`.

The contract is complete and the tests are load-bearing — I could not find an unvalidated path to a payload column, and every call site fails a test when removed. Finding 1 was the one that mattered: not a hole in the validator, a hole in what happens to data written before it. A user with a pre-existing schedule holding non-JSON would have upgraded, watched their entire scheduler stop, and had no way from the error to know why. Findings 2 and 3 were the project's own docs still teaching the behavior this PR removes.

Closing #146 on this is justified. The write path now runs the same parse as the read path at every entry point that accepts caller text, and copies between internal tables never accept caller text.

Suggested follow-up issue, out of scope here: make `save_result` accept a serializable value in the eight bindings where it takes raw text, so it matches `enqueue`/`publish`/`notify` in the same files.

Fixes are `83849a4` and `1ad5c7e` (the second is a one-line clippy follow-up — my first pass at the extension doc wrapped the example and tripped `doc list item overindented`, which I missed by running clippy only on `-p honker-core`).

PR #155, opened 2026-09-02
feat(core): queue-scoped honker_cancel + capability probe (#134)

### LINKED ISSUE #134: Make getJob() and cancel() scope consistent

russellromney:
Core half of #134. No binding touched — the nine binding PRs wire this up in a second pass.

## What lands

- `honker_cancel(queue, job_id)` registered next to the existing `honker_cancel(job_id)`.
  The scoped form removes a pending or processing row only if it is in that queue; a job
  in another queue returns 0, the same answer an already-ack'd id gives.
- The queue check is inside the DELETE: `WHERE queue = ?1 AND id = ?2 AND state IN (...)`.
  Not a `SELECT` then a `DELETE` — that leaves a window for a concurrent claim to change
  the row between the two statements.
- `honker_cancel/1` is unchanged. SQLite dispatches on (name, arity), so both forms live
  on one connection; the 1-arg form becomes the global `Database.cancel(id)` and bindings
  migrate one package at a time instead of in lockstep.
- Connect-time capability probe: `honker_core::has_queue_scoped_cancel(&conn)` and the
  `CANCEL_QUEUE_SCOPED_PROBE_SQL` string behind it. A wrong arity is a hard SQLite error,
  not a fallback, and there is no `honker_version()`, so a binding built for the 2-arg
  form against an older vendored `libhonker_ext` would fail at cancel time in production.
  `pragma_function_list` reports each arity as its own row. SQL-over-extension bindings
  (Go, Ruby, .NET, C++, Elixir, Bun) run the string verbatim; Rust-side bindings call the
  helper. Errors propagate rather than being flattened to "absent".
- CHANGELOG entry. No version bumped.

Nothing existing changes behavior: every current caller still hits `honker_cancel(job_id)`.

## Tests

8 in `honker-core` against rusqlite, 2 in `tests/test_extension_interop.py` through the
real `.load libhonker_ext` path — the path Go, Ruby, .NET, C++, Elixir and Bun use.
`honker-core --lib`: 63 passed, 0 failed. `test_extension_interop.py`: 33 passed.

## Load-bearing proof

Each behavior broken, its test watched fail, then restored.

**1. Queue check removed** (`queue = ?1` → `?1 IS NOT NULL`)

```
wrong_queue_cancels_nothing_and_leaves_the_row FAILED
  assertion `left == right` failed: emails handle must not cancel an sms job
    left: 1
   right: 0
scoping_holds_for_processing_rows FAILED
  assertion `left == right` failed: wrong queue must not cancel a processing row
    left: 1
   right: 0
```
Through the loaded extension: `assert conn.execute("SELECT honker_cancel('emails', ?)", [sms_id]).fetchone()[0] == 0` → `assert 1 == 0`.

**2. `honker_cancel/1` unregistered**

```
one_arg_form_is_still_global FAILED
  called `Result::unwrap()` on an `Err` value: SqliteFailure(...,
  Some("wrong number of arguments to function honker_cancel()"))
both_arities_coexist_on_one_connection FAILED   (same error)
```
Through the loaded extension: `sqlite3.OperationalError: wrong number of arguments to function honker_cancel()`.

**3. `honker_cancel/2` unregistered** — 5 of 8 fail:

```
both_arities_coexist_on_one_connection FAILED
wrong_queue_cancels_nothing_and_leaves_the_row FAILED
scoping_holds_for_processing_rows FAILED
repeat_and_missing_return_zero FAILED
  ... Some("wrong number of arguments to function honker_cancel()")
probe_reports_true_with_the_two_arg_form FAILED
  helper must see honker_cancel/2 on a fully attached connection
```
Through the loaded extension: `assert conn.execute(probe).fetchone()[0] == 1` → `assert 0 == 1`.

**4. Probe queries the wrong arity** (`narg = 2` → `narg = 1`)

```
probe_reports_false_without_the_two_arg_form FAILED
  helper must not report a queue-scoped cancel on an old extension
```
Through the loaded extension: `assert old.execute(probe).fetchone()[0] == 0` → `assert 1 == 0`.
Note that `probe_reports_true_with_the_two_arg_form` still passed under this mutation —
the positive case alone cannot catch a probe that matches on name and ignores arity. That
is why the negative test registers a 1-arg-only `honker_cancel` (standing in for an older
extension: right name, missing arity) instead of using a bare connection.

Closes #134's core half.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,9 +27,16 @@
   changes: every current caller keeps hitting `honker_cancel(job_id)`.
 - Tests cover both arities on one connection, the wrong-queue no-op on
   pending and processing rows (asserting the row is still live afterwards),
-  idempotence, the arity error, and the probe with and without the 2-arg
-  form — in honker-core against rusqlite, and again through the real
+  the owning queue cancelling a claimed row at both arities, idempotence,
+  the arity error, and the probe with and without the 2-arg form,
+  including that an unanswerable probe is an error and not a `false` —
+  in honker-core against rusqlite, and again through the real
   `.load libhonker_ext` path in `tests/test_extension_interop.py`.
+- The shared ORM surface (`scripts/proof/orm/surface.json`, replayed by
+  every documented ORM recipe in 11 languages) gains the scoped cancel
+  and the capability probe. That is what proves the new arity and
+  `pragma_function_list` work through each binding's own SQLite build,
+  not just through rusqlite and CPython's.
 
 ## 2026-08-27 — Node 0.5.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,36 @@
 # CHANGELOG
 
+## Unreleased — queue-scoped cancel in core (issue #134)
+
+- `honker_cancel(queue, job_id)` joins the existing `honker_cancel(job_id)`.
+  The 2-arg form only removes a pending or processing row that is in that
+  queue; a job in another queue is a miss and returns 0, the same answer an
+  already-ack'd id gives. The queue check is part of the DELETE, not a read
+  before it — a `SELECT queue` followed by a `DELETE` leaves a window for a
+  concurrent claim to change the row between the two statements.
+- The 1-arg form is unchanged and stays: it is the global cancel that
+  `Database.cancel(id)` will use. SQLite dispatches on (name, arity), so both
+  forms live on one connection and bindings can move to the scoped form one
+  package at a time instead of in lockstep.
+- New connect-time capability probe, `honker_core::has_queue_scoped_cancel`
+  and the `CANCEL_QUEUE_SCOPED_PROBE_SQL` string behind it. Calling
+  `honker_cancel` at an arity the loaded extension does not have is a hard
+  SQLite error, not a fallback, and there is no `honker_version()` to ask
+  first — so a binding built for the 2-arg form running against an older
+  vendored `libhonker_ext` would fail at cancel time in production.
+  `pragma_function_list` reports each arity as its own row, so one cheap
+  query at startup turns that into a connect-time check. Bindings that talk
+  to the extension over SQL run the string; Rust-side bindings call the
+  helper. The probe propagates errors rather than reporting "absent" for
+  "cannot tell".
+- Core only. No binding calls the new arity yet, so no existing behavior
+  changes: every current caller keeps hitting `honker_cancel(job_id)`.
+- Tests cover both arities on one connection, the wrong-queue no-op on
+  pending and processing rows (asserting the row is still live afterwards),
+  idempotence, the arity error, and the probe with and without the 2-arg
+  form — in honker-core against rusqlite, and again through the real
+  `.load libhonker_ext` path in `tests/test_extension_interop.py`.
+
 ## 2026-08-27 — Node 0.5.1
 
 - Node `@russellthehippo/honker-node`: 0.5.1, with the four
--- a/honker-core/src/honker_ops.rs
+++ b/honker-core/src/honker_ops.rs
@@ -1022,6 +1022,12 @@ pub fn cancel_in_queue(conn: &Connection, queue: &str, job_id: i64) -> rusqlite:
 /// Bindings that talk to the extension over SQL (Go, Ruby, .NET, C++,
 /// Elixir, Bun) run this string verbatim; Rust-side bindings can call
 /// [`has_queue_scoped_cancel`] instead.
+///
+/// If the query itself raises, that is "cannot tell", not "absent" —
+/// a SQLite built with SQLITE_OMIT_INTROSPECTION_PRAGMAS has no
+/// `pragma_function_list` to read. Surface the error. A binding that
+/// rescues it into `false` reports an old extension while running a
+/// new one, which is the exact failure this probe exists to prevent.
 pub const CANCEL_QUEUE_SCOPED_PROBE_SQL: &str =
     "SELECT EXISTS(SELECT 1 FROM pragma_function_list WHERE name = 'honker_cancel' AND narg = 2)";
 
@@ -2067,6 +2073,22 @@ mod cancel_scoping {
             Some("sms"),
             "the processing sms row must still be live"
         );
+
+        // The owning queue does reach the claimed row. This half is
+        // what pins the 2-arg form to the same states as the 1-arg
+        // form: without it, narrowing the scoped DELETE to
+        // `state = 'pending'` leaves every other test in this module
+        // passing, and the two arities silently disagree about
+        // processing rows.
+        let n: i64 = conn
+            .query_row("SELECT honker_cancel('sms', ?1)", [sms_id], |r| r.get(0))
+            .unwrap();
+        assert_eq!(n, 1, "the owning queue cancels a processing row");
+        assert_eq!(
+            live_queue(&conn, sms_id),
+            None,
+            "the processing row is gone after its own queue cancels it"
+        );
     }
 
     /// The 1-arg form keeps its shipped meaning: global, ignores the
@@ -2081,6 +2103,35 @@ mod cancel_scoping {
             .unwrap();
         assert_eq!(n, 1, "1-arg cancel reaches a job in any queue");
         assert_eq!(live_queue(&conn, sms_id), None);
+    }
+
+    /// The 1-arg form's other shipped promise, the one its own doc
+    /// comment makes: it removes a claimed (processing) row, not just a
+    /// pending one. Nothing else in honker-core covered that, so the
+    /// two arities could have drifted apart on state — one keeping
+    /// 'processing' in its IN-list, the other losing it — with every
+    /// other test here still green.
+    #[test]
+    fn one_arg_form_cancels_a_processing_row() {
+        let conn = db();
+        let id = enqueue(&conn, "sms");
+        let _: String = conn
+            .query_row("SELECT honker_claim_batch('sms', 'w1', 8, 300)", [], |r| {
+                r.get(0)
+            })
+            .unwrap();
+        let state: String = conn
+            .query_row("SELECT state FROM _honker_live WHERE id = ?1", [id], |r| {
+                r.get(0)
+            })
+            .unwrap();
+        assert_eq!(state, "processing");
+
+        let n: i64 = conn
+            .query_row("SELECT honker_cancel(?1)", [id], |r| r.get(0))
+            .unwrap();
+        assert_eq!(n, 1, "1-arg cancel still reaches a claimed row");
+        assert_eq!(live_queue(&conn, id), None);
     }
 
     /// Idempotence and the missing-id case at the new arity: a second
@@ -2165,5 +2216,36 @@ mod cancel_scoping {
             .query_row(CANCEL_QUEUE_SCOPED_PROBE_SQL, [], |r| r.get(0))
             .unwrap();
         assert_eq!(raw, 0, "the exported probe SQL must return 0");
+    }
+
+    /// "Cannot tell" must not collapse into "absent". A real table
+    /// named `pragma_function_list` shadows the eponymous pragma table
+    /// on this connection, so the probe query no longer resolves —
+    /// the same shape a SQLite built with
+    /// SQLITE_OMIT_INTROSPECTION_PRAGMAS presents, where the pragma
+    /// table is not there at all.
+    ///
+    /// This is the test for the claim the doc comment makes. Without
+    /// it, `has_queue_scoped_cancel` could be rewritten to end in
+    /// `.unwrap_or(false)` and every other test in this module would
+    /// still pass — while a binding on a new extension got told the
+    /// extension was old and took the wrong migration path.
+    #[test]
+    fn probe_error_is_not_flattened_to_false() {
+        let conn = db();
+        assert!(
+            has_queue_scoped_cancel(&conn).unwrap(),
+            "precondition: this connection answers the probe with true"
+        );
+
+        conn.execute_batch("CREATE TABLE pragma_function_list (x)")
+            .unwrap();
+
+        let err = has_queue_scoped_cancel(&conn)
+            .expect_err("an unanswerable probe must be an Err, never Ok(false)");
+        assert!(
+            err.to_string().contains("no such column"),
+            "expected the probe query itself to fail, got: {err}"
+        );
     }
 }
--- a/honker-core/src/honker_ops.rs
+++ b/honker-core/src/honker_ops.rs
@@ -1970,6 +1970,7 @@ mod cancel_scoping {
         has_queue_scoped_cancel,
     };
     use rusqlite::Connection;
+    use rusqlite::OptionalExtension;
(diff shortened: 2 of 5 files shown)


russellromney:
## Adversarial review — queue-scoped `honker_cancel` + capability probe

I did not write this code. Every claim below is marked RUN (I executed it) or INFERRED (reasoned from the source). Reviewed `75a98c3`; I pushed one fix, head is now `e8c9da7`.

Verdict: the core change is correct, the tests are genuinely load-bearing, and the API shape is right. Two real gaps: the bug #134 actually describes is still live in every binding after this PR, and `honker_get_job` — which #134 names alongside `cancel` — is untouched with no written decision. One defect found and fixed on this branch.

### What I ran

Everything below was executed on this branch, macOS arm64, against real infrastructure (rusqlite for core, a freshly built `libhonker_ext.dylib` loaded via `.load` for the extension path).

RUN — the exact case #134 describes. Two queues, one job id: `honker_cancel('emails', <sms job id>)` returns 0 and leaves the row live with `queue='sms'`; `honker_cancel('sms', <same id>)` returns 1 and the row is gone. True for a pending row and for a claimed (`state='processing'`) row. Verified in `honker-core` (`cancel_scoping::wrong_queue_cancels_nothing_and_leaves_the_row`, `::scoping_holds_for_processing_rows`) and again through the loaded extension (`tests/test_extension_interop.py::test_extension_cancel_carries_both_arities`).

RUN — `cargo test -p honker-core --lib`: 65 passed, 0 failed, 1 ignored. `cargo fmt --all --check` clean. `cargo clippy -p honker-core --all-targets -- -D warnings` clean.

RUN — the extension really carries both arities. Built `libhonker_ext.dylib` from this branch, loaded it into stdlib `sqlite3`, and called both forms on one connection. Adding an arity is additive: existing `honker_cancel(id)` callers still resolve to the 1-arg form. Confirmed empirically, not just from the code.

RUN — the shared ORM catalog replays. Drove all 64 `scripts/proof/orm/surface.json` steps against the loaded extension with the repo's own `scripts/proof/orm/python/surface.py` runner: PASS. `scripts/proof/orm/check_surface.py`: PASS (33 functions, 16 atomicity proofs).

### Are the tests load-bearing?

Yes. I broke the guard six times and watched the right test fail each time, then restored and watched it pass.

| Mutation | Caught by |
|---|---|
| 2-arg `DELETE` ignores its queue argument | `wrong_queue_cancels_nothing_and_leaves_the_row`, `scoping_holds_for_processing_rows` FAIL |
| 2-arg narrows to `state IN ('pending')` (the arities drift apart) | `scoping_holds_for_processing_rows` FAIL: "the owning queue cancels a processing row", left 0 right 1 |
| 1-arg narrows to `state IN ('pending')` | `one_arg_form_cancels_a_processing_row` FAIL |
| `has_queue_scoped_cancel` ends in `.unwrap_or(false)` | `probe_error_is_not_flattened_to_false` FAIL |
| probe SQL drops `AND narg = 2` | `probe_reports_false_without_the_two_arg_form` FAIL |
| 2-arg deletes cross-queue but *returns 0* (return code lies) | 4 tests FAIL, incl. "the sms row must still be live", left None right Some("sms") |

That last one matters: the tests assert the row's actual state in `_honker_live`, not just the return value. A cancel that deleted the wrong row while reporting 0 does not slip through.

RUN — the extension-path test is load-bearing too, not just the rusqlite one. I applied the ignore-the-queue mutation, rebuilt the `.dylib`, and `test_extension_cancel_carries_both_arities` failed with `assert 1 == 0`. Same mutation also failed the ORM catalog replay at step `cancel_scoped_wrong_queue`.

**On `b00cb85`'s claim that both arities pin to the same job states: verified, and the test would catch them drifting.** Mutations 2 and 3 above narrow each arity independently and each is caught by its own test. Before `b00cb85` there was no test covering the 1-arg form against a processing row at all, so that commit closes a real hole.

### Does the probe report accurately?

RUN, positive half — on a connection with `attach_honker_functions` applied, and on a real `.load libhonker_ext` connection, both `has_queue_scoped_cancel()` and `CANCEL_QUEUE_SCOPED_PROBE_SQL` return true/1.

RUN with a stand-in, negative half — the "old extension" is simulated by registering a 1-arg `honker_cancel` and nothing else (Rust: `create_scalar_function(..., 1, ...)`; Python: `create_function("honker_cancel", 1, ...)`). The probe returns false/0, and a call at arity 2 raises `wrong number of arguments`. I did **not** load an actually-old `libhonker_ext` — no pre-change binary exists to load.

INFERRED — the stand-in is faithful. `pragma_function_list` has one row per `(name, narg)` registration and does not care how the function got registered, so a real 0.5.x extension that only calls `create_scalar_function("honker_cancel", 1, ...)` produces exactly the row set the stand-in produces. I consider the negative half proven for practical purposes.

RUN — "cannot tell" does not collapse to "absent". `probe_error_is_not_flattened_to_false` shadows `pragma_function_list` with a real table and asserts an `Err`. Confirmed load-bearing by mutation 4.

### Findings

**F1 — `.ok()` in the new test helper swallowed every query error. Fixed and pushed (`e8c9da7`).** `live_queue` ended in `.ok()`, which maps *all* rusqlite errors to `None`, not just `QueryReturnedNoRows`. Every `assert_eq!(live_queue(...), None)` — "the row is gone" — therefore passed whether the cancel deleted the row or the query never ran. RUN proof: I typo'd the table to `_honker_liv`; with `.ok()`, 8 of 10 tests still passed, including `one_arg_form_is_still_global` and `one_arg_form_cancels_a_processing_row`, whose only row-state check is that `None`. With `.optional().unwrap()` the broken query panics in every test that reads a row. This is the repo's known `.ok()` trap landing in new code; it is not the `scheduler_update` instance in #156.

**F2 — MEDIUM, completeness. The user-visible bug in #134 is still live in every binding after this PR.** RUN — `grep -rn "honker_cancel" packages`: all nine bindings call the 1-arg global form from a *queue* method.

- `packages/honker-node/api.js:461` (`Queue.cancel`)
- `packages/honker-dotnet/src/Honker/Queue.cs:198` (`Queue.Cancel`)
- `packages/honker-ex/lib/honker/queue.ex:119`
- `packages/honker-ruby/lib/honker.rb:807`
- `packages/honker-go/honker.go:793`
- `packages/honker/python/honker/_honker.py:565`
- `packages/honker-rs/src/lib.rs:634`
- `packages/honker-bun/src/index.ts:645`
- `packages/honker-cpp/src/honker.zig:579`

So `emails.cancel(smsId)` still deletes the SMS job on `main` today and on this branch. The CHANGELOG is honest about this ("Core only. No binding calls the new arity yet"), and core-first is a reasonable split — but this PR does not close #134, it only makes closing it possible. It should not close the issue on merge, and the follow-up should be filed before merge so the new arity does not sit unused.

**F3 — MEDIUM, completeness. `honker_get_job` is untouched and no decision is written down.** #134 asks explicitly: "Decide whether queue-scoped `getJob()` should keep checking the returned row in each binding or use a shared core function." The PR makes no `honker_get_job(queue, job_id)` and the CHANGELOG never mentions `get_job`. RUN — `honker_get_job` is registered at arity 1 only (`honker_ops.rs:520`), and `packages/honker-node/api.js:466` `Queue.getJob` does no queue filtering (#124, which adds that filter, is still OPEN — RUN: `gh pr view 124` → `"state":"OPEN"`).

Defending the omission is defensible: `get_job` is a pure read, so per-binding filtering has no TOCTOU window and is genuinely different from `cancel`. But that is a decision, and it needs a sentence in the CHANGELOG rather than silence — otherwise the next reader assumes it was forgotten. My own recommendation is the opposite of what's implied: nine bindings each re-implementing "parse the JSON, compare `queue`, return null" is nine chances to get it wrong, versus one 2-arg overload that costs almost nothing now that the pattern exists.

**F4 — LOW, design. The probe is cancel-shaped, and there is already a competing probe in the tree that gets this wrong.** RUN — `packages/honker-dotnet/src/Honker/Database.cs:393`:

```csharp
public bool FunctionExists(string name)
{
    var found = Convert.ToInt64(ExecuteScalar(
        "SELECT COUNT(*) FROM pragma_function_list WHERE name = @p0", null, name) ?? 0L);
    return found > 0;
}
```

`FunctionExists("honker_cancel")` returns **true against an old extension** — exactly the wrong answer, and exactly the failure the new probe exists to prevent. Anyone reaching for the obvious existing helper gets bitten. And #134 plans a scoped `getJob` next, which would need `has_queue_scoped_get_job` under this naming. A single `has_function(conn, name, narg)` (plus an `narg` overload on the .NET helper) generalizes for the same cost and removes the footgun. Not a blocker; worth doing before the second scoped function lands.

**F5 — LOW. The probe SQL now exists in two hand-maintained copies with nothing tying them together.** `CANCEL_QUEUE_SCOPED_PROBE_SQL` (`honker_ops.rs:1032`) and the literal string in `surface.json:195` (plus a third in `tests/test_extension_interop.py:951`). Nothing asserts they match, so the exported constant can change and the ORM proofs keep replaying the stale string. Cheap to pin.

**F6 — nit. `has_queue_scoped_cancel` has no caller outside tests.** Expected for a core-only PR, but it means the "connect-time probe" story is entirely aspirational until F2's follow-up lands. Nothing in `attach_honker_functions` runs it.

### Design and API

The overload is the right call, and the arg order (`queue` first, mirroring `honker_enqueue` / `honker_claim_batch`) is consistent with the rest of the surface.

RUN — argument coercion is identical across the two arities, which is what I'd worry about most with an overload. Both use `arg_i64` for the job id: text ids error, `REAL` whole numbers are accepted (the better-sqlite3 path), fractional reals get the good error message, NULL errors. A NULL or non-TEXT queue is a hard error, matching `honker_enqueue`. A transposed call `honker_cancel(5, 'emails')` fails loudly rather than doing something surprising. No silent defaults anywhere.

**One rationale I'd sharpen.** The doc comment says the two-statement alternative loses because "a concurrent claim can change the row between the two statements". That specific race does not actually produce a wrong answer here, and it's worth knowing why: RUN/INFERRED — `_honker_live.id` is `INTEGER PRIMARY KEY AUTOINCREMENT` (`lib.rs:359`), so ids are never reused; `queue` is never `UPDATE`d anywhere in core (`grep -rn "SET queue" honker-core/src` → no hits); and a claim only flips `state` between two values that are *both* cancellable. So a `SELECT queue` then `DELETE ... WHERE id AND state IN (...)` would land on the same row with the same queue.

The one-statement form is still clearly right — one round trip, one place to be correct instead of nine — but the safety margin currently rests on those three invariants, not on the atomicity argument as stated. If anyone ever drops `AUTOINCREMENT` or adds a move-job-between-queues feature, the two-statement version becomes genuinely unsafe. Worth saying that in the comment, because it's the version that survives contact with a future change.

INFERRED — no trap violations in the diff: no `BEGIN`/`ROLLBACK` inside a UDF (single-statement `DELETE`, no need for `in_savepoint`), no `ALTER TABLE ADD CONSTRAINT`, no schema change at all. F1 was the only new `.ok()`.

### For new issues, not this PR

1. Bindings: move every `Queue.cancel` to the 2-arg form, add `Database.cancel(id)` / `Database.getJob(id)`, run the probe at connect time. This is F2 and it is what actually closes #134.
2. `honker_result_get(job_id)` is global by job id and is exposed as a *queue* method in at least `packages/honker-jvm/src/main/java/dev/honker/Queue.java:112` (`Queue.getResult`) — `queue.getResult(foreignJobId)` returns another queue's result blob. `honker_result_save` (`Queue.saveResult`, same file, line 107) is the write side of the same hole. Same class of bug as #134, not named by it.
3. `result_get` (`honker_ops.rs:1683`) ends its `query_row` in `.ok()` — the same swallow as #156's `scheduler_update`, in a different function. Out of scope here; worth folding into #156 or filing alongside it.

### CI

RUN — green on `e8c9da7` (the F1 fix): 19 pass, 2 skipping, 0 failures. `gh pr checks 155`.

Not merging. F2 and F3 are the calls for the author: whether this lands as core-only groundwork with the binding work filed as a follow-up (my recommendation, with #134 left open), and whether the `get_job` decision gets written down.

russellromney:
Two follow-ups on the review above, one correction and one issue filed.

**Correction to proposed issue 3.** `result_get`'s `.ok()` is not still open — PR #138 already fixes it and adds `result_get_on_a_missing_result_is_still_a_miss` to cover it. It is one of the five sites #133/#138 named, not a sixth. Checked on `origin/fix/propagate-sqlite-errors`: no `.ok()` remains in `honker_ops.rs`. The review read `main`, where it is still present. Nothing to file.

**Proposed issue 2 is real and is now #159.** Confirmed: `honker_result_get` takes only `job_id`, `honker_result_save` only `job_id, value, ttl`, and all ten bindings hang them off `Queue`. Same class as #134, not named by it.

Worth flagging for whoever picks up #159: it is harder than scoped cancel. `_honker_live` has a `queue` column to filter on; `_honker_results` is keyed by `job_id` alone. Joining back to recover the queue fails in exactly the case that matters, since results outlive the job row.

**F2 and F3 are consistent with the plan, not gaps in it.** ROADMAP Phase Clemente step 3 already schedules the per-binding move to the scoped form as a second pass after these PRs merge, so #155 landing without closing #134 is intended sequencing. F3 still stands: the `honker_get_job` decision should be written down rather than left silent.

Claude-Session: https://claude.ai/code/session_01YMCFasvKC52r37DDKSTeMC

